PR #5, opened 2016-07-18
Unknown html span that swallows everything (tag/attr/content) below an unknown tag

maxme:
Note we must re-use tag handler to manage Knife tags (ul/li/etc...). To keep this PR readable, I added a TODO [here](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/7b3220042213757311e5a555b7b992e23c3d77d3/aztec/src/main/java/org/wordpress/aztec/Html.java#L531) and I'm going to open a new issue.


theck13:
:shipit:

PR #11, opened 2016-07-18
Fix #6 reintroduce taghandler

### LINKED ISSUE #6: Re-introduce the TagHandler
Re-introduce the [tag handler](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/7b3220042213757311e5a555b7b992e23c3d77d3/aztec/src/main/java/org/wordpress/aztec/Html.java#L531) after #5  is merged

maxme:
Fix #6 reintroduce taghandler

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
@@ -34,15 +34,22 @@ class AztecTagHandler : Html.TagHandler {
     private class Li
     private class Strike
 
-    override fun handleTag(opening: Boolean, tag: String, output: Editable, xmlReader: XMLReader) {
+    override fun handleTag(opening: Boolean, tag: String, output: Editable, xmlReader: XMLReader) : Boolean {
         if (opening) {
             if (tag.equals(BULLET_LI, ignoreCase = true)) {
                 if (output.length > 0 && output[output.length - 1] != '\n') {
                     output.append("\n")
                 }
                 start(output, Li())
-            } else if (tag.equals(STRIKETHROUGH_S, ignoreCase = true) || tag.equals(STRIKETHROUGH_STRIKE, ignoreCase = true) || tag.equals(STRIKETHROUGH_DEL, ignoreCase = true)) {
+                return true
+            } else if (tag.equals(STRIKETHROUGH_S, ignoreCase = true)
+                    || tag.equals(STRIKETHROUGH_STRIKE, ignoreCase = true)
+                    || tag.equals(STRIKETHROUGH_DEL, ignoreCase = true)) {
                 start(output, Strike())
+                return true
+            } else if (tag.equals(BULLET_UL, ignoreCase = true)) {
+                // no op
+                return true
             }
         } else {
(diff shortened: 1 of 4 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -34,26 +34,41 @@ class AztecTagHandler : Html.TagHandler {
...
                 }
                 end(output, Li::class.java, BulletSpan())
-            } else if (tag.equals(STRIKETHROUGH_S, ignoreCase = true) || tag.equals(STRIKETHROUGH_STRIKE, ignoreCase = true) || tag.equals(STRIKETHROUGH_DEL, ignoreCase = true)) {
+                return true
+            } else if (tag.equals(STRIKETHROUGH_S, ignoreCase = true)
+                    || tag.equals(STRIKETHROUGH_STRIKE, ignoreCase = true)
+                    || tag.equals(STRIKETHROUGH_DEL, ignoreCase = true)) {
                 end(output, Strike::class.java, StrikethroughSpan())
+                return true
+            } else if (tag.equals(BULLET_UL, ignoreCase = true)) {
```
0nko: Wouldn't it be simpler to just add it to the KNOWN_TAGS list?

maxme: Yep, that would produce the same effect, I wasn't sure. I opted for keeping a separated list of "known but ignored tags" and a second "supported tags that does nothing".

I think we could add HTML formatting there.

0nko: OK :+1:

theck13:
The functionality is the same, but it seems [this block](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/27f8c2c072f06b6ff0e57df3c2417d35b2349351/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt#L38..L70) seems more readable as the following.

```
if (tag.equals(BULLET_LI, ignoreCase = true)) {
    if (output.length > 0 && output[output.length - 1] != '\n') {
        output.append("\n")
    }
    if (opening) {
        start(output, Li())
    } else {
        end(output, Li::class.java, BulletSpan())
    }
    return true
} else if (tag.equals(STRIKETHROUGH_S, ignoreCase = true)
        || tag.equals(STRIKETHROUGH_STRIKE, ignoreCase = true)
        || tag.equals(STRIKETHROUGH_DEL, ignoreCase = true)) {
    if (opening) {
        start(output, Strike())
    } else {
        end(output, Strike::class.java, StrikethroughSpan())
    }
    return true
} else if (tag.equals(BULLET_UL, ignoreCase = true)) {
    // no op
    return true
}
```

maxme:
> The functionality is the same, but this block seems more readable as the following.

Right, and we could even have a better block using Kotlin `when`, will make that change.

maxme:
> Right, and we could even have a better block using Kotlin when, will make that change.

Updated in  af72e1d

0nko:
:shipit:

PR #13, opened 2016-07-19
Allow unknown tags to be used without root node.

### LINKED ISSUE #2: TagSoup fixes HTML when there is no root node
From the TagSoup [doc](http://home.ccil.org/~cowan/tagsoup/#what), something we could try:

> --norootbogons causes unknown elements not to be allowed as the document root element. Instead, they are made children of the default root element (the html element for HTML).

khaykov:
Redoing the patch from previous repo.

Fixes #2


maxme:
:shipit:

PR #27, opened 2016-07-26
Hidden block & inline elements

### LINKED ISSUE #3: Handle nesting of block elements in correct order
The order of block elements like `<div>` or unknown HTML tags needs to be preserved when encoded within `TextView` spans.

0nko:
### Fix
Fixes #3 - allows to have an arbitrary nested hierarchy of block and inline elements (`<div>` and `<span>` for now), which are not visible in the visual editor mode but are preserved after a conversion to HTML. 

The tricky part is to keep the correct order of start/end tags. The problem is that when the HTML is parsed, spans are created as the _opening_ tags are encountered. However, the _closing_ tag order may not be exactly reversed because the elements can be nested. That's why the closing tag order is saved in [HiddenHtmlSpan](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/009f9b0c0421b471b67805ec0a57f3f9b7435640/aztec/src/main/kotlin/org/wordpress/aztec/HiddenHtmlSpan.kt#L16) when HTML is parsed. 

When reconstructing the HTML from spans, we keep a track of the closing tag [index](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/6d8afe3dba939d8711c83de670bd12ddc243e09b/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L48). The spans are then processed in the correct [closing order](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/009f9b0c0421b471b67805ec0a57f3f9b7435640/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L272-L274).

#### Example:
```
<div class="first">
    <div class="second">
        <div class="third">
            Div<br><span class="fourth"><b>Span</b></span>Hidden
        </div>
        <div class="fifth"></div>
        <div class="sixth"></div>
    </div>
    <span class="seventh"></span>
</div>
```
The open tag order is "first", "second", "third", "fourth", "fifth", "sixth" and "seventh" but the closing tag order is "fourth", "third", "fifth", "sixth", "second", "seventh" and "first".

### Review
@maxme, @theck13 or @khaykov 

I'll keep the PR unmerged for now so that it's easier to see what's new and once it's OK I'll resolve the conflicts.


aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -26,15 +26,19 @@ import android.text.Spannable
...
 import android.text.style.StrikethroughSpan
+import org.xml.sax.Attributes
 
 import org.xml.sax.XMLReader
 
 class AztecTagHandler : Html.TagHandler {
 
     private class Li
     private class Strike
+    private class Hidden(internal var attributes: Attributes?)
```
khaykov: Looks like this class is not used.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -55,15 +59,54 @@ class AztecTagHandler : Html.TagHandler {
...
+                    order++
+                }
+                return true
+            }
             BULLET_UL -> return true // no op
         }
         return false
     }
 
+    fun handleP(text: Editable) {
```
khaykov: This method is also not used.

aztec/src/main/kotlin/org/wordpress/aztec/HiddenHtmlSpan.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun close(order: Int) {
+        isClosed = true
+        endOrder = order
+    }
+
+    fun parse() {
+        isParsed = true
+    }
+
+    private class Hidden(internal var attributes: Attributes)
```
khaykov: Here Hidden is also unused.

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -200,6 +214,23 @@ object AztecParser {
                     if (spans[j] is CommentSpan) {
                         out.append("<!--")
                     }
+
+                    if (spans[j] is HiddenHtmlSpan) {
+                        val span = spans[j]  as HiddenHtmlSpan
```
khaykov: With Kotlin we can rewrite this method to avoid casting.
Instead of accessing span through `spans[j]` every time, we can declare it at the beginning of the loop, something like `val span = spans[j]` and then, after checking `span` type with `is` condition, it will automatically be casted if true.

0nko: Oh cool!

khaykov:
@0nko : Great job with the PR! Works awesome! Except for couple of really minor things (that you fixed already), I didn't noticed anything suspicious. But the logic behind parsing is a bit complicated so I might have missed something :) 

It will also make sense to add some tests later, to make sure everything keeps working.

0nko:
Conflicts resolved.

PR #42, opened 2016-08-10
Feature/bullet lists support

khaykov:
As my original PR went out of control, I decided to split it into more manageable parts.

This is the first and most complicated (at lest for now) part of it - BulletList.

Why are lists complicated? You can do a lot of stuff with them - style in different ways, split lists, close/open and more (see `BulletListTest`), and all those manipulations must to produce valid and expected html.

I probably missed some of the usage scenarios, but I think this is a good base, and we can fine tune them later.


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -345,100 +346,125 @@ class AztecText : EditText, TextWatcher {
...
-                editableText.setSpan(AztecBulletSpan(bulletColor, bulletRadius, bulletGapWidth), bulletStart, bulletEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+        } else {
+            val boundsOfSelectedText = getSelectedTextBounds(editableText, start, end)
+
+            val startOfLine = boundsOfSelectedText.start
+            var endOfLine = boundsOfSelectedText.endInclusive
+
+            if (startOfLine == endOfLine) {   //line is empty
+                consumeEditEvent = true
+                editableText.insert(startOfLine, "\u200B")
```
0nko: I've used the same special character to aid parsing of `<div>`s [here](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L43). Couldn't that cause some conflicts for you?

khaykov: It used to! But I'm checking for it here > https://github.com/wordpress-mobile/WordPress-Aztec-Android/pull/42/files#diff-6c9a1ca267d132c2a5e6394a44bd0f1dR119
So it's fine!

0nko:
Nice work, Klym! I realized how much non-trivial functionality is missing when I was writing unit tests for bullets. 

I played around with the editor and I noticed that when you add a bullet without text and convert it to HTML, the empty bullet is skipped (there is no `<li></li>`). Is that intentional?

PR #44, opened 2016-08-17
Basic toolbar and inline styling

khaykov:
PR introduces basic toolbar and support for inline styling on the fly/on selected text.

I managed to reduce the number of tags inline styling generates by reorganising spans, but there is still couple of instances where HTML produced in not optimal (like**bold**_**italic**_ will produce `<b>bold</b><b><i>italic</i></b>,` instead of `<b>bold<i>italic</i></b>`. This is just the way current parser works (tags are closed at the beginning of other span).

This PR brought up a lot of new challenges and probably there is bunch of stuff that not working as expected, looks weird or could be written in better way :) But I believe it's something we can iterate on.

Some additional changes:
This PR includes changes to travis.yml, that are aimed at increasing the allocated memory for the test process.

Next steps:

- Link dialog and logic (already implemented, need to cover with tests).
- Fix current issues with lists and add numbered lists support.
- Proper support for quotes (share logic with lists (LeadingMarginSpan))


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -93,27 +106,45 @@ class AztecText : EditText, TextWatcher {
...
+    }
+
+    fun setOnSelectionChangedListener(onSelectionChangedListener: OnSelectionChangedListener) {
+        this.onSelectionChangedListener = onSelectionChangedListener
+    }
+
+    override fun onSelectionChanged(selStart: Int, selEnd: Int) {
+        super.onSelectionChanged(selStart, selEnd)
+        if (!consumeSelectionEvent) {
+            consumeSelectionEvent = false
```
0nko: If `consumeSelectionEvent` is `false`, set it to `false`? Is this correct?

khaykov: Oops. That's silly of me. Looks like I dropped the code that depends on it somewhere during the split of big PR. Fixed.

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -0,0 +1,134 @@
...
+
+        mEditor!!.setSelectedStyles(appliedStyles)
+        highlightActionButtons(ToolbarAction.getToolbarActionsForStyles(appliedStyles))
+    }
+
+
+    private fun onToolbarAction(action: ToolbarAction) {
+        if (!isEditorAttached()) return
+
+        //if noting is selected just mark the style as active
```
0nko: Typo ;)

0nko:
Looks great! :+1:

PR #55, opened 2016-08-23
Issue/54 hidden html bug

### LINKED ISSUE #54: Converting same instance of HiddenHtmlSpan to Html fails
Could be replicated with following test:

When we use same span with second toHtml call this test fails:
```
fun parseHtmlToSpanToHtmlNestedMixed_isEqual() {
        val input = HTML_NESTED_MIXED
        val span = SpannableString(mParser.fromHtml(input, context))

        Assert.assertEquals(input, mParser.toHtml(span))
        Assert.assertEquals(input, mParser.toHtml(span))
    }
```
When we use new instance of span with each fromHtml call this test passes:
```
   fun parseHtmlToSpanToHtmlNestedMixed_isEqual() {
        val input = HTML_NESTED_MIXED
        val spanA = SpannableString(mParser.fromHtml(input, context))
        val spanB = SpannableString(mParser.fromHtml(input, context))

        Assert.assertEquals(input, mParser.toHtml(spanA))
        Assert.assertEquals(input, mParser.toHtml(spanB))
    }
```

Could it be because the flags inside HiddenHtmlSpan are not being reseted?

0nko:
Fixes #54

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.2'
+        classpath 'com.android.tools.build:gradle:2.1.3'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.2'
+        classpath 'com.android.tools.build:gradle:2.1.3'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
theck13: We should probably make these changes in a separate issue/request in the future.

0nko: You're right. I'll do that next time.

theck13:
:shipit:

PR #58, opened 2016-08-23
Issue/56 edit hidden span bug

### LINKED ISSUE #56: Hidden HTML span parsing broken after editing
When one of the hidden HTML spans is removed in the visual editor the HTML conversion gets broken.

0nko:
Fixes #56


theck13:
:shipit:

PR #59, opened 2016-08-24
Issue/47 add url dialog

### LINKED ISSUE #47: Add/Edit link dialog
Add support for add/edit link dialog. Dialog from current editor can be used with slight modifications and improvements.

khaykov:
This PR adds Add/Edit Link dialog and link span handling logic.

Dialog layout is taken from wpandroid editor, with "Remove Link" button added.
In wpandroid editor, dialog is a fragment, but because we would want to bundle it in a library I just used regular AlertDialog attached to toolbar view lifecycle.

Appearance of dialog should be inherited from host app style.

Fixes #47


0nko:
Works great, but I found a case when selecting some text within a `<div>` doesn't convert it to a link but adds one instead. Also, I really like the toolbar divider next to the HTML button :+1:.

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -28,15 +36,43 @@ class AztecToolbar : FrameLayout {
         initView()
     }
 
+    override fun onSaveInstanceState(): Parcelable {
+        val bundle = Bundle()
+        bundle.putParcelable("superState", super.onSaveInstanceState())
+        bundle.putBoolean("isUrlDialogVisible", if (addLinkDialog != null) addLinkDialog!!.isShowing else false)
```
0nko: The dialog visibility is saved, but the link text and URL isn't when you rotate the screen.

khaykov: Totally forgot about that! Thanks, fixed!

PR #65, opened 2016-08-29
Update format toolbar HTML button

theck13:
### Fix
Update HTML toggle button from `HTML` text to `< >` icon taken from https://design.google.com/icons/#ic_code as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/64.

### DIFF
--- a/aztec/src/main/res/drawable/format_bar_button_html.xml
+++ b/aztec/src/main/res/drawable/format_bar_button_html.xml
@@ -1,18 +1,13 @@
-<vector xmlns:android="http://schemas.android.com/apk/res/android"
-        android:width="44dp"
-        android:height="44dp"
-        android:viewportWidth="44.0"
-        android:viewportHeight="44.0">
+<vector
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:height="44dp"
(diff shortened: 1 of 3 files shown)


maxme:
IMO, that's not a good idea. See the first comments here https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/370

theck13:
I think you're referring to this, right?
> On the other hand, keeping it as `HTML` is more consistent with the Calypso editor

I agree consistency is important.  That's why I suggested an icon, actually.  The rest of the buttons on the format toolbar are icons.  The `HTML` text seemed out of place.  Also, the iOS app uses the icon (i.e. `< >`) button.

I think it makes sense to use `HTML` on the desktop editor with the tab UI when you have more space as well as the `Visual` tab, and to use `< >` on the mobile editor with the button UI when real estate is at a premium.

maxme:
I think some users might be confused with `< >`, that could mean "code" or "quote" - using `HTML` makes the function obvious.

0nko:
I agree with @theck13. Since it's used on iOS I think it should be the same on Android.

khaykov:
:shipit:

PR #70, opened 2016-09-02
Issue/53 add header tags

theck13:
### Fix
Add support for header tags (`<h1>`, `<h2>`, `<h3>`, `<h4>`, `<h5>`, and `<h6>`) as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/53.

### Test
1. Select text or line with text.
2. Tap header button from format toolbar
3. Tap header size from header button menu.
4. Tap HTML button.


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -446,6 +443,207 @@ class AztecText : EditText, TextWatcher {
...
+            }
+
+            if (headerStart < headerEnd) {
+                val spans = editableText.getSpans(headerStart, headerEnd, AztecHeaderSpan::class.java)
+
+                for (span in spans) {
+                    editableText.removeSpan(span)
+                }
+
+                refreshText()
```
khaykov: Do you think we can put this method outside of the loop? Maybe we can call it once after spans are removed from each line.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -446,6 +443,207 @@ class AztecText : EditText, TextWatcher {
...
+                    TextFormat.FORMAT_HEADER_4 ->
+                        editableText.setSpan(AztecHeaderSpan(Header.H4), headerStart, headerEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+                    TextFormat.FORMAT_HEADER_5 ->
+                        editableText.setSpan(AztecHeaderSpan(Header.H5), headerStart, headerEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+                    TextFormat.FORMAT_HEADER_6 ->
+                        editableText.setSpan(AztecHeaderSpan(Header.H6), headerStart, headerEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+                    else -> {}
+                }
+
+                refreshText()
```
khaykov: Can this method be also moved outside loop?

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -938,6 +1136,15 @@ class AztecText : EditText, TextWatcher {
         }
     }
 
+    fun getAppliedHeader(selectionStart: Int, selectionEnd: Int): TextFormat? {
+        TextFormat.values().forEach {
+            if (contains(it, selectionStart, selectionEnd)) {
```
khaykov: Will the output of this method change if we change the order of elements inside TextFormat and apply some other style to header?

Let's say the TextFormat will look like this:
```
enum class TextFormat {
    FORMAT_BOLD,
    FORMAT_ITALIC,
    FORMAT_UNDERLINED,
    FORMAT_STRIKETHROUGH,
    FORMAT_BULLET,
    FORMAT_QUOTE,
    FORMAT_LINK,
    FORMAT_NORMAL,
    FORMAT_HEADER_1,
    FORMAT_HEADER_2,
    FORMAT_HEADER_3,
    FORMAT_HEADER_4,
    FORMAT_HEADER_5,
    FORMAT_HEADER_6
}
```
Now, we apply bold to some text, and then apply header style to it. In this case getAppliedHeader will return FORMAT_BOLD, because it encountered it first.

khaykov:
I noticed, that If you apply header style to multiline text, it's not being applied to partially selected lines:
 
8d9caf86a91e8df0057f8b649e5024c3

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1296,12 +1522,18 @@ class AztecText : EditText, TextWatcher {
...
     }
 
     fun isTextChangedListenerDisabled(): Boolean {
         return consumeEditEvent
     }
+
+    fun refreshText() {
+        val selStart = selectionStart
+        val selEnd = selectionEnd
+        text = editableText
```
khaykov: I think we should probably call `disableTextChangedListener()` before resetting a text, just in case.

khaykov:
There is another thing I noticed, header is set with `Spanned.SPAN_EXCLUSIVE_EXCLUSIVE`, so we can't extend it without reapplying the style. This also makes header, which should be a block element to look like it's inline.
screenshot-2016-09-06_16 13 09 475

We probably should fix this a bit later - opening/closing of spans is something that is already mostly solved with inline styles and lists ( #71 ) so we could reuse some of that code to deal with headers (and Quote) too.

theck13:
> I noticed, that If you apply header style to multiline text, it's not being applied to partially selected lines:

That is odd.  It only applies the style to fully selected lines or the entire line which the cursor is on if no text is selected.  Is that the behavior we want or should we apply it to any line with selected text?

theck13:
> We probably should fix this a bit later - opening/closing of spans is something that is already mostly solved with inline styles and lists ( #71 ) so we could reuse some of that code to deal with headers (and Quote) too.

Good catch.  I created a new issue https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/74.

khaykov:
I think we should apply style to any line with selected text. This feels pretty normal, and that's how Calypso editor does it.

PR #76, opened 2016-09-07
Issue/66 add toolbar hints

theck13:
### Fix
Add a hint message when a format toolbar button is long-pressed as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/66.  The message is the content description of the button.

### Test
1. Long-press a format button.
2. Notice hint message.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/RippleToggleButton.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/RippleToggleButton.kt
@@ -5,17 +5,18 @@ import android.graphics.Canvas
 import android.graphics.Paint
 import android.util.AttributeSet
 import android.view.MotionEvent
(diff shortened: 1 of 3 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/RippleToggleButton.kt
```diff
@@ -75,6 +77,11 @@ class RippleToggleButton : ToggleButton {
         }
     }
 
+    override fun onLongClick(view: View?): Boolean {
+        Toast.makeText(context, contentDescription, Toast.LENGTH_SHORT).show()
```
khaykov: We should probably add empty string check for contentDescription. It wont cause NPE, but the toast will be empty when content description is not set.

khaykov:
Looks good!

PR #77, opened 2016-09-08
Remove unused tag attributes and string resources

theck13:
### Fix
Remove unused `android:tag` attributes and associated string resources.


khaykov:
:shipit:

PR #81, opened 2016-09-09
Issue/update app

theck13:
### Fix
Update the app to include heading and ordered list stylings.

We can keep this branch active if we want to separate updating the `app` and `aztec` packages.


0nko:
Oh, it's already been merged, n/m.

Issue #10, opened 2016-07-18
Keep raw attributes for all supported nodes


maxme:
For each node we support, we must make sure attributes are kept when we do the import / export dance. Users could have `class` attributes on any kind of nodes.

PR #89, opened 2016-10-11
Attributes

0nko:
Addresses #10 and implements the attribute preservation for the following tags:

- [x] Strikethrough
- [x] Ordered list
- [x] Unordered lists
- [x] Individual list items
- [x] Headers
- [x] Bold style
- [x] Italics style
- [x] Underline style
- [x] Blockquote
- [x] Link
- [x] Small & large style
- [x] Subscript & superscript
- [x] Monospace style
- [x] Font
- [x] Paragraph

This PR is long as it is, the unit tests will be pushed in a separate PR. @khaykov, would you mind taking a look a this tomorrow? We'd like to prepare an alpha build of the demo app for others to try out. Thanks! It seems like a lot of code but there's a lot of repetition for different tags, so hopefully it shouldn't be too hard to figure out :).


khaykov:
Great Job with the PR! Code looks good to me. I think we can refactor some things later, after the tests. 

Also, there is a conflict with the copy/paste branch that was merged today, so I'm working on alternative that will probably resolve big part of the conflict.

PR #92, opened 2016-10-16
Issue/90 fix copy paste styling

### LINKED ISSUE #90: Styling not preserved when copy/pasting
Styling not preserved when copy/pasting and switching to HTML mode.

### Reproduced
1. Copy styled text
2. Paste it elsewhere
3. Switch to HTML mode

khaykov:
Fixes #90 and in addition turns off active inline styling after user inputs new line, like in Calypso editor.


theck13:
Is there a reason why headings were not included?

khaykov:
@theck13 Yes! This PR covers only inline styles. Block and "inline-block" (?) elements would require different logic, and I wanted to cover them later. For example, with headings we will have to split the string where we past too. With block elements we will have to copy block styling only if the first character was selected, and also split string on paste. And stuff like this :)

theck13:
That makes sense.  The code looks good to me.  :shipit:

PR #99, opened 2016-10-18
Add Special Comments and Update Style

theck13:
### Fix
Add special comments (i.e. more, page) to app and minor style tweaks.


0nko:
:shipit:

PR #100, opened 2016-10-19
Issue/93 fixing editor space newline issue

### LINKED ISSUE #93: Html editor adds whitespace between inline elements
Type something like **bold**_italic_, then switch to html mode and back. The text will become **bold** _italic_, with space between words.

khaykov:
Fixes #93 

When we show html in source editor it goes through "pretty-print" feature of Jsoup, which ads spaces and newlines around inline elements, and those spaces sneak back into visual editor.

This PR removes all the newlines around non-block elements, so the source resembles actual text more (same as Calypso). This also spaces from appearing between inline elements where they are not supposed to be. 

To test:

1. - Style text to look something like this: **bold**_italic_
2. - Go to HTML mode.
3. - Switch back to visual editor.
4. - Everything should remain as before.

PS: I'm no so good with regexp, so if anyone know's a better way to dealing with this - let me know!


aztec/src/main/kotlin/org/wordpress/aztec/source/Format.kt
```diff
@@ -6,22 +6,27 @@ import java.util.regex.Pattern
...
     // list of block elements
-    private val block = "div|span|br|blockquote|ul|ol"
+    private val block = "div|span|br|blockquote|ul|ol|li|h1|h2|h3|h4|h5|h6"
 
     fun addFormatting(content: String): String {
-        // let's use Jsoup for HTML formatting for now because it works out of the box
         val doc = Jsoup.parseBodyFragment(content)
-        return doc.body().html()
+
+        //remove newline around all non block elements
```
theck13: I'm wary of altering regular expressions, especially if they already work.  We could nest the replacements like this below, but I'm not sure that's an improvement.  It actually might make the code more difficult to read.

```
return replaceAll(
    replaceAll(
        replaceAll(
            doc.body().html(),
            "(?<!</?$block>)\n<(/?(?!$block).)>",
            "<$1>"
        ),
        "<(/?(?!$block).)>\n(?!</?($block)>)",
        "<$1>"
    ),
    "(<br>)(?!\n)",
    "$1\n"
)
```

@0nko, would you like to take a look?

0nko: What would nesting accomplish? Or is it just for readability? I guess either's fine. But from Klym's code you can maybe better understand what's going on from the variable names.

theck13: Right.  The nested way wouldn't change much, but the variable way is probably more readable.  I asked if you would like to take a look to get another look at the regular expressions.

0nko: Oh I see. There is a bunch of stuff that I converted but have no idea if they're really necessary. The thing is some of the replacements were solely because the hybrid editor did not support certain tags/features. I only learned about that after the conversion, so I'm not sure we want to just blindly use these.. For example, the `<p>` tags get removed for some reason, which we probably don't want.

theck13:
:shipit:

PR #103, opened 2016-10-20
Save and Restore Edit History on Device Rotation

theck13:
### Fix
Keep edit history after device orientation change as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/67.

### Test
1. Enter text.
2. Tap Undo.
3. Notice history.
4. Tap Redo.
5. Notice history.
6. Rotate device.
7. Tap Undo.
8. Notice history.
9. Tap Redo.
10. Notice history.

### Alternative
I <i>really</i> wanted to put this logic in `AztecText` to keep it bundled within the library, but I couldn't get it to work.  As it is now, we'll just have to add instructions for maintaining edit history in the README.  I'm open to alternative approaches.


khaykov:
@theck13 I think we can move it to AztecText! View is saving and restoring it's state similar to activity or fragment (`onSaveInstanceState` and `onRestoreInstanceState`), and we can "inject" our custom state into this process. With the help of [this article](http://trickyandroid.com/saving-android-view-state-correctly/) I implemented basic save state/restore inside AztecText, and looks like it's working. I haven't tested it in details, so there might be some problems.

Here is the [gist](https://gist.github.com/khaykov/48140cf0ed4bf8a0778031c95b5e8af5). Probably could use some more work, but looks like the concept is working.

theck13:
That's awesome!  Using your gist, I moved the logic to the library and removed any unnecessary parts.  Let me know what you think.

0nko:
Almost works :) but I think the history cursor is not correctly restored on device rotation:

1. Type something in the visual mode (at least 2 characters)
2. Switch to HTML mode
3. Undo one change
4. Rotate the device
5. You are automatically switched to visual mode and the history is restored incorrectly

0nko:
Nice work! :shipit:

PR #104, opened 2016-10-20
Add Cursor Check to Avoid Index Out of Bounds Exception

theck13:
### Fix
Add check to ensure cursor is positive integer before removing character at cursor position from list.  The cursor was decrementing to -1 and causing an `IndexOutOfBoundsException` crash as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/101.

### Test
1. Edit text exceeding history limit (demo app is 10).
2. Tap Undo repeatedly until history limit is reached.
3. Type any character.


0nko:
🚢

PR #107, opened 2016-10-21
Fix the incorrect history being loaded

### LINKED ISSUE #106: History being overwritten incorrectly
### Reproduced
1. Type something to reach the history limit (10)
2. Undo all the way back
3. Type a character
4. Click on undo twice
5. You should be at the initial state, but the history is loaded incorrectly with previously undone state

0nko:
Fixes #106.


theck13:
:shipit:

PR #110, opened 2016-10-25
Issue/109 history scroll

### LINKED ISSUE #109: Applying undo/redo always scrolls to the bottom of the editor
### Expected
After applying undo/redo the text scroll position should stay the same as before.

### Observed
The text is scrolled all the way to the bottom, which might make the applied changes invisible if they're at the top.

0nko:
Fixes #109.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -1534,7 +1534,7 @@ class AztecText : EditText, TextWatcher {
         builder.append(parser.fromHtml(Format.clearFormatting(source), context).trim())
         switchToAztecStyle(builder, 0, builder.length)
         disableTextChangedListener()
-        text = builder
+        setTextKeepState(builder)
         enableTextChangedListener()
     }
 
--- a/aztec/src/main/kotlin/org/wordpress/aztec/History.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/History.kt
@@ -76,6 +76,7 @@ class History(val historyEnabled: Boolean, val historySize: Int) {
 
         editText.isFocusable = true
         editText.isFocusableInTouchMode = true
+        editText.requestFocus()
     }
 
     fun undo(editText: EditText) {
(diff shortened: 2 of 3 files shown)


theck13:
Nice catch creating this issue and fixing it.  Testing undo/redo was annoying when the cursor kept jumping to the end so I imagine it would have been a common complaint eventually.

By removing the end selection and disabling/enabling focus, the cursor moves to the beginning and is hidden, but the user can still input text.  Try making an edit, tapping Undo, tapping Redo, and type some text.  The text is inserted at the beginning rather than where the cursor was originally.  I think the user would expect the cursor to be visible and remain wherever it was located during the edit/undo/redo.  I'm not sure if that falls under this issue, https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/69, or a new issue.  What do you think?

0nko:
Thanks! Funny I can only reproduce that on a physical phone, not the emulator. The visible cursor was actually what was causing the scroll jumps -- I needed to remove the focus from the EditText to prevent it. It's weird that it is possible to type without the focus though. 

I tried playing with `setTextKeepingState()`, which keeps the cursor position unchanged after text update but as the text moves with undo/redo and the cursor doesn't it's not very intuitive where it ends up (happens when you undo & remove text _before_ the cursor). But I suppose it's better than before.

theck13:
Yeah, I think it's more intuitive than before.

I'm seeing some weird behavior on Android 7.1 (API 25).  Check out the animation below.

cursor

### Steps
1. Select end of ***Heading 3***.
2. Tap ***Backspace*** on keyboard twice.
3. Tap ***Undo*** in action bar three times.

Notice the heading format is removed during deletion, the text is removed during undoing, and the cursor jumps lines during undoing.  I couldn't reproduce the odd behavior on Android 6.0.1 (API 23).  I'm not sure if it's these changes are causing it or `AztecHeadingSpan` is to blame.

0nko:
Strange, I can't reproduce that.. I'm running 7.1.1, you?

theck13:
I'm testing on a Google Pixel running Android 7.1.  Maybe it's just my device or version.  Let's merge this and create another issue if it pops up again.

0nko:
Alright :)

PR #112, opened 2016-10-27
Broken hidden HTML parsing

### LINKED ISSUE #82: Adding a newline above hidden HTML span breaks it
### Expected

Adding a newline in visual mode right before hidden HTML text should add `<br>` and keep hidden HTML untouched.
### Observed

After adding a newline the hidden HTML gets modified.
### Reproduced
1. Go to the the beginning of the 1st line of hidden HTML text
2. Hit the Return key
3. Switch to HTML mode
4. The hidden HTML is modified

0nko:
Fixes #82. The problem was that when a hidden html span wasn't attached to any non-whitespace character (such as `<br><div /><br>`), the parser would skip it and break the parsing of the other hidden spans.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
@@ -68,7 +68,11 @@ class AztecParser {
             hiddenSpans[hiddenIndex++] = it.endOrder
 
             // make sure every hidden span is attached to a character
-            data.insert(data.getSpanStart(it), "" + '\uFEFF')
+            val start = data.getSpanStart(it)
+            val end = data.getSpanEnd(it)
+            if (start == end && data[start] == '\n') {
+                data.insert(start, "" + '\uFEFF')
+            }
         }
         hiddenIndex = 0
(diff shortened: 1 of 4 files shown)


khaykov:
They empty hidden html span still disappears in following circumstances:

- Load editor with `<span></span>`
- Press `enter` in visual mode
- Switch to source editor
- Notice the source is:
```
<span></span>
<br>
```
- Switch back to visual mode
- Switch to source editor again
- Notice the html dissapeared

0nko:
Yeah, I know. It's actually not the problem with the parsing. It's caused by [this line](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L1558). I tried removing the `trim()` call but it breaks several unit tests for lists. I didn't want to spend too much time on it, thinking a post with just white space is probably not very useful anyway. But if you know how to fix it that would be cool :).

khaykov:
Thanks for pointing out the exact spot! I'll make a separate issue for this problem and investigate it a bit later. Ideally we should have all the logic like this handled by parser.

PR #116, opened 2016-10-31
Issue/fixing inline span issue

khaykov:
When testing headings I noticed two issues with inline spans, and this PR addresses them:

- When using soft keyboard while adding a character to the word that is not at the end of the span, (ex. adding a letter after "Heading" in "Heading 1") the part of the span after the entered character get's removed.
- When adding text to the middle of Heading span, the span after the entered character get's removed (unrelated to above mentioned issue).

Needs review: @theck13


theck13:
:shipit:

Issue #95, opened 2016-10-17
Add unit tests for attribute persistence


0nko:
Add unit tests for attribute persistence code, added in #89.

PR #117, opened 2016-10-31
Attribute unit tests

0nko:
Addresses #95.

Moved the `<p>` parsing to `AztecTagHandler` and used the existing `AztecBlockSpan` parsing logic for it.

There is one test that's commented out because list item attribute preservation is not working yet.

### DIFF
--- a/aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
+++ b/aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
@@ -1,0 +1,280 @@
+package org.wordpress.aztec
+
+import android.app.Activity
+import android.text.SpannableString
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.Robolectric
+import org.robolectric.RobolectricGradleTestRunner
+import org.robolectric.annotation.Config
+import org.wordpress.aztec.source.Format
+
+/**
+ * Testing attribute preservation for supported HTML elements
+ */
+@RunWith(RobolectricGradleTestRunner::class)
+@Config(constants = BuildConfig::class)
+class AttributeTest {
+
+    companion object {
+        private val HEADING =
+                "<h1 a=\"A\">Heading 1</h1><br>" +
+                "<h2 b=\"B\">Heading 2</h2><br>" +
+                "<h3 c=\"C\">Heading 3</h3><br>" +
+                "<h4 d=\"D\">Heading 4</h4><br>" +
+                "<h5 e=\"E\">Heading 5</h5><br>" +
+                "<h6 f=\"F\">Heading 6</h6>"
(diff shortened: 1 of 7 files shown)


aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
```diff
@@ -0,0 +1,291 @@
...
+        editText.fromHtml(input)
+        val output = editText.toHtml()
+        Assert.assertEquals(input, output)
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun nestedAttributes() {
+        val input = NESTED
+        val expected = NESTED_REVERSED
```
theck13: Why is the order of nested attributes reversed after parsing?

aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
```diff
@@ -0,0 +1,291 @@
...
+    @Throws(Exception::class)
+    fun mixedAttributes() {
+        val input = MIXED + NESTED
+        val expected = MIXED + NESTED_REVERSED
+        editText.fromHtml(input)
+        val output = editText.toHtml()
+        Assert.assertEquals(expected, output)
+    }
+
+    //TODO: After fixing list item attribute preservation, uncomment this
```
theck13: Is there an issue for list item attributes?  I saw [this](https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/98), but I'm not sure if it's related.

aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
0nko: This is a known behavior; the order of nested elements is always reversed when parsed from visual to HTML and back. I thought there existed an issue for this, but I don't see it anywhere..

aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
0nko: Yeah, it's that one. When we were merging the attributes PR with @khaykov's work on lists there was some issues with list items so we'll need to go back to it.

aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
theck13: Sounds good.  I just wanted to make sure we were tracking it.  I saw the new issue, https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/118.

aztec/src/test/kotlin/org/wordpress/aztec/AttributeTest.kt
theck13: That works.  I just wanted to make sure we were tracking it.

theck13:
:shipit:

Issue #69, opened 2016-09-01
Cursor Position is Not Retained when Toggling HTML View


theck13:
### Expected

Switching between visual and HTML views keeps the cursor position where it is located prior to toggling.
### Observed

The cursor is reset to the first position regardless of selection in visual and HTML views.
### Reproduced
1. Select any position other than first.
2. Tap HTML button.
3. Notice cursor position is first.
4. Select any position other than first.
5. Tap HTML button.
6. Notice cursor position is first.
#### Tested

LG G4 (LG-H815) on Android 6.0.1 with AztecDemo 1.0

PR #119, opened 2016-11-04
Issue/69 retain cursor position between editor modes

khaykov:
This PR adds support for cursor synchronization between visual and source editors (#69).

This task turned out to be not as trivial as I expected - there is a huge difference between spanned text and "prettified" html, and placing cursor to be where it actually should be was quite tricky.

When we want to show cursor in source editor we are marking the end of selection with cursor span, and then in parser we insert "aztec_cursor" word where we think it should be. We don't use `<aztec_cursor></aztec_cursor>` tag because Jsoup tries to "prettify" it and everything becomes really complicated.

Moving cursor from source to visual editor was relatively easy, and we actually use the tag (`<aztec_cursor></aztec_cursor>`), parse it to span and then just consume it.

Tests will come a bit later.


0nko:
Looking good! :shipit:

PR #127, opened 2016-11-08
Issue/fix cursor inside unknown html

khaykov:
This PR fixes issue with cursor not working correctly inside uknown html tag.

### DIFF
--- a/aztec/src/main/java/org/wordpress/aztec/Html.java
+++ b/aztec/src/main/java/org/wordpress/aztec/Html.java
@@ -496,6 +496,10 @@ class HtmlToSpannedConverter implements ContentHandler, LexicalHandler {
 
     private void handleStartTag(String tag, Attributes attributes) {
         if (mUnknownTagLevel != 0) {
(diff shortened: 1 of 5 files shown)


0nko:
It's not a big deal, but if switch to HTML and leave the cursor inside the `<iframe>` tag, switch to visual and then back to HTML , you'll notice an extra newline is inserted.

khaykov:
I think you are talking about the case when some text is on same line as unknown html span. This is actually the original behavior of the parser, and not related to cursor. When you switch from source to visual editor, the "html" that is passed to it, is actually mix of html and EditText markup - so the text with unknown html span on same line will look like this -> `text\n<iframe class="classic">Menu</iframe>`. Space is being added [here](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/aztec/src/main/java/org/wordpress/aztec/Html.java#L938) to replace a newline symbol. I'm still not sure why it's like this, but will take a look.

khaykov:
@0nko Ready for another pass :)

0nko:
Cool! :shipit:

PR #130, opened 2016-11-16
Issue/124 fixing comment in unknown

### LINKED ISSUE #124: Comment tag gets removed from unknown HTML
### Reproduced
1. Go to HTML mode and start with `<test><!-- soemething --></test>`
2. Switch to visual mode
3. Switch back to HTML and notice `<test></test>`

#### Tested
Emulator with Android 7.1.1

khaykov:
### Fix #124 

Instead of applying comment span when parser is inside unknown html, just append html of the comment to it.

Also, modified the source formatting regex to include possible space when looking for newline characters.


0nko:
:ship: :it:

PR #133, opened 2016-12-05
Issue/121 fixing heading linebreaks

### LINKED ISSUE #121: Line breaks around headings
Headings should have line breaks (perhaps double) around them in visual mode like block elements.

## Observed

screenshot_1478251810

## Expected

screenshot_1478251794

khaykov:
Fixes #121 and couple of related issues.

Revisiting headings. Heading is a quite interesting element, because it has traits of both block and inline spans. Took me couple of tried to figure a approach to it, but looks like it finally works.
When parsing html to spanned we are treating heading as a Block element, like lists or quotes. This gives the resulting span a properties of block element, and we don't need `br` tag in html to separate them anymore.

When we parse heading back to html we are treating it as a block element when it comes to newline's around it, and as a special inline container when parsing it's content.

In addition, fixing original issue brought up couple of obscure bugs, that were quite annoying and often breaking everything when working with headings, so I had to fix them too. First one, was a fix for detecting of spans below cursor - this fixed the entry of styled text inside heading. Second one was fix for broken parsing of newlines between block elements, this fixed restyling of heading when moving it to line above.


0nko:
I found some weird behavior with redo/undo. When you edit the text, for example you add a character somewhere, and tap undo, then redo then the first heading gets blown up. When you switch to HTML, extra `<h1>` tags are being added:
```
<h1></h1>
<h1>Heading 1</h1>
<h1></h1>
```
When you hit redo again you'll get the correct text. So it seems like there is some extra stuff added to the history or something and it's in some sort of corrupted state.

khaykov:
If I remember correctly, when working with headings, Tyler mentioned something about weird behaviour of edit text, related to it's state. Clearing the editor before setting new content from html seems to fix this issue.

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -180,7 +202,7 @@ class AztecParser {
         out.append(unknownHtmlSpan.getRawHtml())
     }
 
-    private fun withinListThenQuote(out: StringBuilder, text: Spanned, start: Int, end: Int, list: AztecListSpan, quote: AztecQuoteSpan) {
```
0nko: good catch ;)

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -406,6 +406,8 @@ class AztecText : EditText, TextWatcher {
     }
 
     fun fromHtml(source: String) {
+        editableText.clear()
```
0nko: This creates a phantom empty text history point (edit text, switch to HTML, switch back, tap redo and the text is gone). Adding `disableTextChangedListener()` should help though.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecLineBlockSpan.kt
```diff
@@ -0,0 +1,10 @@
+package org.wordpress.aztec.spans
+
+
+interface AztecLineBlockSpan : Cloneable{
```
0nko: nitpick - whitespace before `{`

aztec/src/main/kotlin/org/wordpress/aztec/spans/UnknownHtmlSpan.kt
```diff
@@ -10,7 +10,7 @@ class UnknownHtmlSpan(rawHtml: StringBuilder, context: Context, drawable: Int) :
 
     val mRawHtml: StringBuilder = rawHtml
 
-    fun getRawHtml() : StringBuilder {
+    fun getRawHtml(): StringBuilder {
```
0nko: nitpick - whitespace before `:`

PR #134, opened 2016-12-08
Add Media/Image Support

theck13:
### Fix
Add support to allow inserting images from the device's camera or photos as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/19.

### Test
**Camera**
1. Select location or text.
2. Tap ***Media*** button in format toolbar.
3. Tap ***Photo*** option.
4. Tap ***Device camera*** option.
5. Take picture with camera.
6. Notice picture replaces selection.

**Photos**
1. Select location or text.
2. Tap ***Media*** button in format toolbar.
3. Tap ***Photo*** option.
4. Tap ***Device photos*** option.
5. Take photo from device.
6. Notice photo replaces selection.

#### Exclude
The image upload service, media library (i.e. Media > Photo > Media library), and gallery interface (i.e. Media > Gallery) reside in the [WordPress project](https://github.com/wordpress-mobile/WordPress-Android/) and are excluded from these changes.

This includes the starting framework (i.e. menu item, dialog options, runtime permissions, and activity intents) for video support as well, but the video option in the media popup menu has been hidden until video is fully supported.


theck13:
> selecting a text, and then adding an image, does not replaces selection for me, image is inserted at selectionEnd

Which version of Android are you testing?  For API 23 and above, the image replaces the selection as expected.  For API 22 and below, the popup menu displayed when tapping the Media button on the toolbar causes the contextual action bar to dismiss and the selection to reduce to the end.

> inserting image into block elements (lists, quotes and headings) is not working as expected yet

I think you're referring to the image separating block elements rather than being inserted as a list item or within a quote.  That's not something I had considered originally.

> you can edit img tag in landscape mode, but this does not affect image span

That is a weird feature of `EditText`.  The HTML toggle is disabled after adding an image to avoid editing the tag, but you can still access the HTML of the span in landscape orientation.  Maybe we could detect that plain text mode of `EditText` and disable/enable it along with the HTML toggle.

> loading images from img tag is out of scope of this PR

For the initial version, yes.  I considered parsing the `src` value and loading images from the web part of the uploading service.

khaykov:
Thanks for the clarification! The changes look good :shipit:
I'll open issues for next steps in media handling.

PR #135, opened 2016-12-09
Issue/118 fixing order of nested elements

### LINKED ISSUE #118: Order of nested elements gets reversed
During parsing, the order of nested elements gets reversed. So going from visual to HTML, where you have `<i><u>text</u></i>`, then going to visual and back to HTML you'll see `<u><i>text</i></u>`.

khaykov:
Fixes #118

The problem was caused by the way we parse html. When going through something like this `<b><i>bold italic</i> bold</b>` the first span that will be added to text is italic, because we add span when hitting the end tag, and `</i>` comes before `</b>`. This affected all the tags that have at least one shared start/end index.

What we are doing to fix it, is after html->span parsing is over, we are going through all the spans and resetting nested elements in correct order.

This PR also fixed the issue with nested block elements being reversed in visual mode.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
@@ -105,22 +105,23 @@ class AztecParser {
         }
     }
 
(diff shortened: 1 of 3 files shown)


0nko:
Nice work! :shipit:

PR #137, opened 2016-12-09
Adds support for <code> elements

### LINKED ISSUE #123: Add support for <code> element
We should support the `<code>` element so it is shown in the visual mode and highlight it similarly like the hybrid editor.

hypest:
### Fix #123 

Support `<code>` tags. This is implemented as an inline formatting span and altering the background and text colour of the text. Those colours are stylable via `colours.xml`.


aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecCodeSpan.kt
```diff
@@ -0,0 +1,85 @@
...
+import android.graphics.Typeface
+import android.os.Parcel
+import android.text.ParcelableSpan
+import android.text.TextPaint
+import android.text.TextUtils
+import android.text.style.CharacterStyle
+import android.text.style.UpdateAppearance
+import org.wordpress.aztec.formatting.InlineFormatter
+
+class AztecCodeSpan : CharacterStyle, UpdateAppearance, ParcelableSpan, AztecContentSpan, AztecInlineSpan {
```
0nko: Why did you implement the `ParcelableSpan`?

hypest: Hey @0nko so, was under the impression that that's what we need to do when we have local variables, like the colors. I got that impression from `AztecQuoteSpan`.

Judging from your question, I now presume that implementing the `ParcelableSpan` might have some other reasoning. Can you elaborate if so? Thanks!

0nko: I looked inside the `ParcelableSpan` and it's mentioned there it should only be used by the framework, that's all. `AztecQuoteSpan` inherits and overrides the `Parcelabe` methods. But it's not a big deal.

Issue #140, opened 2016-12-09
IndexOutOfBoundsException when restoring cursor position


0nko:
This [line](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L424) causes an `IndexOutOfBoundsException` because `cursorInsertionPoint < 0` with the following HTML:
```
a<ul>
  <li></li>
</ul>
```
When you switch to visual mode, place the cursor right after `a` and switch to HTML you'll get the exception.

I assigned it to @khaykov because I'm not sure where the -1 is coming from and you can probably find out a lot faster :).

0nko:
Can't reproduce anymore, closing.

Issue #128, opened 2016-11-08
Removing the first item does not renumber list


0nko:
Removing the first item of an ordered list does not renumber the rest of the list. 

Use the following HTML and press backspace at the beginning of the 1st line:

```
TEXT
<ol>
  <li>a</li>
  <li>b</li>
  <li>c</li>
</ol>
```

### Before

![screenshot_1478600412](https://cloud.githubusercontent.com/assets/1522856/20095345/71a7e0c8-a5a5-11e6-816a-ef93f4fb5cfa.png)

### After

![screenshot_1478600381](https://cloud.githubusercontent.com/assets/1522856/20095357/7845e524-a5a5-11e6-9ff8-53d9876232f7.png)

khaykov:
Was fixed in #203

Issue #120, opened 2016-11-04
Empty list items disappear


0nko:
Empty list items are removed when going from HTML mode to visual mode.

Issue #98, opened 2016-10-18
Pass attributes of <li> tags to new items added to list


khaykov:
Keep attributes of <li> tags when adding new items to list or splitting lists.

PR #142, opened 2016-12-10
Empty list items & more

### LINKED ISSUE #115: Remove source trimming logic from editor
We need to move all the source trimming/tidying logic to parser, and only let editor change source through user input, specifically to avoid issues like [this](https://github.com/wordpress-mobile/WordPress-Aztec-Android/pull/112#issuecomment-257238579).

### LINKED ISSUE #138: Block elements add extra line break to HTML above them
When converting a simple list like:
```
a
<ul>
  <li>b</li>
</ul>
c
```
After switching to visual mode and back to HTML we get:
```
a
<br>
<ul>
  <li>b</li>
</ul>
c
```

0nko:
#This PR primarily addresses issue #120, but also fixes #115, ~~#128~~ and #98. ~~I also fixed #138 when I found it, but I saw @khaykov pushed a PR that should fix this, so it can be replaced after the merge~~. I added a workaround for #140, but I'm not sure this is ideal. 

There are a lot of minor tweaks to the way lists behave and I'll add a couple of unit tests to test that later.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -385,8 +385,6 @@ class AztecText : EditText, TextWatcher {
             removeLeadingStyle(text, LeadingMarginSpan::class.java)
         }
 
-        history.handleHistory(this)
-
         blockFormatter.handleBlockStyling(text, textChangedEventDetails)
         inlineFormatter.handleInlineStyling(textChangedEventDetails)
         lineBlockFormatter.handleLineBlockStyling(textChangedEventDetails)
@@ -397,6 +395,9 @@ class AztecText : EditText, TextWatcher {
 
         // preserve the attributes on the previous list item when adding a new one
         realignAttributesWhenAddingItem(text)
+
+        history.handleHistory(this)
+
     }
 
     private fun realignAttributesWhenAddingItem(text: Editable) {
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -403,7 +403,7 @@ class AztecText : EditText, TextWatcher {
             val lists = text.getSpans(textChangedEventDetails.inputStart, textChangedEventDetails.inputStart, AztecListSpan::class.java)
             if (lists.isNotEmpty()) {
                 val listEnd = text.getSpanEnd(lists[0])
-//
+
                 // when newline inserted before the list item's newline the item's attributes must be shifted up
                 if (textChangedEventDetails.inputEnd < text.length && text[textChangedEventDetails.inputEnd] == '\n') {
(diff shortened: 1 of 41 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -230,12 +229,29 @@ class AztecText : EditText, TextWatcher {
...
+        if (selStart == selEnd && isCursorBeforeZwjChar(selEnd)) return
+
+        previousCursorPosition = selEnd
 
         onSelectionChangedListener?.onSelectionChanged(selStart, selEnd)
 
         setSelectedStyles(getAppliedStyles(selStart, selEnd))
     }
 
+    private fun isCursorBeforeZwjChar(selEnd: Int): Boolean {
```
0nko: You're right.. Good catch.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -362,20 +383,88 @@ class AztecText : EditText, TextWatcher {
...
+        // preserve the attributes on the previous list item when adding a new one
+        realignAttributesWhenAddingItem(text)
+
+        history.handleHistory(this)
+
+    }
+
+    private fun realignAttributesWhenAddingItem(text: Editable) {
+        if (textChangedEventDetails.isNewLine()) {
+            val lists = text.getSpans(textChangedEventDetails.inputStart, textChangedEventDetails.inputStart, AztecListSpan::class.java)
```
0nko: Thank for the tip!

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -362,20 +383,88 @@ class AztecText : EditText, TextWatcher {
...
+            if (lists.isNotEmpty()) {
+                val listEnd = text.getSpanEnd(lists[0])
+//
```
0nko: Hmm, I see. I had that working but the fix for the extra line breaks you submitted didn't work in my branch. So I was trying to figure out what was wrong until I finally discovered I disabled [span shortening](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L46) for lists and reverted it. And that broke the emty item at the end, since it removes the last `\n`.

0nko:
Hmm, I see. I had that working but the fix for the extra line breaks you submitted didn't work in my branch. So I was trying to figure out what was wrong until I finally discovered I disabled [span shortening](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L46) for lists and reverted it. And that broke the empty item at the end, since it removes the last \n.

0nko:
I've updated the code and it's ready for another pass. I've decided to keep #128 out of this PR because I'm not sure how long that'd take to finish. Let's get this merged :P.

khaykov:
:shipit:

PR #159, opened 2016-12-23
Mark All Inline Formats in Selection as Active in Toolbar

theck13:
### Fix
Mark all inline formats including bold, italic, and strikethrough as active in the format toolbar if they are within the selected text as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/155.

### Test
1. Enter multiple bold, italic, and strikethrough.
2. Select multiple lines.
3. Notice all formats in selection are active.


0nko:
Looking good! :shipit:

PR #187, opened 2017-01-17
Add Focus State to Format Toolbar Buttons

theck13:
### Fix
Add focus state to format toolbar buttons as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/183.

### Test
1. Run app with hardware keyboard.
2. Tap ***Tab*** key.
3. Notice format toolbar button is not shown as focused.


khaykov:
:shipit:

PR #192, opened 2017-01-18
Update Redo and Undo Buttons in App Menu

theck13:
### Fix
Update redo and undo buttons in the app menu by adding AppCompat dependency to app build file.

### DIFF
--- a/app/src/main/res/menu/menu_main.xml
+++ b/app/src/main/res/menu/menu_main.xml
@@ -3,7 +3,7 @@
 <menu
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="schemas.android.com/tools" >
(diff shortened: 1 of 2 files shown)


app/src/main/res/menu/menu_main.xml
```diff
@@ -1,16 +1,24 @@
...
-<menu xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <item android:id="@+id/undo"
-          android:title="@string/menu_undo"
-          android:icon="@drawable/ic_action_undo"
-          android:showAsAction="always">
+<menu
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="schemas.android.com/tools" >
```
khaykov: Looks like there need's to be a `http://` in front of `schemas.android.com/tools` for` tools:ignore` to work.

PR #199, opened 2017-01-24
Handle missing URL in links to prevent crash

### LINKED ISSUE #195: App crashes when link is missing the `href` attribute
If you type `<a>Test</a>` in the source editor the app crashes, because it expects the `href` attribute and does not check if it exists.

0nko:
Fixes #195.


khaykov:
:shipit:

PR #202, opened 2017-01-24
Issue/201 image blocks

### LINKED ISSUE #201: Handle image display like visual editor
~~The hybrid visual editor always shows images on a separate line, i.e. they are displayed like block elements. This should be the same in Aztec.~~

Support image download and sizing that mimics the visual editor.

0nko:
Resolves #201.


theck13:
I don't think we should make images block elements.  Here are a few unwanted side effects to doing that.
- Images cannot have surrounding text (https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/204)
- Images in block elements are broken
- Behavior contradicts Calypso

For those reasons, I would keep images inline.

0nko:
I made the images inline again and they behave and look more or less like in visual editor. There is a bunch of improvements and tweaks to make images work, so this branch is still relevant, even though the original issue was about making images block elements.

I had to, for example, change the way the spans are rendered because removing the placeholder and adding the actual image was causing text overlaps within posts. Now the span stays the same but the drawable inside is replaced.

As a consequence of these changes, #204 is not relevant anymore.

There is an associated [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/5059) in wpandroid, which fixes image loading in the WordPress app. So to test the whole thing, please use that branch.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecMediaSpan.kt
```diff
@@ -1,63 +1,92 @@
...
-class AztecMediaSpan(val context: Context, drawable: Drawable, source: String) : ImageSpan(drawable), ParagraphStyle {
-    private val html = source
+class AztecMediaSpan @JvmOverloads constructor(val context: Context?, private var image: Drawable?, val source: String, attributes: String = "") : DynamicDrawableSpan()  {
+
+    private val TAG: String = "img"
+
+    var attributes: String = ""
 
     companion object {
         private val rect: Rect = Rect()
```
theck13: The `rect` value is no longer used.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecMediaSpan.kt
```diff
@@ -1,63 +1,92 @@
...
+        val b = image
+        canvas.save()
 
-        /*
-         * Following Android guidelines for keylines and spacing, screen edge margins should
-         * be 16dp.  Therefore, the width of images should be the width of the screen minus
-         * 16dp on both sides (i.e. 16 * 2 = 32).
-         *
-         * https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-baseline-grids
-         */
```
theck13: These comments should be moved [here](https://github.com/wordpress-mobile/WordPress-Aztec-Android/pull/202/files#diff-d6152856756259d707419e470fe4f340R79) in the `setBounds` function to clarify how the width is determined.

0nko:
The images are stretched so that their dimension in px matches the value in dp (so a 300px by 300px image is stretched to 300dp by 300dp). This is what the visual editor does.

theck13:
I'm not sure stretching the images from 300px to 300dp is good.  Images smaller than the screen look pixellated.

@maxme and @aforcier, do you know why that approach was taken with the visual editor?

aforcier:
> The images are stretched so that their dimension in px matches the value in dp (so a 300px by 300px image is stretched to 300dp by 300dp). This is what the visual editor does.

> I'm not sure stretching the images from 300px to 300dp is good. Images smaller than the screen look pixellated.
> @maxme and @aforcier, do you know why that approach was taken with the visual editor?

I'm not sure if I'm missing something, but the visual editor should not be stretching images. The visual editor will display images at their natural size, scaled down to fit the screen if the width is larger than the screen's width. Otherwise, they're displayed at their natural size, which may appear larger or smaller on different screens with different densities (but should never appear stretched).

This handling is done by the WebView - all the editor does explicitly is limit the maximum allowed width the size of the screen.

The only case of stretching would be for images smaller than 30px x 30px, which are scaled up to that size so that the edit overlay is big enough to be selectable.

0nko:
Based on the observation of the visual editor I figured the images are strechted. The cat image in Elisa's intro post is 300px wide. The native resolution on the Pixel is 788px. 
### This is what it looks like in the visual editor:
screenshot_1485431215

### This is what the image looks like in Aztec in its native resolution (300px):
screenshot_1485431038

### This is what the image looks like in Aztec, when it's stretched to 300dp (788px native):
screenshot_1485452210

khaykov:
I noticed that images inserted in the line breaks it, like this image is (72x72):
break

Nexus 5X, 6.0.0

aforcier:
@0nko I think you meant that the native screen with for a Pixel is `1080px`.

I see what you mean, but I wouldn't call that stretched, the WebView is displaying the image at its natural size by correcting for the device's pixel density. The Chrome app does the same thing (and that way it also looks the same as it does on desktop, web browser or not).

In which cases are images smaller than the screen appearing pixelated? @theck13

theck13:
The visible part of this icon is 67 pixels square.  In the PX screenshot, it's 67 pixels.  In the DP screenshot, it's 175 pixels (i.e. ~2.6 times PX).

### PX
px

### DP
dp

0nko:
Right, converting the native dimensions (67px) to its dp (175px) = correcting for the device's pixel density, which is what the visual editor does and what this PR did before.

khaykov:
I fixed little issue that caused crash when adding image to unordered lists. The code looks good, but there is one big issue - adding images into block elements. 

In [this](https://github.com/wordpress-mobile/WordPress-Aztec-Android/pull/202/commits/68707742b2bfcd4f5325ee76c337022053447536) commit we removed the line splitting, which is fine when adding images into inline elements, but we still need to split block elements, and without newlines they get messed up.

I worked a bit on adding images into block elements, and as always it has gazilion of edge cases, like adding image to empty element in list, empty line in quote (logic is different from the list), adding image between block elements and so on. Will probably take another day or so, so should we merge this PR and add the support for inserting images into block elements a bit later?

0nko:
Yeah, let's do that. I'll create an issue for that and merge this.

PR #203, opened 2017-01-25
Issue/73 remove block styling with backspace

### LINKED ISSUE #73: Add ability to remove list styling with backspace
When cursor is positioned right after bullet/number and backspace is pressed remove list style from the line.

khaykov:
Fixes #73 

Adds ability to remove style from the first line of block element with backspace.

Works even when no characters were deleted (at 0 index of EditText). Unfortunately, I don't know how to test this particular behaviour programmatically, because we need to send KEYCODE_DEL to EditText, which is not supported by robolectrics yet.

To test:
- Position your cursor at the first line of any list or quote.
- Press backspace.
- Style should be removed from that line.


0nko:
Hmm, it doesn't seem to work when I run the demo app with the sample text. When I try removing either of the lists it doesn't work. But when I try removing the quote (which also doesn't work) and THEN go to a list and press backspace, it gets removed.

khaykov:
Thanks, @0nko ! Found the problem. I also modified logic a bit, to make it possible to remove style without actually deleting/adding any characters. This makes everything simple, but unfortunately,  I don't know how to test it, as we cant simulate press of a backspace :)

0nko:
Nice, that works great! :) As for testing, have you tried [EditText.dispatchKeyEvent()](http://stackoverflow.com/questions/4779000/how-to-send-key-event-to-an-edit-text)?

I'll merge this now and we can add the tests later.

:shipit:

PR #210, opened 2017-01-27
Cursor positioned while composing html output

### LINKED ISSUE #209: Cursor breaks emoji
### Expected
Adding text around amoji or moving the cursor around should not disrupt the emoji handling between html and visual mode

### Observed
Cursor position disrupts emoji when switching between visual and html mode.

### Reproduced
1. Set the `EXAMPLE` val in `MainActivity.kt` to `"&#x1F44D;&#x2764;"` and run the demo app. The app should display a "thumbs up" and a "heart" emoji
2. Using the device's keyboard normally, add an `a` at the start of text
3. Switch to HTML mode. Observe that the second emoji (the heart) has disappeared and instead "garbage" text `d84;` is at the end of text.

#### Tested
Nexus 5X, Android v7.1.1, with app version c166f1c

hypest:
### Fix #209 

Instead of trying to place the cursor at the html output after having outputted the spans and characters, place the cursor while traversing the input. This way, we don't need to know the output size of each element/character and we don't need to compute the output cursor position.

Known issues: undo/redo doesn't preserve the cursor position while switching to html. That's an issue prior to this PR anyway.

### Test
1. In the demo app, place the cursor anywhere inside the visual editor
2. Switch to html mode and observe the position of the cursor. It should be "preserved" staying in a position relevant to the visual mode position
3. While at it, notice that even placing the cursor around emojis and switching to html mode, emojis are preserved

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -69,6 +69,7 @@ class MainActivity : AppCompatActivity(), OnMediaOptionSelectedListener, OnReque
                 "<br>"
         private val CODE = "<code>if (value == 5) printf(value)</code><br>"
         private val IMG = "<img src=\"https://cloud.githubusercontent.com/assets/3827611/21950131/3def4804-d9b5-11e6-88e6-d7d8864392e0.png\" />"
+        private val EMOJI = "aaa&#x1F44D;&#x2764;ccc"
         private val EXAMPLE =
                 IMG +
                 HEADING +
@@ -85,7 +86,8 @@ class MainActivity : AppCompatActivity(), OnMediaOptionSelectedListener, OnReque
                 COMMENT_MORE +
                 COMMENT_PAGE +
                 CODE +
-                UNKNOWN
+                UNKNOWN +
+                EMOJI
     }
 
     private val MEDIA_CAMERA_PHOTO_PERMISSION_REQUEST_CODE: Int = 1001
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -46,7 +46,7 @@ class MainActivity : AppCompatActivity(), OnMediaOptionSelectedListener, OnReque
         private val ITALIC = "<i>Italic</i><br>"
         private val UNDERLINE = "<u>Underline</u><br>"
         private val STRIKETHROUGH = "<s class=\"test\">Strikethrough</s><br>" // <s> or <strike> or <del>
-        private val ORDERED = "<ol><li>Ordered</li></ol>"
+        private val ORDERED = "<ol><li>Ordered</li><li></li></ol>"
         private val UNORDERED = "<ul><li>Unordered</li><li></li></ul>"
         private val QUOTE = "<blockquote>Quote</blockquote>"
         private val LINK = "<a href=\"https://github.com/wordpress-mobile/WordPress-Aztec-Android\">Link</a><br>"
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -47,7 +47,7 @@ class MainActivity : AppCompatActivity(), OnMediaOptionSelectedListener, OnReque
         private val UNDERLINE = "<u>Underline</u><br>"
         private val STRIKETHROUGH = "<s class=\"test\">Strikethrough</s><br>" // <s> or <strike> or <del>
         private val ORDERED = "<ol><li>Ordered</li></ol>"
-        private val UNORDERED = "<ul><li>Unordered</li></ul>"
+        private val UNORDERED = "<ul><li>Unordered</li><li></li></ul>"
         private val QUOTE = "<blockquote>Quote</blockquote>"
         private val LINK = "<a href=\"https://github.com/wordpress-mobile/WordPress-Aztec-Android\">Link</a><br>"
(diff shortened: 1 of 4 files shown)


theck13:
This also fixes https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/178.

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -643,6 +585,19 @@ class AztecParser {
             }
             i++
         }
+
+        if (nl == 0 && text.length > i && text[i] == '\n') {
+            consumeCursorIfThere(out, text, i)
+        }
+    }
+
+    private fun consumeCursorIfThere(out: StringBuilder, text: CharSequence, position: Int) {
```
theck13: The name `consumeCursorIfThere` is a little ambiguous.  Adding comments describing the function and/or renaming the `There` part of the function would be helpful.

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -314,13 +317,6 @@ class AztecParser {
                 out.append("<li>")
             }
 
-            //special case for when cursor might be in empty list item
```
theck13: Toggling visual and HTML with an empty list moves the cursor to the beginning of the text.  Removing this code might be the reason.  Enter the HTML below, place the cursor between the `<li>` and `</li>` tags, and toggle between views for an example.

```
Text
<ul>
  <li></li>
</ul>
```

hypest: Good catch! Fixed with 49cc43c

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
hypest: Right, I see what you mean. Renamed it and added some comments with c097296.

hypest:
Ready for another pass @theck13 , thanks!

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -591,15 +591,27 @@ class AztecParser {
...
+            consumeCursorIfInInput(out, text, i)
         }
     }
 
-    private fun consumeCursorIfThere(out: StringBuilder, text: CharSequence, position: Int) {
+    /**
+     * Append a cursor to the output string if input string has one at the specified position.
+     * Cursor is removed from the input if found at that position.
+     *
+     * The algorithm that uses this function goes like this: While traversing the input (spannable) string and producing
```
theck13: Beginning with `The algorithm that uses this function goes like this:` is unnecessary.

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -591,15 +591,27 @@ class AztecParser {
...
+     *   the output chunk chunk, look for the cursor span in the input string at a location before or after the chunk.
+     *   If cursor is found then remove it while appending a cursor literal to the output string. This way, the cursor
+     *   gets inserted without the need to know which position in the output string corresponds to the position in the
+     *   input string.
+     */
+    private fun consumeCursorIfInInput(out: StringBuilder, text: CharSequence, position: Int) {
```
theck13: The comments and renaming make this much more grokkable.  Thanks!

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -591,15 +591,27 @@ class AztecParser {
...
         }
     }
 
-    private fun consumeCursorIfThere(out: StringBuilder, text: CharSequence, position: Int) {
+    /**
+     * Append a cursor to the output string if input string has one at the specified position.
+     * Cursor is removed from the input if found at that position.
+     *
+     * The algorithm that uses this function goes like this: While traversing the input (spannable) string and producing
+     *   the output chunk chunk, look for the cursor span in the input string at a location before or after the chunk.
```
theck13: Is `chunk chunk` a typo?

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -47,7 +47,7 @@ class MainActivity : AppCompatActivity(), OnMediaOptionSelectedListener, OnReque
         private val UNDERLINE = "<u>Underline</u><br>"
         private val STRIKETHROUGH = "<s class=\"test\">Strikethrough</s><br>" // <s> or <strike> or <del>
         private val ORDERED = "<ol><li>Ordered</li></ol>"
```
theck13: Adding an empty list item to the ordered list might help illustrate the difference in lists.

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
hypest: Hmm, not sure what you mean. That "intro" is for making clear that it's not the commented function that performs the algorithm but rather, the caller(s) of it. For example, is not `consumeCursorIfInInput` that traverses the input string. That said, perfectly open to revise this to something different that has the same meaning. Let me know! Thanks!

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
hypest: It kinda is yeah, it messes a "by" between the "chunk"s. Will fix it. Thanks!

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
hypest: I'll add an empty list item to the `ol` as well for completeness, thanks!

PR #223, opened 2017-02-03
Issue/162 adding padding to block elements

### LINKED ISSUE #162: Add more space around heading and block elements
### Expected
image

### Observed
image

khaykov:
### Fixes #162 

Adds more vertical padding (+8dp) to block elements and headings.

The only approach that proven to work was using LineHeightSpan, but the drawback to it is that cursor height is also affected. With moderate padding it's not so noticeable, but maybe it's something we should address in the future. 

padding


0nko:
Cool! This will make posts look much better 👍. But I noticed when you add a character at the end of a heading and then remove it, the padding gets removed.

khaykov:
Looks like I messed up with heading's constructor a bit. Should be fixed now!

0nko:
Nice PR. :shipit:

PR #226, opened 2017-02-06
Issue/218 uploading images

### LINKED ISSUE #218: Support uploading of images added to the editor
Like the current wpandroid visual editor, Aztec needs to support uploading of images.

hypest:
### Fix #218 

Delegates the media menu to the client while adding extra drawing capabilities to AztecMediaSpan to let it display overlays on top of the image. This way for example, the client can mark the image with extra info to denote that the image may be uploading or has failed to upload.

### Test

1. Add a photo from device's gallery
2. Watch the photo being added, tinted with a grey overlay and a progress bar at its top being advancing every 2 seconds. That's a simulation of the "upload" process. In a real app, the client should use its own machinery to perform the actual upload.


0nko:
The upload progress bar is awesome! 🥇 But I've noticed some strange behavior when typing next to images. The text typed before the image pushes it out of the screen and the text after it is displayed over it:

image

The version in `develop` doesn't do that:

image

0nko:
Another problem, that's probably related to the issue above is that when you put 2 images right next to each other, they overlap:

image

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -605,8 +611,8 @@ class AztecText : EditText, TextWatcher {
              *
              * https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-baseline-grids
              */
-            val width = DisplayUtils.getDisplayPixelWidth(context) - DisplayUtils.dpToPx(context, 32)
-            imageGetter?.loadImage(it.source, callbacks, width)
+            val width = context.resources.displayMetrics.widthPixels - DisplayUtils.dpToPx(context, 32)
```
hypest: I assume you mean the change from `it.source` to `it.getSource()` here. If so, it's not the same as now, the `src` attribute lives inside the Attributes member, not exposed directly as a property anymore.

OTOH, if you mean the change from `DisplayUtils.getDisplayPixelWidth(context)` to `context.resources.displayMetrics.widthPixels` then I'm not sure why is displayed in the diff but it's not part of this PR. Instead, it's part of https://github.com/wordpress-mobile/WordPress-Aztec-Android/commit/52b9cadfb96f07133f290a8c1002823861f386e7#diff-7ca71cd1f009ab8438e0901205001b14R569

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -773,7 +779,8 @@ class AztecText : EditText, TextWatcher {
                 val textToPaste = clip.getItemAt(i).coerceToText(context)
 
                 val builder = SpannableStringBuilder()
-                builder.append(parser.fromHtml(Format.clearFormatting(textToPaste.toString()), context).trim())
+                builder.append(parser.fromHtml(Format.clearFormatting(textToPaste.toString()),onMediaTappedListener,
```
hypest: Good eye! Fixed with 0505142.

hypest:
Hey @0nko , thanks for picking this up!

26c5ac6 has the fix for the image problem you saw. Both the text overlapping and the image overlapping were happening due to the same bug.

Ready for another pass, thanks!

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -888,4 +895,81 @@ class AztecText : EditText, TextWatcher {
         }
     }
 
+    fun insertMedia(drawable: Drawable?, attributes: Attributes) {
+        lineBlockFormatter.insertMedia(drawable, attributes, onMediaTappedListener)
+    }
+
+    fun removeMedia(attributePredicate: AttributePredicate) {
+        text.getSpans(0, 999999999, AztecMediaSpan::class.java).forEach {
```
hypest: Ah, no special reason other than one less method call and laziness that stuck. Changed with d057ca8!

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
0nko: Yeah, I meant `DisplayUtils.getDisplayPixelWidth(context)`

hypest:
Hey @0nko FYI, added the implementation for removal of failed uploads on html toggle with 5f4e56b.

0nko:
Nice work, Stefanos! :shipit:

PR #227, opened 2017-02-07
Issue/225 allow images in block elements

### LINKED ISSUE #225: Support images inside block elements
Currently images don't work inside block elements (lists, quotes, etc.).

khaykov:
### Fixes #225

Inserting images does not break block elements anymore.
Also, Instead of splitting them, like before, we are inserting images inside them instead.

In addition, Images are now allowed inside headings.

PS: I noticed rare visual rendering issues - heigh of line might fluctuate when images are inserted.


0nko:
When you insert an image that's too wide it scrolls the editor horizontally, because the list items are indented.. This is a side effect of making an image as wide as possible but we forget they could be offset.

0nko:
I'll merge this and create a separate issue for the width adjustment, since the smaller images work and it's just a visual thing for large ones. :shipit:

Issue #231, opened 2017-02-08
Can't apply inline style at new line in block element


khaykov:
Steps to reproduce:

1. Make a list or quote.
2. Press enter to add new line to it.
3. Press Bold style in toolbar and start typing.
4. Notice that bold style is not applied.

#### Tested
With Nexus 5x on android 6.0.0

PR #232, opened 2017-02-08
Issue/220 fix highlighting of shared styles

### LINKED ISSUE #220: Toolbar highlight of shared inline styles sometimes does not work
### Expected
When selecting multiline text with shared style it should be highlighted in the toolbar.

### Observed
When styling text in certain way shared style is not highlighted in the toolbar.

### Reproduced
1. Load demo app with couple of inline styles each on new line, like `BOLD + ITALIC + UNDERLINE +STRIKETHROUGH`
2. Select each individual line and apply style, Bold, for example.
3. Select all the lines and confirm that Bold is not highlighted.

Now, if you select whole text, and apply bold to it, it will be properly highlighted in the toolbar.

khaykov:
### Fixes #220 and #231

To test: 

#220
1. Load demo app with couple of inline styles each on new line, like BOLD + ITALIC + UNDERLINE +STRIKETHROUGH
2. Select each individual line and apply style, Bold, for example.
3.  Select all the lines and confirm that Bold is highlighted in toolbar.

#231
1. Make a list or quote.
2. Press enter to add new line to it.
3. Press Bold style in toolbar and start typing.
4. Notice that bold style is properly applied.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -174,6 +174,7 @@ class AztecText : EditText, TextWatcher {
             var consumeKeyEvent = false
             history.beforeTextChanged(toFormattedHtml())
             if (keyCode == KeyEvent.KEYCODE_DEL && event.action == KeyEvent.ACTION_DOWN) {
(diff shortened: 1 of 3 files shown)


0nko:
:shipit:

PR #234, opened 2017-02-08
Only click the Clickable span if inside it

### LINKED ISSUE #229: Unable to place cursor next to an image
Tapping next to an image immediately opens the image detail screen, which blocks users from typing. A possible solution could be showing an overlay first and only after tapping on the image again should the detail screen be opened.

hypest:
### Fix #229 

Restrict the triggering of ClickableSpan's click method to only happen if the tap happened inside its range.

Note: the implementation is accurate for the horizontal detection of the span's range but not the vertical.

### Test
1. Using the demo app, tap _around_ an image
2. The cursor should move without triggering the "media tapped" event (i.e. the "Media tapped!" toast won't show)

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -202,6 +202,8 @@ class MainActivity : AppCompatActivity(), OnRequestPermissionsResultCallback, Vi
         aztec.imageGetter = PicassoImageLoader(this, aztec)
 //        aztec.imageGetter = GlideImageLoader(this)
 
+        aztec.setOnMediaTappedListener(this)
+
         source = findViewById(R.id.source) as SourceViewEditText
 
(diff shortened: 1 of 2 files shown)


0nko:
Hm, it doesn't seem to work very well. Most of the time it doesn't detect a tap on an image. Tried it both with the demo app and WP.

hypest:
@0nko confirmed! It seems that tapping on the left half of the image doesn't register as image tap. Will look into it. Thanks!

0nko:
Nope, still doesn't work. Tried it on API 23 emulator, as well. Are you testing it with the demo app? I can't get it to show the toast even once.

hypest:
Sorry for the trouble @0nko , it should work now. Fixed with 2501e61. I was testing it with the wpandroid app directly. The demo app actually had a bug introduced in a different PR. Please give it a try again. Thanks!

0nko:
No worries. This works! :shipit:

PR #240, opened 2017-02-09
Issue/184 add keyboard shortcuts

theck13:
### Fix
Add keyboard shortcuts as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/184.  The shortcuts available are displayed in a dialog shown by pressing `Alt`+`Ctrl`+`H`.  The dialog can not be shown by holding the meta key (i.e. Search or Windows or Command) since the operating system overrides that key to launch search.  Android also overrides the meta key and `/` combination to show system-level shortcuts.  The `Alt`+`Ctrl`+`H` combination was chosen to match Calypso.  On the bright side, the standard editor shortcuts (i.e. copy, cut, paste, and select all) were included for free since `AztecText` extends `EditText`.

### Test
0. Connect external keyboard.
1. Open demonstration app.
2. Press `Alt`+`Ctrl`+`H`.
3. Notice keyboard shortcuts legend.
4. Try keyboard shortcuts.


khaykov:
Look's like I cant really test this one with Genymotion/Emulator, so will have to pass on it, as I don't have a HW keyboard.

0nko:
Yeah, neither do I.. :/

0nko:
But @astralbodies has Pixel C ;).

theck13:
@astralbodies and @hypest, do either of you have an external keyboard?

hypest:
I have a tablet+keyboard so I can give this a try. Will report back.

hypest:
OK, gave it a go with a BT tablet case keyboard and works OK, good job!

Not sure if this has been code reviewed already but, any particular reason the keyboard mapping is handled by the demo app instead of the Aztec lib? I mean, the toolbar itself is part of Aztec so, i'd expect the shortcuts to be handled by Aztec as well. Any insight? Thanks!

theck13:
> any particular reason the keyboard mapping is handled by the demo app instead of the Aztec lib?

I considered handling the keyboard shortcuts within the library rather than the app.  If they are in the library, they would be handled in the textbox <b><i>or</I></b> the toolbar.  That means if the user has focus on the textbox, the toolbar wouldn't recognize the keyboard.  Similarly, if the user has focus on the toolbar, the textbox wouldn't recognize the keyboard.  In other words, it's possible the library wouldn't recognize the keyboard depending on the view in focus.  If the keyboard shortcuts are handled in the app, they are recognized regardless of focus.

That being said, I'm up for trying a different implementation if you have another idea.

hypest:
Ah, thanks for the insight @theck13! Haven't tested it but I think the toolbar can implement the mapping and the editor can ask the toolbar to consume the key before trying to consume it itself. So, both will implement the inteception but the editor will just delegate the key. Does that make sense?

astralbodies:
Let me know if you'd still like me to test out this PR!

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -86,6 +234,16 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
                 highlightAppliedStyles(selStart, selEnd)
             }
         })
+        this.editor!!.setOnKeyListener {
```
hypest: Hey @theck13 , thanks for the change from `MainActivity` to Aztec/Formatbar.

My only concern atm is that this listener set is kinda "magic", in the sense that the toolbar view is essentially "stealing" the key click from another view (the Aztec EditText's in this case), see https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/View.java#9952. I think this is less readable and easier to break.

May I suggest a slight variation: 
* Introduce a `setToolbar(toolbar: AztecToolbar)` method in AztecText. That will be used by MainActivity to offer AztecText a reference to the toolbar
* Override `AztecText.onKeyUp()` and call `toolbar.onKeyUp` directly if a toolbar is set. Return true if the toolbar handled the event, otherwise just call the super.

The idea with this variation is to be able to see the key override while reading the AztecText code or when trying to override the keyUp event. Let me know what you think, thanks!

hypest: Hey @theck13 , thanks for the change from MainActivity to Aztec/Formatbar.

My only concern atm is that this listener set is kinda "magic", in the sense that the toolbar view is essentially "stealing" the key click from another view (the Aztec EditText's in this case), see https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/View.java#9952. I think this is less readable and easier to break.

May I suggest a slight variation:
* Introduce a setToolbar(toolbar: AztecToolbar) method in AztecText. That will be used by MainActivity to offer AztecText a reference to the toolbar
* Override AztecText.onKeyUp() and call toolbar.onKeyUp directly if a toolbar is set. Return true if the toolbar handled the event, otherwise just call the super.

The idea with this variation is to be able to see the key override while reading the AztecText code or when trying to override the keyUp event. Let me know what you think, thanks!

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -319,6 +322,14 @@ class AztecText : EditText, TextWatcher {
         return super.onKeyPreIme(keyCode, event)
     }
 
+    override fun onKeyUp(keyCode: Int, keyEvent: KeyEvent): Boolean {
+        if (formatToolbar.onKeyUp(keyCode, keyEvent)) {
```
hypest: There is no guarantee that the client will call `setToolbar()` (or not pass `null`) so, we prolly should check it here. It might be simpler to just declare `formatToolbar` as `AztecToolbar?`.

Issue #167, opened 2017-01-05
Allow for block editing of "unknown HTML"


0nko:
The "unknown HTML" span should probably be on its own line and take up the entire width of the screen. Tapping on it should display the hidden HTML and allow editing.

This could be a useful feature in a subsequent release in the future.

PR #241, opened 2017-02-09
Issue/167 block editing

0nko:
Adresses #167.

When you click on an unknown HTML question mark picture a dialog will pop up and you can edit the unknown HTML and save it.

I haven't made the unknown HTML a block span because that would create problems if it was embedded inside another block element and I don't see any issues as it is right now.


khaykov:
I noticed a bit weird behavior:
Open hidden dialog with hidden html in demo app.
Add `<b>` tags around "Menu", so the html will look like this:
`<iframe class="classic"><b>Menu</b></iframe>`
Press save, and open hidden tag again.
`<b>` get's escaped, and html looks like this:
`<iframe class="classic">&lt;b&gt;Menu&lt;/b&gt;</iframe>`

Another minor issue is that the long code get's clipped at the bottom of the dialog:
screenshot-2017-02-10_15 41 41 543

khaykov:
Cursor related issue:

Open demo app.
Go to source code, copy everything.
Return to visual mode.
Open Hidden html tag and replace "Menu" with copied code.
Click save.
Notice "aztec_cursor" appearing at the end of the editor.

0nko:
At last, all of the issues should be fixed. Ready for another pass.

0nko:
OK, now :P.

khaykov:
Thanks for the changes, everything works nicely now!
I noticed only one minor thing - when there is a lot of content in dialog, the keyboard overlaps it, and bottom of dialog becomes hidden. It's especially noticeable in landscape mode.

Maybe setting soft input mode of dialog window to something like `SOFT_INPUT_ADJUST_RESIZE` will help?

0nko:
Thanks for the tip!

khaykov:
:shipit: 🎉

PR #242, opened 2017-02-10
Issue/fix visual glicthes with block elements

khaykov:
This PR addresses visual glitches when adding/removing characters.

I was able to fix all of them except for one - when you have a line of plain text above block element, add/remove character from that line, and the line with block element moves up/down a bit.

Same as @0nko , I traced this down to DynamicLayout reflow method, and this happens because we are using `LineHeightSpan` decorated with `UpdateLayout` interface for block element paddings.

At this point I don't know how to fix it, except for coming with other way of implementing paddings. But the issue is minor, so I think we could address it a bit later, when more critical issues are fixed.

In addition, I changed the drawing logic of lists indicator, to base them on `baseline` value, instead of `bottom` value, which makes more sense, and does not requires any adjustment's that come from variable line height.


0nko:
The lists look/work so much better! The same movement happens still happens for other lines, not just when you have a block element below it. For example, start the app with the sample text and add something to "Bold" or "Italic". The text below moves up a bit. 

I'm going to merge this so that it's included in the next release and we can address the other movement in a subsequent PR.

Nice work ;). :shipit:

PR #246, opened 2017-02-11
Issue/245 image in paragraph

### LINKED ISSUE #245: Image inside a paragraph breaks the HTML
1. Start with `<p>hello</p>` in HTML mode
2. Switch to visual mode
3. Add an image
4. Notice the text is modified to `hell` and the and `>` is added after the image

0nko:
Fixes #245.

Can someone have a quick look so that it can be included in the release? Thanks!


aztec/src/main/java/org/wordpress/aztec/Html.java
```diff
@@ -751,28 +751,6 @@ private static void end(SpannableStringBuilder text, TextFormat textFormat) {
         }
     }
 
-//<<<<<<< HEAD
```
theck13: Good idea.  We can remove [this unused code](https://github.com/wordpress-mobile/WordPress-Aztec-Android/pull/246/files#diff-7177c376b871e9e1aa2e4983072f0705L164) too.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
```diff
@@ -273,32 +270,31 @@ class LineBlockFormatter(editor: AztecText, val headerStyle: LineBlockFormatter.
 
     fun insertMedia(drawable: Drawable?, attributes: Attributes, onMediaTappedListener: OnMediaTappedListener?) {
         val span = AztecMediaSpan(editor.context, drawable, attributes, onMediaTappedListener)
-        val html = span.getHtml();
 
-        val isAddingMediaToEndOfBlockElement = editableText.getSpans(selectionStart, selectionEnd, AztecBlockSpan::class.java)
-                .any {
-                    selectionStart == editableText.getSpanEnd(it)
-                }
+        val spanAtEndOfWhichMediaIsAdded = editableText.getSpans(selectionStart, selectionEnd, AztecBlockSpan::class.java)
```
theck13: It looks like `spanAtEndOfWhichMediaIsAdded` was renamed similar to `spanAtStartOfWhichMediaIsAdded`.  What about naming them `spanBeforeMedia` and `spanAfterMedia`, respectively?  That's easier to grok.

0nko: Oh yeah :) makes sense..

aztec/src/main/java/org/wordpress/aztec/Html.java
0nko: Good catch, time to get rid of it.

0nko:
Thanks for the quick review @theck13.

PR #248, opened 2017-02-13
Issue/221 fixing empty blockquote parsing

### LINKED ISSUE #221: Empty blockquote is removed during html2span parsing
### Expected
Empty blockquote is not removed during html2span parsing.

### Reproduced
1. Load demo app with empty blockquote tag - `<blockquote></blockquote>`, either standalone or within text.
2. Notice that quote is missing in visual editor.

khaykov:
### Fix #221

Same as with lists, we are adding newline to zero-width block spans to preserve them.


0nko:
:shipit:

PR #249, opened 2017-02-13
Fixin visibility of cursor in quote

khaykov:
Similar to @theck13 's PR (#244) adds an alpha channel to quote background color, to make cursor visible.


0nko:
:shipit:

PR #252, opened 2017-02-14
Issue/235 enable disable actions

theck13:
### Fix
Update the state of the redo and undo actions as described in https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/235.

### Test
1. Open demonstration app.
2. Notice ***Redo*** is disabled and ***Undo*** are disabled.
3. Type one character.
4. Notice ***Redo*** is disabled and ***Undo*** is enabled.
5. Type nine or more characters.
6. Notice ***Redo*** is disabled and ***Undo*** is enabled.
7. Tap ***Undo*** action button one time.
8. Notice ***Redo*** is enabled and ***Undo*** is enabled.
9. Tap ***Undo*** action button nine times.
10. Notice ***Redo*** is enabled and ***Undo*** is disabled.
11. Tap ***Redo*** action button one time.
12. Notice ***Redo*** is enabled and ***Undo*** is enabled.
13. Tap ***Redo*** action button nine times.
14. Notice ***Redo*** is disabled and ***Undo*** is enabled.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -27,6 +27,7 @@ import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.PermissionUtils
 import org.wordpress.android.util.ToastUtils
 import org.wordpress.aztec.AztecText
+import org.wordpress.aztec.HistoryListener
 import org.wordpress.aztec.picassoloader.PicassoImageLoader
 import org.wordpress.aztec.source.SourceViewEditText
 import org.wordpress.aztec.toolbar.AztecToolbar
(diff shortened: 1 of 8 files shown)


0nko:
Works well, but the the ripple effect on the buttons is gone for some reason, so it looks like nothing's happening if you can't see the text changing. I suspect it's because `invalidateOptionsMenu()` is called immediately after tapping.

theck13:
Yep, it's because of `invalidateOptionsMenu()`.  I thought about doing something like this, but then the ripple would show part of the time and not show up part of the time which might be more confusing than not showing ever.

```
    override fun onRedoEnabled(state: Boolean) {
        if (isRedoEnabled != state) {
            isRedoEnabled = state
            invalidateOptionsMenu()
        }
    }
```

If you have an idea of how to work around `invalidateOptionsMenu()`, I'm all for it.

0nko:
I think this could work:
```
Handler().postDelayed(Runnable { invalidateOptionsMenu() }, 250)
```

0nko:
:shipit:

PR #254, opened 2017-02-14
Issue/maintain toolbar state

theck13:
### Fix
Save and restore toolbar state (i.e enabled or disabled) when rotating the device.

### Test
1. Open demonstration app.
2. Notice toolbar is enabled.
3. Rotate device.
4. Notice toolbar is enabled.
5. Tap ***HTML*** format button.
6. Notice toolbar is disabled.
7. Rotate device.
8. Notice toolbar is disabled.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
@@ -1,6 +1,8 @@
 package org.wordpress.aztec.toolbar
 
 import android.content.Context
+import android.os.Bundle
+import android.os.Parcelable
 import android.util.AttributeSet
 import android.view.MenuItem
 import android.view.View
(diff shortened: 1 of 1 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -73,6 +75,22 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
         }
     }
 
+    override fun onRestoreInstanceState(state: Parcelable?) {
+        val savedState = state as SourceViewEditText.SavedState
```
khaykov: So instead of using internal `SavedState` from `SourceViewEditText` what do you think about moving it to external class, like `AztecSavedState` and using it here, in `SourceViewEditText` and `AztecText`?

theck13: That's fine with me.  Go for it.

khaykov:
:shipit:

PR #270, opened 2017-02-21
HR tag support

### LINKED ISSUE #261: Add support for the horizontal rule ("hr" tag)
### Expected

The Calypso editor include a horizontal rule (the HTML `hr` tag) as a button in the editor's toolbar. I expect elements from the web editor to be visually represented in Aztec.

### Observed

The `hr` tag is represented with a question mark in Aztec:

android-horizontal-line

#### Tested

0nko:
Resolves #261.


khaykov:
I found couple of issues:

- Load demo app with single `<hr>` tag.
- Put cursor at the beginning of line and type something.
- Typed text will appear above the line.
- Now try to place cursor at the end of the text, and notice that it places itself to the position before the last character.

Additionally, try to select the text from the example above and notice that the line is also selected (same behavior with text added directly after line).

Also, If you add text to the end/beginning of the line as described above, and use suggestion from keyboard to correct inputed text, the line will be gone.

The problems are gone when you switch to html and back.

0nko:
All of the problems should be resolved now. Can you give it another look, please?

khaykov:
I found couple of issues with the line. I believe they will apply to other similar elements, and most of them do come from the ability to place a cursor in front/after the span.

- Put cursor in front or behind the line, activate any inline style, start typing, notice that style is not applied.
- Put cursor in front or behind the line, apply list. Notice that line inside list was rendered correctly. Switching to html shows that line is not inside list, and switching back to visual mode renders list below line.

This one only happens if there is text before and after the line:
- Put cursor in front or behind the line, apply quote. Notice that visually nothing happened. Switching to html and back makes quote appear, but the line is not inside it.

0nko:
>Put cursor in front or behind the line, activate any inline style, start typing, notice that style is not applied.

This is fixed.

> Put cursor in front or behind the line, apply list. Notice that line inside list was rendered correctly. Switching to html shows that line is not inside list, and switching back to visual mode renders list below line.

This is also fixed.

> Put cursor in front or behind the line, apply quote. Notice that visually nothing happened. Switching to html and back makes quote appear, but the line is not inside it.

This seems to be a [separate issue](https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/298).

While working on `<hr>` rendering I managed to fix the nested lists 🙌:

image

khaykov:
Awesome job, @0nko ! :shipit: 🎉

PR #277, opened 2017-03-02
issue/toolbar-media-mode

nbradbury:
Adds "media mode" to the toolbar which will be used with the new photo chooser in WP Android. This highlights the media button and disables all the others.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
@@ -26,6 +26,7 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
     private var headingMenu: PopupMenu? = null
     private var sourceEditor: SourceViewEditText? = null
     private var dialogShortcuts: AlertDialog? = null
+    private var isMediaModeEnabled: Boolean = false;
 
     constructor(context: Context) : super(context) {
         initView()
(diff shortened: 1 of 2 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -408,6 +411,21 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
         }
     }
 
+    fun isMediaModeEnabled(): Boolean {
```
theck13: Will this method be used in a client app (e.g. WordPress) in the future?  I'm just wondering since it isn't used in the library or demonstration app.

aztec/src/main/res/drawable/format_bar_button_media_selector.xml
```diff
@@ -6,6 +6,7 @@
     <item android:drawable="@drawable/format_bar_button_media_disabled" android:state_enabled="false" />
     <item android:drawable="@drawable/format_bar_button_media_highlighted" android:state_pressed="true" />
     <item android:drawable="@drawable/format_bar_button_media_highlighted" android:state_focused="true" />
+    <item android:drawable="@drawable/format_bar_button_media_highlighted" android:state_checked="true" />
```
theck13: Can we move the `state_checked` entry between the `state_enabled` and `state_pressed` entries to mimic the other selector files (e.g. [`format_bar_button_bold_selector`](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/aztec/src/main/res/drawable/format_bar_button_bold_selector.xml))?  That way we can avoid potential bugs from differing behavior.

nbradbury: > Can we move the state_checked entry between the state_enabled and state_pressed entries

Addressed in 7b50e25

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
nbradbury: I'm not sure if this will be used in WP Android but thought it should be added in case it's needed. I can remove it if you like then add it in a separate PR if and when it's needed.

theck13:
:shipit:

Issue #264, opened 2017-02-17
Double space to end sentence inserts a space before the period.


zandyring:
### Expected
I expect a double space to insert a period at the end of the previous word. `abc.`

### Observed
A double space inserts a space+period after the last word. `abc .`

### Reproduced
1. Type a word in the content area
2. Tap a space twice rapidly
3. You will see a space+period at the end of the last typed word.

#### Tested
Moto G (3rd Gen) on Android 6.0 with alpha-35

rachelmcr:
I see the same behavior on my Nexus 9, Android 7.1.1, WPAndroid alpha-35.

khaykov:
Same as with #266, this seems like a feature exclusive to of iOS hybrid editor.
It would certainly be nice to have, with the ability to turn it on/off with a flag.

theck13:
This is another bug which various by Android version.  On pre-Lollipop versions (API <= 19), tapping double space after "Test" produces "Text. " as expected.  On Lollipop and after (API > 19), tapping double space after "Text" produces "Text ." which is unexpected.

Issue #251, opened 2017-02-13
Heading inside list inside quote keeps adding newlines


0nko:
### Expected

Having a heading inside list inside quote should stay the same when switching between HTML and visual mode.

### Observed

It doesn't. Extra newlines are added to the quote.

![image](https://cloud.githubusercontent.com/assets/1522856/22904957/a57caf56-f23e-11e6-9b7b-24222d2c6340.png)

### Reproduced
1. Type some word
2. Apply quote to it
3. Apply bullet list to it
4. Apply heading 1 to it
5. Switch to HTML mode
6. Switch back to visual mode
7. Observe the extra newline

#### Tested
Emulator with Android API 18

0nko:
Fixed by #278.

Issue #247, opened 2017-02-12
Text Inside Paragraph Moves Outside Block Format When Toggling HTML


theck13:
### Expected
Text remains inside block format when toggling HTML view.

### Observed
Text is moved outside block format when toggling HTML view.  This produces unformatted text above an empty list (for ordered list and unordered list) or removes the block format completed (for quote).  See the screenshots below for examples.

![paragraph](https://cloud.githubusercontent.com/assets/3827611/24050862/d5565d7e-0af5-11e7-95bc-bcf8e209887f.png)

### Reproduced
1. Tap ***HTML*** format button.
2. Enter `<p>hello</p>`.
3. Tap ***HTML*** format button.
4. Tap block format button (i.e. ***Ordered List***, ***Unordered List***, or ***Quote***).
5. Notice `hello` is inside block format.
6. Tap ***HTML*** format button.
7. Notice `hello` is outside block format.
a. If ordered list, empty list is below `hello`.
b. If unordered list, empty list is below `hello`.
c. If quote, format is removed completely.

#### Tested
Google Pixel on Android 7.1.1 with AztecDemo 1.0

hypest:
Fixed with #278 . But, a new issue was introduced in the process: https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/304

Issue #194, opened 2017-01-20
Closing list with double-enter adds extra newline


0nko:
Pressing enter twice to close a list introduces an extra newline that is not supposed to be there.

PR #278, opened 2017-03-06
Paragraph spans

### LINKED ISSUE #125: Nested lists show multiple bullets
screenshot_1478255998

A list within a list like 
```
<ul>
   <li>Item
   <ul>
      <li>Inner item</li>
   </ul>
   </li>
</ul>
```
should probably only show the innermost bullet (but still keep the indenting).

### LINKED ISSUE #126: Nested lists get removed
When going from visual mode to HTML, the inner nested lists get removed.

### Expected

```
<ul>
   <li>Item
   <ul>
      <li>Inner item</li>
   </ul>
   </li>
</ul>
```

### Observed

```
<ul>
   <li>Item</li>
</ul>Inner item
```

### Reproduced
1. Start with a nested list inside a list
2. Switch to visual mode
3. Switch back to HTML mode and notice the change

hypest:
### Fix #125, #126, #170, #188, #194, #247 (but with a new issue), #251, #264

🎉 

This PR introduces a different method of marking and handling block elements. Here's a list of the changes it introduces:
* Mark the whole content of a block with the span
* Set the block spans with the `SPAN_PARAGRAPH` flag and never change it. See the [Spanned.java source](https://github.com/android/platform_frameworks_base/blob/master/core/java/android/text/Spanned.java#L72-L88) for details. Paragraph spans must start after a newline (or the start of buffer) and end right after a newline (or the end of the buffer). The auto-changing nature of the SPAN_PARAGRAPH span becomes tricky when text is replaced instead of just added/removed and a few classes are dedicated to handling that ([ParagraphBleedAdjuster](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/experiment/paragraph-spans/aztec/src/main/kotlin/org/wordpress/aztec/watchers/ParagraphBleedAdjuster.kt), [ParagraphCollapseAdjuster](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/experiment/paragraph-spans/aztec/src/main/kotlin/org/wordpress/aztec/watchers/ParagraphCollapseAdjuster.kt), [ParagraphCollapseRemover](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/experiment/paragraph-spans/aztec/src/main/kotlin/org/wordpress/aztec/watchers/ParagraphCollapseRemover.kt))
* Each block element/span has a `nestingLevel` field, corresponding to its nesting level in html
* Block spans are sorted by position and nesting and then exported recursively to html
* Code that performs each block element's behaviour (list, listitem, heading, quote) runs "in parallel" to the other block elements to enforce a separation of concerns between the various types of block elements. AztecText's text mutation (like deleting the double-newline to close a quote) happens only after all block handlers have finished their processing.
* Listitem behaviour handling is implemented in a separate handler than the List itself
* Double-enter at block's end closes only the inner block when nested. This allows editing in the outer block to continue.
* Complex conditionals are split into series of simpler conditionals that return early in order to make the algorithm more readable. [Example](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/experiment/paragraph-spans/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L172-L209).

Known issues:
* Text is now never empty (there's a ZWJ if buffer empty) so, hint text is never shown. We can prolly tackle that in a separate PR.
* Nested blockquotes inside lists don't show the bullet/number of the list item they're on

This is a fairly big and broad PR, sorry for that :(. The changes are too close to the core to keep it small though. I'm open to ideas on how to break it down to smaller PRs if needed :). Thanks!


theck13:
Add https://github.com/wordpress-mobile/WordPress-Aztec-Android/issues/264 to the list of fixes.  Thanks, @hypest!

hypest:
Added a couple of commits (76b356e, 1a438af) that fix an issue where closing a nested list didn't properly close the parent list item.

Also added a fix for #125 (Nested lists show multiple bullets) with commits f068746 and 8ec075c.

hypest:
Fixed (with 6981223) a rather "invisible" regression that happened with 56f9427.

0nko:
Awesome work, @hypest! 
:shipit:

PR #279, opened 2017-03-06
Drop Merriweather font, use default serif and monospace fonts

maxme:
Drop Merriweather font, use default serif and monospace fonts

Related to https://github.com/wordpress-mobile/WordPress-Android/pull/5324


0nko:
:shipit:

PR #291, opened 2017-03-17
Different headings in consecutive lines

### LINKED ISSUE #289: Adding multiple headings produces strange behavior
### Reproduced
1. Tap on Heading 1 in toolbar menu
2. Type a word
3. Press enter
4. Tap on Heading 2
5. Notice the first heading changes in size
6. Switch to HTML
7. Notice there are two H1s, one being empty

#### Tested
Emulator on 7.1.1

hypest:
### Fix #289 

Correct check of whether any surrounding headings are of the same type. Also, better calculation of the boundaries to set the heading span.

### Test
1. Empty the text and tap on Heading 1 in toolbar menu
2. Type a word
3. Press enter
4. Tap on Heading 2
5. Notice the first heading doesn't change in size
6. Switch to HTML
7. Notice there are two heading styles, one following the other (sisters, not parent-child relationship)


0nko:
:shipit:

PR #292, opened 2017-03-17
Quote gets removed

### LINKED ISSUE #288: Having quote & list at the same time results in quote removal
### Reproduced
1. Start with the demo text
2. Tap the word inside blockquote and apply a bullet list
3. Switch to HTML
4. Switch back to visual
5. Notice the blockquote style is lost

#### Tested
Emulator on 7.1.1

hypest:
### Fix #288 

Expanding a block to neighbouring same-type blocks should only happen if neighbours are at the same nesting level.

### Test
1. Start with the demo text
2. Tap the word inside blockquote and apply a bullet list
3. Switch to HTML
4. Switch back to visual
5. Notice the blockquote style is still there and a bullet list is nested inside it


hypest:
There's an not-related issue where an extra <br> is inserted between the list and the blockquote when going back and forth to html. I'll fix this in this PR.

0nko:
When you apply the bullet list to the quote, the list appears inside it. When you switch to HTML and switch back, the blockquote is now visually inside the list.

hypest:
Hey @0nko , thanks for the review!

> When you apply the bullet list to the quote, the list appears inside it. When you switch to HTML and switch back, the blockquote is now visually inside the list.

I've investigated that issue and it seems not related to this PR. You can encounter that issue in current develop as well if you put add a nested UnorderedList inside the quote via the demo app's source code example directly.

While investigating though, I found that the issue is caused by the old arch-enemy, the messed up order returned by `getSpans()`. The visual rendering of the spans follows the `getSpans()` order and in the case you mention the list span ends up before the quote span, causing the list bullet to be rendered before/outside the quote :(.

All in all, I'd suggest we don't block this PR on this issue.

Beyond that, may I ask you to do some more testing of the PR in regards to the `<hr>` support since I had to make some changes to the same lines of code that `<hr>` depends on? Thanks in advance!

0nko:
👍 :shipit:

PR #302, opened 2017-04-03
Issue/293 editing special comments

### LINKED ISSUE #293: Weird cursor behavior around more & page special comments
Same problems as described for `<hr>` in #270:

```
I found couple of issues:

Load demo app with single <hr> tag.
Put cursor at the beginning of line and type something.
Typed text will appear above the line.
Now try to place cursor at the end of the text, and notice that it places itself to the position before the last character.
Additionally, try to select the text from the example above and notice that the line is also selected (same behavior with text added directly after line).

Also, If you add text to the end/beginning of the line as described above, and use suggestion from keyboard to correct inputed text, the line will be gone.

The problems are gone when you switch to html and back.
```

Once #270 is merged, the same fix can be applied to this issue.

0nko:
Fixes #293. Editing behavior around special comments is now the same as for `<hr>`. 

This PR also fixes [#5334](https://github.com/wordpress-mobile/WordPress-Android/issues/5334) in wpandroid, which was caused by an infinite loop when using `StaticLayout`.


0nko:
Hey @hypest, I've removed the `<br>` fix and removed disabling of text change listener from `FullWidthImageElementWatcher`.

hypest:
Hey @0nko ,  thanks for the changes!

I think I encounter an issue while editing around special comments though. Here are the steps:
1. In the demo app, place the cursor on the left of the "MORE" comment
2. Type a word, like "welcome"
3. Reposition the cursor at the beginning of the "MORE" comment
4. Hit backspace. That deletes the "e" from "welcome". All good up to this point.
5. Type "e". Notice that a new line is created underneath the "welcom" and holds the "e". That's counter-intuitive since backspace removed the "e", typing an "e" right after should place it back.

WDYT?

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -225,12 +225,12 @@ class AztecParser {
                 return@forEach
             }
 
-            if (!repelling && spanned[spanStart - 2] == '\n' && it is AztecBlockSpan) {
-                // there's another newline before and we're not repelling a parent so, the adjacent one is not a visual one so, return
+            if (!repelling && spanned[spanStart - 2] == '\n') {
+                // there's another newline before so, the adjacent one is not a visual one so, return
```
hypest: Minor one, let's reinstate the comment about repelling. Thanks!

aztec/src/test/kotlin/org/wordpress/aztec/AztecParserTest.kt
```diff
@@ -1024,7 +1026,8 @@ class AztecParserTest : AndroidTestCase() {
     @Throws(Exception::class)
     fun parseHtmlToSpanToHtmlLineBreakBetweenHeadings_isEqual() {
         val input = HTML_HEADING_ONE + "<br>" + HTML_HEADING_ONE
-        val span = SpannableString(mParser.fromHtml(input, null, null, context))
+        val span = SpannableStringBuilder(mParser.fromHtml(input, null, null, context))
+        mParser.syncVisualNewlinesOfBlockElements(span)
```
hypest: Are this change (along with line 1150) still needed? The tests seem to pass without it anyway. Thanks!

0nko: Good catch!

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
0nko: I'll add it back.

PR #309, opened 2017-04-08
Issue/258 fixing missing paragraphs

### LINKED ISSUE #258: Paragraphs/Line Breaks are missing
### Expected

In the Calypso editor, hitting Enter creates a paragraph break with space between paragraphs. I expect those paragraph breaks to be maintained when I open the post in Aztec.

### Observed

When I open that post in Aztec, the paragraph breaks are completely gone. There aren’t any line breaks between the different paragraphs I created in Calypso.

Calypso:

screen-shot-2017-02-15-at-12-02-12-pm

Aztec:

aztec-android-paragraphs

### Reproduced
1. Create a new post at https://wordpress.com/post
2. Create new paragraphs in the post.
3. Save or publish the post.
4. Open the post in Aztec and note that the paragraph breaks are gone.

#### Tested

khaykov:
This PR fixed #258 and adds support for input with mix of visual/html formatting.

Short story:
wpcom api return’s post where `p` and `br` tags are replaced with newlines and mixed with visual newlines (attached to block elements, for example).

To address this, we are replication "calypso style” pre/post processing of content (with some changes related to span2html parser).  `p` and `br` tags are replaced with newlines, empty span tags are removed.

Now, this brings a lot of “calypso” like editor behavior (mostly newline merging) when you switch between html and visual modes, so if you think something is wrong, check out how it works in calypso first.

I don't particularly like calypso's processing of html, so In a future, ideally, we should get the "real" post html with unparsed shortcodes, and preprocessed excerpt which will require some changes to API.


khaykov:
Added "paragraph on enter" feature. Now, the post editing experience should be really similar to calypso.
Couple of warning: calypso mode is enabled by default, for easier integration with wpandroid. All the tests except for `CalypsoFormattingTest` are not using calypso mode. 'toHtml()' method in calypso mode produces mix of HTML and newlines (which is converted by calypso backend into normal post).

0nko:
I've noticed a curious behavior. I opened a post with multiple paragraphs. Within a paragraph, each successive line had larger and larger height:

image

khaykov:
Fixed!

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -679,30 +692,69 @@ class AztecText : EditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlClickListe
...
+                val spanStart = output.getSpanStart(it)
+                val spanEnd = output.getSpanEnd(it)
+
+                if (output[spanStart] == '\n'
+                        && output.getSpans(spanEnd, spanEnd + 1, AztecParagraphStyle::class.java).filter { it !is ParagraphSpan }.isEmpty()) {
+                    output.insert(spanEnd, "\n")
+                }
+            }
+
+            //we don't care about actuall ParagraphSpan in calypso - paragraphs are made from double newline
```
0nko: typo: actuall

hypest:
> The "Calypso mode" code is deeply intertwined with the core editor. If Aztec is to be a general-purpose editor library I think this feature shouldn't be there.

I too agree with @0nko that this mode is deeply intertwined with the rest of the lib and would prefer it factored out.

By the way, I think I notice some code changes not being wrapped in the `isInCalypsoMode` either (example: the `EndOfParagraphMarker` code in `AztecParser.kt`) and should probably be wrapped as well. While at it, I think the marking of `EndOfParagraphMarker`s should be in its own function rather than in `addVisualNewlinesToBlockElements`.

Even though it will be quite an effort to design a "plugin system" to have the Calypso paragraph treatment in one and so we can do it in a separate PR, I think some work should be done here in this PR first. Otherwise I can easily see it becoming even harder to do it in the future. For now, I would highly recommend to at least have all Calypso-mode code be in a single class, it's methods of which will be used throughout the lib, instead of the code snippets be scattered around the lib as it is at the moment. The benefit will be to at least make it easy for our future-selves to identify/pinpoint which code we need to factor out to a plugin.

0nko:
> I think some work should be done here in this PR first. 

Since we have a beta-release today and this PR is functionally OK I don't want this to be left out just for the refactoring. We can start working on it right away but I'd like this feature to ship because the editor isn't really usable in WP without it.

khaykov:
I fixed the issue @hypest mentioned, and also did a quick refactoring - moved most of the pre/post processing code to the ```Format.kt```, and to ```EndOfParagraphMarkerAdded```.
Also fixed visual appearance of paragraph in quote.

Hope I didn't break anything :)

khaykov:
One moment, forgot to add another minor change.

khaykov:
Done!

0nko:
A return inside a quote adds a paragraph break in the visual mode but when I switch to HTML there is only a regular single newline. When I switch back to visual the paragraph break is gone.

aztec/src/main/kotlin/org/wordpress/aztec/spans/EndOfParagraphMarker.kt
```diff
@@ -12,10 +12,18 @@ class EndOfParagraphMarker(var verticalPadding: Int = 0) : LineHeightSpan, Updat
...
+        val actualPadding: Int
+
+        if (spanned.getSpans(spanEnd, spanEnd, AztecQuoteSpan::class.java).any { spanned.getSpanEnd(it) == spanEnd }) {
+            actualPadding = 0
+        } else {
+            actualPadding = verticalPadding * 2
+        }
+
         if (end == spanEnd) {
             //padding is applied only to the bottom of paragraph, so we multiply it by 2
```
0nko: This comment isn't relevant anymore, is it?

0nko:
I also noticed it's now not possible to close the quote. Newlines just expand it.

khaykov:
This fix for closing blockquotes will work in calypso, because we don't really use paragraph spans after certain point. It will still happen in normal mode, so I will make an issue for it.

0nko:
Awesome! :shipit:

PR #310, opened 2017-04-11
Don't disable text-change listener in image text watcher

0nko:
Removes the need to disable text-change listener when normalizing text around full-width images, as it could have undesired side effects.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/watchers/FullWidthImageElementWatcher.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/watchers/FullWidthImageElementWatcher.kt
@@ -1,10 +1,12 @@
 package org.wordpress.aztec.watchers
 
 import android.text.Editable
+import android.text.Spanned
 import android.text.TextWatcher
 import org.wordpress.aztec.AztecText
 import org.wordpress.aztec.Constants
 import org.wordpress.aztec.spans.AztecFullWidthImageSpan
+import org.wordpress.aztec.spans.FullWidthImageProcessingMarker
 
 class FullWidthImageElementWatcher(val aztecText: AztecText) : TextWatcher {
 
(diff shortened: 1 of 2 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/watchers/FullWidthImageElementWatcher.kt
```diff
@@ -30,18 +32,19 @@ class FullWidthImageElementWatcher(val aztecText: AztecText) : TextWatcher {
     }
 
     private fun normalizeEditingAroundImageSpans(count: Int, start: Int) {
-        if (!aztecText.isTextChangedListenerDisabled()) {
+        if (aztecText.text.getSpans(0, 0, FullWidthImageProcessingMarker::class.java).isEmpty()) {
```
0nko: Thanks for the comment. Can you please explain why it's better to also check for `isTextChangedListenerDisabled()`? Thanks!

hypest: > Can you please explain why it's better to also check for isTextChangedListenerDisabled()

Sure! So, the rationale behind the disabling/enabling the "TextChangeListener" is to limit the reactions to text changes. All text listeners should check for the flag and refrain from reacting to the text change, unless it's paramount to perform the reaction.

We trust the caller of the disabling to know exactly what they're doing by wanting to bypass the listeners. An example of bypassing the flag is the `EndOfBufferMarkerAdder` which makes sure that the end-of-buffer-marker presence is always consistent. But even that should be revisited some day.

In our case here, with FullWidthImageElementWatcher, its functionality doesn't seem critical enough to circumvent the flag so, it should respect the flag. Do we have evidence/cases where it should perform its function no matter what? Let's expand if we do have such cases in mind.

That said, the whole notion of the listeners disable flag is "dangerous" and we better be always vigilant when using it.

0nko: Thanks, @hypest!

Issue #314, opened 2017-04-13
Allow element attributes to be modified from the outside


0nko:
We need to allow attributes to be modified from the outside so that the library clients can build custom UI to modify properties of elements.

The motivation behind this is the [image settings issue](https://github.com/wordpress-mobile/WordPress-Android/issues/5296), which requires a way to pass image properties back to the image element attributes.

PR #321, opened 2017-04-19
Issue/314 editable attrs

0nko:
Addresses #314. 

Basically it allows for an arbitrary element to have its attributes set by the library client. This PR also refactors the attribute property of spans for easier manipulation.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -212,7 +212,7 @@ class MainActivity : AppCompatActivity(),
         setContentView(R.layout.activity_main)
 
         // Setup hiding the action bar when the soft keyboard is displayed for narrow viewports
-        if (resources.configuration.orientation === Configuration.ORIENTATION_LANDSCAPE
+        if (resources.configuration.orientation == Configuration.ORIENTATION_LANDSCAPE
                 && !resources.getBoolean(R.bool.is_large_tablet_landscape)) {
             mHideActionBarOnSoftKeyboardUp = true
         }
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -671,7 +671,7 @@ class MainActivity : AppCompatActivity(),
         addVideoMediaDialog!!.show()
     }
 
-    override fun mediaTapped(attrs: Attributes?, naturalWidth: Int, naturalHeight: Int) {
+    override fun mediaTapped(attrs: AztecAttributes, naturalWidth: Int, naturalHeight: Int) {
         ToastUtils.showToast(this, "Media tapped!")
     }
 }
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -173,7 +173,7 @@ class MainActivity : AppCompatActivity(),
         aztec.insertMedia(BitmapDrawable(resources, bitmap), attrs)
 
         val predicate = object : AztecText.AttributePredicate {
-            override fun matches(attrs: Attributes): Boolean {
+            override fun matches(attrs: AztecAttributes): Boolean {
                 return attrs.getValue("id") == id
             }
         }
@@ -568,12 +568,12 @@ class MainActivity : AppCompatActivity(),
 
     override fun onToolbarHtmlModeClicked() {
         val uploadingPredicate = object : AztecText.AttributePredicate {
-            override fun matches(attrs: Attributes): Boolean {
+            override fun matches(attrs: AztecAttributes): Boolean {
                 return attrs.getIndex("uploading") > -1
             }
         }
 
-        val mediaPending = aztec.getAllMediaAttributes(uploadingPredicate).size > 0
+        val mediaPending = aztec.getAllMediaAttributes(uploadingPredicate).isNotEmpty()
 
         if (mediaPending) {
             ToastUtils.showToast(this, R.string.media_upload_dialog_message)
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -17,6 +17,7 @@ import android.os.Environment
 import android.os.Handler
 import android.provider.MediaStore
 import android.support.v4.app.ActivityCompat.OnRequestPermissionsResultCallback
+import android.support.v4.content.ContextCompat
 import android.support.v4.content.FileProvider
 import android.support.v7.app.AlertDialog
 import android.support.v7.app.AppCompatActivity
@@ -181,7 +182,7 @@ class MainActivity : AppCompatActivity(),
         aztec.setOverlay(predicate, 0, ColorDrawable(0x80000000.toInt()), Gravity.FILL)
         aztec.updateElementAttributes(predicate, attrs)
 
-        val progressDrawable = resources.getDrawable(android.R.drawable.progress_horizontal)
+        val progressDrawable = ContextCompat.getDrawable(this, android.R.drawable.progress_horizontal)
         // set the height of the progress bar to 2 (it's in dp since the drawable will be adjusted by the span)
         progressDrawable.setBounds(0, 0, 0, 4)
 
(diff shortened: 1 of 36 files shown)


app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -178,23 +179,29 @@ class MainActivity : AppCompatActivity(),
             }
         }
 
-        aztec.setOverlay(predicate, 0, ColorDrawable(0x80000000.toInt()), Gravity.FILL, attrs)
-        val progressDrawable = resources.getDrawable(android.R.drawable.progress_horizontal)
+        aztec.setOverlay(predicate, 0, ColorDrawable(0x80000000.toInt()), Gravity.FILL)
+        aztec.updateElementAttributes(predicate, attrs)
```
hypest: Any particular reason to do the attribute setting in a separate call? With the separate call, the loop/search for the element is performed again, effectively doubling the CPU effort to find and alter the element. The `updateElementAttributes()` is a good API to have and expose to Aztec's client but, not sure if dropping the attribute updating from `setOverlay()` (and similar methods) is really necessary.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
```diff
@@ -245,12 +246,12 @@ class BlockFormatter(editor: AztecText, val listStyle: ListStyle, val quoteStyle
...
             TextFormat.FORMAT_HEADING_4,
             TextFormat.FORMAT_HEADING_5,
-            TextFormat.FORMAT_HEADING_6 -> return AztecHeadingSpan(nestingLevel, textFormat, "", headerStyle)
+            TextFormat.FORMAT_HEADING_6 -> return AztecHeadingSpan(nestingLevel, textFormat, attrs, headerStyle)
             else -> ParagraphSpan(nestingLevel, attrs)
         }
     }
 
-    fun <T : Class<out AztecBlockSpan>> makeBlockSpan(type: T, nestingLevel: Int, attrs: String = ""): AztecBlockSpan {
+    fun <T : Class<out AztecBlockSpan>> makeBlockSpan(type: T, nestingLevel: Int, attrs: AztecAttributes = AztecAttributes()): AztecBlockSpan {
```
hypest: May I suggest doing the same treatment to `liftBlock()` as well? Currently, that method still has `attrs: String = ""` in its signature. It's a non used parameter as well so, safe to remove it anyway. Thanks!

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecMediaSpan.kt
```diff
@@ -98,26 +98,26 @@ class AztecMediaSpan(context: Context, drawable: Drawable?, var attributes: Attr
...
+        attributes.removeAttribute("aztec_id")
 
+        sb.append(attributes)
         sb.append("/>")
+
         return sb.toString()
     }
 
     fun onClick(view: View) {
+        // add a unique ID that can identify this image span
```
hypest: Interesting. Any background info you can share about the need/usage of this attribute? Thanks!

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
0nko: The method's name is `setOverlay()` yet it performed attribute updates as well, which was a hidden side effect that the caller might not be aware of just by looking at it (I know I wasn't). By the Single Responsibility Principle `setOverlay()` shoud set overlays, IMO. Also, in wpandroid there are cases where you need to apply several overlays and each call sets same attributes over and over.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
0nko: Nice catch!

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecMediaSpan.kt
0nko: Because unless the image is currently being uploaded (it has `data-wpid` attribute) there is no `id` to use for finding elements.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
hypest: Thanks! I think that the performance hit from looping over every element is larger than the redundant setting of the same attributes but, this is not something I would block this PR over. Plus, I like the `Single Responsibility Principle` too so, 👍  😃

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecMediaSpan.kt
hypest: Hmm, that sounds like something the client should do, not Aztec. Besides, why should that attribute be set on `onClick()`? I guess it's an opportunistic setting but still, too "hackish". Also, the attribute name itself seems "magical", in the sense that it's hardcoded and not advertised (i.e. a public constant) anywhere.

Even if we keep the idea of performing this in `onClick`, the client can do that, if it wishes to, in its `onMediaTappedListener?.mediaTapped()` implementation.

Unless there is some reason beyond this one, I'd say let's remove this attribute setting. WDYT?

0nko: > Hmm, that sounds like something the client should do, not Aztec

I'd agree but if you can't identify an element you can't add attributes to it from the client :).

>  I guess it's an opportunistic setting but still, too "hackish". Also, the attribute name itself seems "magical", in the sense that it's hardcoded and not advertised (i.e. a public constant) anywhere.

Perhaps. Maybe it should a more general rule - maybe it should be added to every element, not just images. The "id property" name could then be exposed through some public constant in `AztecText` and the client would know that it's always there if necessary.

> Even if we keep the idea of performing this in onClick, the client can do that, if it wishes to, in its onMediaTappedListener?.mediaTapped() implementation.

As mentioned above, if you don't have an ID already, you won't be able to find the element to add it.

hypest: > maybe it should be added to every element, not just images.

That's basically what I'd expect the client to do: to proactively set IDs to elements, if it needs them.

> > Even if we keep the idea of performing this in onClick, the client can do that, if it wishes to, in its onMediaTappedListener?.mediaTapped() implementation.
> 
> As mentioned above, if you don't have an ID already, you won't be able to find the element to add it.

Let me be more clear: Upon click, client's `mediaTapped` is called. The client can perform the "if attribute missing, add a random value to it" operation in that call since the element's attributes are passed to it via a parameter of `mediaTapped`. There is no need to _search_ for the element inside that call, right?

0nko: > Upon click, client's mediaTapped is called. The client can perform the "if attribute missing, add a random value to it" operation in that call since the element's 

Sorry, I should have checked. You're right, the client can add it there. But that's a specific case when you have access to the `attributes` property of the span. What if you want to update an element without tapping on it? I know this is a hypothetical question but right now you wouldn't be able to do that. This could be solved if the library guaranteed the presence of an "Aztec id" for every element. What do you think>

hypest: > What if you want to update an element without tapping on it?
> ...
> This could be solved if the library guaranteed the presence of an "Aztec id" for every element. What do you think?

OK, let's see, the client will need to know that ID in order to search for it, right? But, how will the client know the ID value since Aztec sets it it to a UUID that cannot be known in advance?

0nko: The client has access to the current attributes, so they'd know the value of that ID from the beginning because the library would add it to every element.

hypest: Let me rephrase: Say the client needs to update a particular element. How would it know its aztecId? Would it have it in a list of its own (outside the Aztec's text html)? That means the client will need to have proactively scanned the document and compile the list of IDs. But that's equivalent to _setting_ the IDs proactively in that same loop. They wouldn't need Aztec to have it done it for them.

0nko: Fair enough. I guess the client doesn't really have a way to really "get" to the element besides tapping on it and if we add a way we can always supply the attributes. I'll move the ID creation to the client 👍.

0nko: Done. Ready for a review ;).

hypest: Thanks for the changes @0nko , good to go now!

PR #334, opened 2017-04-27
Repo cleanup

0nko:
After going through Aztec's dependencies I noticed we had an unnecessary dependency on WP utils library, so I got rid of it. 

Also, [Kotlin 1.1.2 was released](https://blog.jetbrains.com/kotlin/2017/04/kotlin-1-1-2-is-out/) so I thought I might as well update all of the dependencies. It's right after a release so it should be safe and we have lots of time to notice if somethings not right.

I removed a couple of Lint warnings, while at it.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -22,6 +22,10 @@ android {
     sourceSets {
         main.java.srcDirs += 'src/main/kotlin'
     }
+
+    lintOptions {
+        disable 'GradleCompatible'
+    }
 }
 
 dependencies {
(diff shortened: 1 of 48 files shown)


0nko:
There used to be a problem with unit tests when I tried upgrading to Gradle 3.x with an older plugin version. But it seems to be working now. Thanks for the suggestion.

0nko:
Travis started to care about lint errors all of a sudden. After fixing them and disabling the ones that weren't really a problem the build seems to work.

PR #339, opened 2017-05-11
Initial Setup for Espresso UI Tests

rachelmcr:
Adds initial setup for Espresso UI tests in the Aztec demo app. This setup excludes the text & HTML initialized in the demo app, so the UI tests can begin with a blank editor.

h/t @maxme for the setup help!

### Review
@0nko


app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -119,6 +119,8 @@ class MainActivity : AppCompatActivity(),
                 UNKNOWN +
                 EMOJI +
                 LONG_TEXT
+
+        var testMode: Boolean = false
```
0nko: You can detect the the test mode automatically, instead of having to update this variable manually:
```
private val isRunningTest : Boolean by lazy {
    try {
        Class.forName("android.support.test.espresso.Espresso")
        true
    } catch (e: ClassNotFoundException) {
        false
    }
}
```

catehstn:
This is so exciting! 😍

rachelmcr:
@0nko Thanks for the suggestion for detecting the test mode — I like that approach! Ready for another look. :)

0nko:
:shipit:

PR #343, opened 2017-05-13
Issue/format button media

theck13:
### Fix
Remove checked state from media format button to mimic behavior of other buttons which trigger popup menus (e.g. ***Heading***).

#### Before
format_button_media_before

#### After
format_button_media_after

### Test
1. Tap ***Media*** format button.
2. Notice ***Media*** format button is not highlighted.
3. Tap ***Heading*** format button.
4. Notice ***Heading*** format button is not highlighted.


khaykov:
:shipit:

Issue #342, opened 2017-05-12
View Scrolls to Cursor When Uploading Image Below Scroll Position


theck13:
### Expected
View remains at scrolled position when uploading images.

### Observed
View scrolls to the cursor position when uploading images <i>below</i> scroll position, but remains at scrolled position when uploading images <i>above</i>.  See the animation below for example.

![image_uploading_scroll](https://cloud.githubusercontent.com/assets/3827611/26017198/9e051caa-3725-11e7-9b30-433d7da6c54e.gif)

### Reproduced
1. Place cursor near bottom of view.
2. Tap ***Media*** format button.
3. Choose image to insert.
4. Scroll to top of view.
5. Notice view is scrolled to cursor.
6. Place cursor near top of view.
7. Tap ***Media*** format button.
8. Choose image to insert.
9. Scroll to bottom of view.
10. Notice view does not scroll to cursor.

#### Tested
Google Pixel on Android 7.1.2 with AztecDemo 1.0

khaykov:
fixed in #349

PR #340, opened 2017-05-11
Issue/313 image loading scroll

theck13:
### Fix
When first opening content containing images or toggling from HTML to visual view with images, remain at the scrolled position while images are loading as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/313 and https://github.com/wordpress-mobile/WordPress-Android/issues/5184.  The content still moves when images load due to the difference in size between the placeholder and loaded images.  Since images can be loaded from an arbitrary URL, the image size cannot be determined before loading to compensate for the size difference and offset the content shift.

### Test
0. Clear ***AztecDemo*** app data.
1. Launch demonstration app.
2. Scroll down quickly.
3. Notice view doesn't scroll to top when image loads.
4. Tap ***HTML*** format button.
5. Select all text.
6. Paste text below.
```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
<img src="http://goo.gl/i4Z69E">
<img src="http://goo.gl/dWbMtv">
Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
<img src="http://goo.gl/3urcsh">
<img src="http://goo.gl/8C4lZo">
Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
<img src="http://goo.gl/dL6tOu">
<img src="http://goo.gl/FJQ3yw">
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
<img src="http://goo.gl/9fNoiH">
<img src="http://goo.gl/AbW6tw">
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```
7. Tap ***HTML*** format button.
8. Scroll down quickly.
9. Notice view doesn't scroll to top when images load.

#### Note
Using emulators to test may help simulate slow and/or poor network connections to make image loading more pronounced.

Issue #313, opened 2017-04-12
Image-loading scrolls the editor to the cursor


0nko:
When an image is being uploaded and a user scrolls away from the cursor, the editor automatically scrolls back to the cursor on every image-loading progress bar update.

khaykov:
Fixed in #349

PR #349, opened 2017-05-18
Issue/342 image uploading scroll

theck13:
### Fix
View remains at scrolled position when uploading images as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/313 and https://github.com/wordpress-mobile/AztecEditor-Android/issues/342.  The `clearFocus()` change in https://github.com/wordpress-mobile/AztecEditor-Android/pull/340 *does* clear the focus from the `AztecText` view, but the first focusable view (i.e. `AztecText` before this change) regains focus when the system returns from the media chooser.  By allowing the parent view to be focusable, the `AztecText` view is not focused when the activity is resumed and the view is not scrolled when the image uploading overlay is updated.

### Test
1. Place cursor near bottom of view.
2. Tap ***Media*** format button.
3. Choose image to insert.
4. Scroll to top of view.
5. Notice view does not scroll.
6. Place cursor near top of view.
7. Tap ***Media*** format button.
8. Choose image to insert.
9. Scroll to bottom of view.
10. Notice view does not scroll.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -197,7 +197,7 @@ class MainActivity : AppCompatActivity(),
         attrs.setValue("id", id)
         attrs.setValue("uploading", "true")
 
-        aztec.insertMedia(BitmapDrawable(resources, bitmap), attrs)
+        val mediaSpan = aztec.insertMedia(BitmapDrawable(resources, bitmap), attrs)
 
         val predicate = object : AztecText.AttributePredicate {
             override fun matches(attrs: Attributes): Boolean {
@@ -221,7 +221,7 @@ class MainActivity : AppCompatActivity(),
         val runnable: Runnable = Runnable {
             aztec.setOverlayLevel(predicate, 1, progress)
             aztec.updateElementAttributes(predicate, attrs)
-            aztec.refreshText()
+            aztec.updateMediaSpan(mediaSpan)
             progress += 2000
 
             if (progress >= 10000) {
--- a/app/src/main/res/layout-v17/activity_main.xml
+++ b/app/src/main/res/layout-v17/activity_main.xml
@@ -21,9 +21,7 @@
 
         <FrameLayout
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:focusable="true"
-            android:focusableInTouchMode="true" >
+            android:layout_height="wrap_content" >
 
             <org.wordpress.aztec.AztecText
                 android:id="@+id/aztec"
--- a/app/src/main/res/layout-v17/activity_main.xml
+++ b/app/src/main/res/layout-v17/activity_main.xml
@@ -21,7 +21,9 @@
 
         <FrameLayout
             android:layout_width="match_parent"
-            android:layout_height="wrap_content">
+            android:layout_height="wrap_content"
+            android:focusable="true"
+            android:focusableInTouchMode="true" >
 
             <org.wordpress.aztec.AztecText
                 android:id="@+id/aztec"
--- a/app/src/main/res/layout/activity_main.xml
+++ b/app/src/main/res/layout/activity_main.xml
@@ -21,9 +21,7 @@
 
         <FrameLayout
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:focusable="true"
-            android:focusableInTouchMode="true" >
+            android:layout_height="wrap_content" >
 
             <org.wordpress.aztec.AztecText
                 android:id="@+id/aztec"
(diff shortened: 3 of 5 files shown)


khaykov:
Works nicely, but I noticed that the cursor still gets removed on upload progress.
Also, I wonder, would it be a good idea to set focusable state of parent from within AztecText? This way we could make sure that it works with any parent.

theck13:
> I noticed that the cursor still gets removed on upload progress.

The "hack" of resetting the text content in the [`AztecText.refreshText`](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L777-L785) method causes the view to scroll to the cursor if the view is in focus.  As a hack for the hack, the focus was removed from the view, but that means the cursor is removed with the focus.  If there is a way to remove the focus while still keeping the cursor, I'm up for it.

> I wonder, would it be a good idea to set focusable state of parent from within AztecText?

We can get the parent of the `AztecText` view programmatically as with 0f9312c5dc01a5b42588ea67043ed48143390cc2, but the parent is a `ViewParent` which is not always a `View`.  Even though `AztecText` is meant to have at least one parent view, it could be used as the root view in which case the scrolling issue would still occur.

khaykov:
I tried to look into all this focus problem and `AztecText.refreshText` method and looks like we can apply the same fix as in disappearing paragraph's PR. We can replace `refreshText` with a method like this:

```
    fun updateMediaSpanProgress(span: AztecMediaSpan) {
        editableText.setSpan(span, text.getSpanStart(span), text.getSpanEnd(span), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
    }
```
We probably want to pass only a relevant span to it. It's easy to do when we load initial images from Html ([here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L694)) we might want to take this method out of `post {}` and pass span to it.

When adding media directly using [LineBlockFormatter.insertMedia](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt#L121) we might want to return span all the way back to place where we used to call `refreshText` (in case of demo app it will be [here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L224)).
I tested this method a bit, and looks like it works ok, but there might be some that I missed.

I hope it makes sense :)

theck13:
Thanks for the tip, @khaykov!  I tried the changes you suggested.  That method works really well when adding media directly from the format toolbar.  While testing media loaded from HTML, I found that the images overlap other images and text.  Try pasting the example text below into the HTML view then toggling to visual view.  Do you have any issues with your test code?

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
<img src="http://goo.gl/i4Z69E">
<img src="http://goo.gl/dWbMtv">
Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
<img src="http://goo.gl/3urcsh">
<img src="http://goo.gl/8C4lZo">
Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
<img src="http://goo.gl/dL6tOu">
<img src="http://goo.gl/FJQ3yw">
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
<img src="http://goo.gl/9fNoiH">
<img src="http://goo.gl/AbW6tw">
Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```

khaykov:
Yep, it overlaps from time to time. I guess swapping drawable inside span does not work too well. I think we'll have to use refreshText() when downloading images for now. It's a part of initial loading, so it's less annoying.

Btw, so much OOM from those images!

theck13:
The issue with overlapping images loaded from HTML is that we don't know the dimensions beforehand so the span size changes once the image is loaded.  A possible fix would be to first fetch only the dimensions in order to render the placeholder image in the size of the image to be loaded and then do a second network call to load the final image.  In the meantime, I applied the changes you suggested when adding media directly from the format toolbar and kept `refreshText()` when loading media from HTML.

khaykov:
Thanks for the changes! Looks good 👍

Issue #338, opened 2017-05-10
Toggling From HTML View Merges Consecutive Spaces


theck13:
### Expected
All spaces are retained when toggling between visual and HTML views.

### Observed
Consecutive spaces are merged into a single space when toggling between visual and HTML views.

### Reproduced
1. Enter text with multiple spaces.
2. Tap ***HTML*** format button.
3. Notice multiple spaces are retained.
4. Tap ***HTML*** format button.
5. Notice multiple spaces are merged into a single space.

#### Tested
Google Pixel on Android 7.1.2 with AztecDemo 1.0

khaykov:
After trying for some time to implement it in some way and failing, I found out that this behavior is in HTML spec, and for a good reason. To have a formatted HTML, we need to merge all the spaces (unless in pre tags).

[Relevant discussion on SO](https://stackoverflow.com/questions/433493/why-do-multiple-spaces-in-an-html-file-show-up-as-single-spaces-in-the-browser).

maxme:
HTML specs FTW - closing that one as won't fix

PR #355, opened 2017-05-29
Issue/322 keep whitespaces inside pre tags

### LINKED ISSUE #322: Whitespace being stripped from <blockquote>, <pre> or <code>
Newline characters and spaces are being stripped away, which modifies the original text. The only way to have multiple lines is to add the `<br>` tag inside, which is not what's expected.

The editor should not modify the text inside these elements in any way and process it as is.

khaykov:
### Fixes #322 

This PR preserves spaces and newlines in `<pre>` and `<code>` tags.
In addition, instead of adding a paragraph on the press of enter key, it adds a regular newline.

The problem of preserving spaces inside blockquotes is more related to #338, so I will try to address it there.

To test:
Try to add spaces or newlines to content inside `<pre>` or `<code>` tags.
Switch to HTML and back, notice that newlines and spaces are preserved.

### DIFF
--- a/aztec/src/main/java/org/wordpress/aztec/Html.java
+++ b/aztec/src/main/java/org/wordpress/aztec/Html.java
@@ -182,6 +182,7 @@ class HtmlToSpannedConverter implements ContentHandler, LexicalHandler {
 
     public int unknownTagLevel = 0;
     public Unknown unknown;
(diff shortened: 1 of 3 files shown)


0nko:
Nice!! 🎈

PR #356, opened 2017-05-30
Ability to add paragraph to the first line of editor.

### LINKED ISSUE #350: Pressing enter before the first line in editor adds newline instead of paragraph
### Expected
Pressing enter before the first line in editor adds paragraph before this line

### Observed
Pressing enter before the first line in editor adds newline before this line

### Reproduced
1. Put cursor at the beginning of editor and press enter.
2. Notice that instead of paragraph newline was added before the second line

khaykov:
### Fix #350 

Adds ability to add a paragraph to the previous line when pressing enter at 0 index.

### Test
1. Load empty editor and type some text.
2. Place cursor at the beginning of editor and press enter.
3. Notice that the first line in editor is a paragraph (has vertical padding)


0nko:
Weird, I can't reproduce it anymore either.. Good job :).

:shipit:

PR #358, opened 2017-06-02
Video support

### LINKED ISSUE #20: Video support
We should be able to add videos (from the device or hosted online) in the editor.

0nko:
Resolves #20. 

Adds support for the `<video>` tag. Currently only the `src` attribute is supported (multiple formats using the `<source>` sub-elements won't work. 

Also, there is a new video thumbnail loader in the Glide loader project, as an example.

image


0nko:
> I have a couple of thoughts - should the ability to add video be in this PR too? Functionality wise it probably would be same as images.

Sure, I'll add it to the "add media" menu.

> Speaking of which, image span is similar to video span, so I wonder will it make sense to consolidate them into single "MeadiaSpan" that will have type variable (image, video, audio)

Initially that's what I did. But then I realized that videos will always be full-width images (which implement `AztecFullWidthImageSpan`), while image spans can be surrounded by text. Also searching for spans inside spannables by type is easier this way.

> ...and universal MediaTapListener with methods like oImageTapped, onVideoTapped, onAudioTapped, etc.?

I suppose we could have a common interface with all handlers but I think it's unnecessary to force clients to implement all of them if they might not necessarily use all of them. 

What do you think?

khaykov:
Yeah, this makes sense. Agree with both points. The only thing - we should probably simplify addition of click listeners (maybe attach them at `switchToAztecStyle`?) instead of passing them one by one, all the way down to the parser and tag handler.

0nko:
> The only thing - we should probably simplify addition of click listeners (maybe attach them at switchToAztecStyle?) instead of passing them one by one, all the way down to the parser and tag handler.

Great idea!  👍 It's now ready for a review.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -178,27 +187,64 @@ class MainActivity : AppCompatActivity(),
...
+    }
+
+    private fun generateAttributesForMedia(mediaPath: String): Pair<String, AztecAttributes> {
         val id = (Math.random() * Int.MAX_VALUE).toString()
 
         val attrs = AztecAttributes()
         attrs.setValue("src", mediaPath) // Temporary source value.  Replace with URL after uploaded.
         attrs.setValue("id", id)
         attrs.setValue("uploading", "true")
+        attrs.setValue("video", "true")
```
khaykov: This attribute will also be applied to images inserted through `insertImageAndSimulateUpload` and will result in "play" button overlay on top of them

0nko: Good eye! Thanks.

khaykov:
:shipit:

Issue #366, opened 2017-06-08
Format Toolbar Redesign: Remove Features Excluded from First Release


theck13:
Remove features (i.e. ***Preformat*** and ***Page Break***) from the format toolbar which are to be excluded from the first release.  These features will be hidden rather than deleted since they will be included in subsequent releases.  Even though they will be hidden from the format toolbar, the formats will still be able to be added by entering the `<pre>` and `<!--nextpage-->` tags in the ***HTML*** view.

PR #372, opened 2017-06-09
Issue/366 remove excluded features

theck13:
### Fix
Remove ***Preformat*** and ***Page Break*** from heading menu and format toolbar, respectively, as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/366.

### Test
1. Run demonstration app.
2. Tap ***Heading*** format button.
3. Notice ***Preformat*** is not in popup menu.
4. Tap away from popup menu.
5. Scroll format toolbar to the left.
6. Notice ***Page Break*** is not in format toolbar.


0nko:
:shipit:

Issue #367, opened 2017-06-08
Format Toolbar Redesign: Update Layout and Order of Formats


theck13:
Update format toolbar layout and the order of the format buttons as shown below.

![toolbar_expanded](https://user-images.githubusercontent.com/3827611/26945887-97273ffa-4c4a-11e7-96ec-c759751eb7e2.png)

PR #374, opened 2017-06-09
Issue/367 layout and order

theck13:
### Fix
Update format toolbar layout and order of the format buttons as shown below and as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/367.

toolbar_expanded

### Test
1. Run demonstration app.
2. Notice format toolbar appears like screenshot above.


0nko:
:shipit:

Issue #368, opened 2017-06-08
Format Toolbar Redesign: Add List Menu


theck13:
Add a list menu format button which combines the ***Ordered List*** and ***Unordered List*** format buttons into a popup menu.  When the text selection contains one of the list formats or a list is selected from the popup menu, the list menu format button icon is replaced by the associated list icon.

PR #376, opened 2017-06-10
Issue/368 add list menu

theck13:
### Fix
Add a list menu format button which combines the ***Ordered List*** and ***Unordered List*** format buttons into a popup menu as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/368.

A list format can no longer be removed by tapping the list format included in selected text.  To remove a list format, place the cursor between the list icon (i.e. bullet or number) and the first character of a list item and press the backspace key.

### Test
***Apply (Ordered)***
1. Enter "list" text.
2. Tap ***List*** format button.
3. Tap ***Ordered List*** menu item.
4. Notice ordered list is applied.
5. Tap ***HTML*** format button.
6. Notice following HTML is shown.
```
<ol><li>list</li></ol>
```

***Apply (Unordered)***
1. Enter "list" text.
2. Tap ***List*** format button.
3. Tap ***Unordered List*** menu item.
4. Notice unordered list is applied.
5. Tap ***HTML*** format button.
6. Notice following HTML is shown.
```
<ul><li>list</li></ul>
```

***Remove Ordered)***
1. Enter "list" text.
2. Tap ***List*** format button.
3. Tap ***Ordered List*** menu item.
4. Notice ordered list is applied.
5. Place cursor before bullet and "l" character.
6. Tap ***List*** format button.
7. Tap ***Ordered List*** menu item.
8. Notice ordered list is ***not*** removed.
9. Tap ***Backspace*** key.
10. Notice ordered list is removed.

***Remove (Unordered)***
1. Enter "list" text.
2. Tap ***List*** format button.
3. Tap ***Unordered List*** menu item.
4. Notice unordered list is applied.
5. Place cursor before bullet and "l" character.
6. Tap ***List*** format button.
7. Tap ***Unordered List*** menu item.
8. Notice unordered list is ***not*** removed.
9. Tap ***Backspace*** key.
10. Notice unordered list is removed.

***Select***
1. Tap ***HTML*** format button.
2. Enter following HTML.
```
<ol><li>ordered</li></ol>none<ul><li>unordered</li></ul>
```
3. Tap ***HTML*** format button.
4. Place cursor in "ordered" text.
5. Noticed list menu has ordered icon.
6. Tap ***List*** format button.
7. Notice ***Ordered List*** menu item is selected.
8. Place cursor in "unordered" text.
9. Noticed list menu has unordered icon.
10. Tap ***List*** format button.
11. Notice ***Unordered List*** menu item is selected.
12. Place cursor in "none" text.
13. Noticed list menu has ordered icon.
14. Tap ***List*** format button.
15. Notice no menu item is selected.


0nko:
There's a small problem with the new list menu. When I apply the Unordered list to one of the headings, the menu item doesn't get selected. I have to select it again to make it work.

Overall, I'm not convinced the removal of the list style only with a backspace is very intuitive. More so when the backspace not only removes the style but it also deletes the newline (and moves the line up), which is not always what the users want to do. Looking at Google Docs and Calypso, they both support the list removal with a button and I think most people would expect it in Aztec, too (I know I would). But maybe I'm missing some insight about this disparity? cc: @iamthomasbishop

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -371,9 +394,36 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
...
+    fun getHeadingMenu(): PopupMenu? {
+        return headingMenu
+    }
+
+    fun getListMenu(): PopupMenu? {
+        return listMenu
+    }
+
+    fun getSelectedHeadingMenuItem(): TextFormat? {
+        if (headingMenu?.menu?.getItem(0)?.isChecked!!) return TextFormat.FORMAT_PARAGRAPH
```
0nko: I don't think we should be using number literals (the same applies to the list menu). By looking at `headingMenu?.menu?.getItem(0)` there is no way to tell what menu item we're trying to get without a including comment. Also, if the order changed in the resource file, it might break this. With `findItem()` you can specify an id to get a particular item.

theck13: That's a good idea.  I changed from integer literal references with `getItem()` to item identification references with `findItem()` in d5a4838bd92892492fece477cbd9fa5c4ed375f7.

theck13:
> When I apply the Unordered list to one of the headings or the "Bold" (in the demo text), the menu item doesn't get selected.

That has been fixed in 2503a9def924d3bb1ad7004eecdf73163a4f3d50.

> when I apply a list to a line under a heading (again in the demo text), the heading indicator in the toolbar changes to that of the line above

That bug appears to be occurring prior to this pull request.  The [changes to the heading menu](https://github.com/wordpress-mobile/AztecEditor-Android/pull/377) just made it more apparent.  Checkout the last commit before any format toolbar redesign changes, 3603c91b848196bcd0b3b6d9ef95054fc2a95156, then follow these steps.
1. Launch demonstration app.
2. Move cursor anywhere in "Bold" text.
3. Tap ***Ordered List*** or ***Unordered List*** format button.
4. Tap ***Heading*** format button.
5. Notice ***Heading 6*** is selected.

I definitely think we need to fix that, but I suggest we create a new issue and address the bug there since the changes in this pull request are not the cause.

> Looking at Google Docs and Calypso, they both support the list removal with a button and I think most people would expect it in Aztec

The removal of the removal condition was reverted in c2db7b3f277e59b407041a98976800e6be2a6443.  In other words, users can remove a list format by tapping the list menu then tapping the currently active list format.

0nko:
Thanks for the updates 👍. I'll create an issue for the heading bug.

:shipit:

PR #382, opened 2017-06-13
Add UI Tests for Simple Text Formatting

rachelmcr:
This PR adds a suite of UI tests for simple text formatting in the Aztec demo app, using the Android Espresso testing framework.

## Before running the tests

Espresso [advises](https://google.github.io/android-testing-support-library/docs/espresso/setup/#setup-your-test-environment) disabling system animations on devices used for testing:

> On your device, under Settings->Developer options disable the following 3 settings:
> 
> - Window animation scale
> - Transition animation scale
> - Animator duration scale

One additional setup step is also required to handle an Espresso issue with clicks (see the caveats below):

On your device, under Settings -> Accessibility -> Touch & hold delay, set the delay to `Long`.

## Test Details

Each test performs the following steps:

1. Enter text (unformatted).
2. Select a formatting button in the toolbar.
3. Enter more text (with formatting).
4. Verify the correct HTML appears in HTML mode for the unformatted and formatted text.

A couple caveats:

- The Espresso framework sometimes performs a long click instead of a single click (see the [Espresso code](https://android.googlesource.com/platform/frameworks/testing/+/android-support-test/espresso/core/src/main/java/android/support/test/espresso/action/GeneralClickAction.java#75) for an explanation). In the Aztec toolbar, this displays a hint instead of selecting the formatting button and causes the test to fail.
- Espresso also has trouble with the editor's link dialog. Although it can identify the fields and buttons in the dialog (and TalkBack works with the dialog as expected) it clicks in the wrong position. This seems to be a bug with the coordinates select for the click action for a custom `AlertDialog` and so I reversed the link formatting test. (It enters the link HTML and then verifies that the link dialog displays the expected values.)

cc @catehstn @lancewillett @hoverduck

### DIFF
--- a/app/src/androidTest/java/org/wordpress/aztec/demo/BetterClickAction.java
+++ b/app/src/androidTest/java/org/wordpress/aztec/demo/BetterClickAction.java
@@ -1,0 +1,120 @@
+package org.wordpress.aztec.demo;
+
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayingAtLeast;
+import static org.hamcrest.Matchers.allOf;
+
+import android.support.test.espresso.core.deps.guava.base.Optional;
+import android.support.test.espresso.PerformException;
+import android.support.test.espresso.UiController;
+import android.support.test.espresso.ViewAction;
+import android.support.test.espresso.action.CoordinatesProvider;
+import android.support.test.espresso.action.PrecisionDescriber;
+import android.support.test.espresso.action.Tap;
+import android.support.test.espresso.action.Tapper;
+import android.support.test.espresso.util.HumanReadables;
+import android.view.View;
+import android.view.ViewConfiguration;
+import android.webkit.WebView;
+import org.hamcrest.Matcher;
+
+/**
+ * Enables clicking on views with 65% or greater displaying.
+ */
+public final class BetterClickAction implements ViewAction {
+    private final CoordinatesProvider coordinatesProvider;
+    private final Tapper tapper;
+    private final PrecisionDescriber precisionDescriber;
+    private final Optional<ViewAction> rollbackAction;
+    public BetterClickAction(Tapper tapper, CoordinatesProvider coordinatesProvider,
+                              PrecisionDescriber precisionDescriber) {
+        this(tapper, coordinatesProvider, precisionDescriber, null);
+    }
+    public BetterClickAction(Tapper tapper, CoordinatesProvider coordinatesProvider,
+                              PrecisionDescriber precisionDescriber, ViewAction rollbackAction) {
+        this.coordinatesProvider = coordinatesProvider;
+        this.tapper = tapper;
+        this.precisionDescriber = precisionDescriber;
+        this.rollbackAction = Optional.fromNullable(rollbackAction);
+    }
+    @Override
+    @SuppressWarnings("unchecked")
+    public Matcher<View> getConstraints() {
+        Matcher<View> standardConstraint = isDisplayingAtLeast(65);
+        if (rollbackAction.isPresent()) {
+            return allOf(standardConstraint, rollbackAction.get().getConstraints());
+        } else {
+            return standardConstraint;
+        }
+    }
+    @Override
+    public void perform(UiController uiController, View view) {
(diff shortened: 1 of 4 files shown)


rachelmcr:
I just edited the original description to add a setup step for changing the "Touch & hold delay" setting to `Long` on the device used for testing. I'd forgotten this step earlier, but it should resolve most issues with the click turning into a long click in these tests (phones and tablets).

I do want to note that I observed a persistent problem with the click turning into a long click if the formatting button was selected as the first step in the test (before adding any text to the editor). That could indicate a potential performance issue when Aztec is first launched, and I went ahead and opened an issue to track that: https://github.com/wordpress-mobile/AztecEditor-Android/issues/383

app/src/androidTest/java/org/wordpress/aztec/demo/TestUtils.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.hasToString;
+
+/**
+ * Test utilities to be used with instrumentation tests.
+ */
+public class TestUtils {
+
+    // Strings
+    public static String unformattedText = "hello";
```
catehstn: Nice that we can have proper strings in Android - unlike iOS! :)

catehstn:
@0nko ptal - if you're happy let's merge it.

app/src/androidTest/java/org/wordpress/aztec/demo/BetterClickAction.java
```diff
@@ -0,0 +1,120 @@
...
+                              PrecisionDescriber precisionDescriber, ViewAction rollbackAction) {
+        this.coordinatesProvider = coordinatesProvider;
+        this.tapper = tapper;
+        this.precisionDescriber = precisionDescriber;
+        this.rollbackAction = Optional.fromNullable(rollbackAction);
+    }
+    @Override
+    @SuppressWarnings("unchecked")
+    public Matcher<View> getConstraints() {
+        Matcher<View> standardConstraint = isDisplayingAtLeast(65);
```
rachelmcr: Yes, the 65 here is a percentage: The default `ClickAction` in Espresso checks to make sure at least 90% of the element is displayed on-screen before clicking on it. The 65 here checks to make sure at least 65% of the element is displayed on-screen instead.

app/src/androidTest/java/org/wordpress/aztec/demo/SimpleTextFormattingTests.java
```diff
@@ -0,0 +1,286 @@
...
+    @Test
+    public void testSimpleOrderedListFormatting() {
+        // Enter text in visual editor with formatting
+        aztecText.perform(typeText(unformattedText + "\n"));
+        orderedListButton.perform(scrollTo(), click());
+        aztecText.perform(typeText(formattedText));
+
+        // Check that HTML formatting tags were correctly added
+        toggleHTMLView();
+        sourceText.check(matches(withText(unformattedText + "\n<ol>\n\t<li>" + formattedText + "</li>\n</ol>")));
```
rachelmcr: The `\t` indents the list item on screen.

hoverduck:
@rachelmcr I finally got a chance to get my Android dev environment set back up and take a look.  Everything checks out 👍 !

My only comment is that you should probably include an update to the README to provide the "Before running the tests" setup steps from the PR details.  Is there maybe a way to automate those steps via a set of ADB commands prior to starting the tests?

0nko:
After following the device configuration steps the tests pass. I'll add these to the README, like @hoverduck suggested. 

Quickly searching for the `adb` commands I did find something for disabling the animations:

```
adb shell content update --uri content://settings/system --bind value:s:0.0 --where 'name="window_animation_scale"'
adb shell content update --uri content://settings/system --bind value:s:0.0 --where 'name="transition_animation_scale"'
adb shell content update --uri content://settings/system --bind value:s:0.0 --where 'name="animator_duration_scale"'
```

And for setting a long tap delay:

```
adb shell settings put secure long_press_timeout <time in ms>
```

I haven't tested them, though. @rachelmcr, thank you so much for setting up the test environment and writing these!

0nko:
:shipit:

Issue #373, opened 2017-06-09
Format Toolbar Redesign: Add Horizontal Rule Format Button


theck13:
Add horizontal rule format button to the format toolbar using [`gridicons-minus`](http://automattic.github.io/gridicons/) icon.

PR #384, opened 2017-06-14
Issue/373 horizontal rule button

theck13:
### Fix
Add horizontal rule format button to the format toolbar as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/373.

### Test
1. Notice ***Horizontal Rule*** format button.
2. Tap ***Horizontal Rule*** format button.
3. Notice horizontal rule is added at cursor location.

#### Note
The horizontal rule format button icon will appear slightly larger than the other icons due to 71c83fd93e2b0a42b59a0c6a0e1ffcc9a41b6e69.  The other icons are updated in https://github.com/wordpress-mobile/AztecEditor-Android/pull/387.


0nko:
When you add a HR, put a cursor at the end and type a letter it seems to add an extra space below it (unlike the MORE tag). Both spans inherit from the same interface so I'm not sure why...

theck13:
> When you add a HR, put a cursor at the end and type a letter it seems to add an extra space below it (unlike the MORE tag). Both spans inherit from the same interface so I'm not sure why...

That should be fixed with aea1b7d9bc497a50e0193d496ad1f1660468ac42.

0nko:
:shipit:

PR #387, opened 2017-06-16
Issue/386 update format buttons

theck13:
### Fix
Update format toolbar, buttons, and icons size as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/386.

### Test
1. Open demonstration app.
2. Notice updated format toolbar, buttons, and icons.


0nko:
Looking nice! :shipit:

Issue #370, opened 2017-06-08
Format Toolbar Redesign: Add Advanced Toolbar Toggle


theck13:
Add an ellipsis button with the [ic_more_horiz](https://material.io/icons/#ic_more_horiz) icon to the far right of the format toolbar.  In this collapsed mode, only the following format buttons are shown.
- Media
- Headings
- Lists
- Quote
- Bold
- Italic

When tapped, the [ic_more_horiz](https://material.io/icons/#ic_more_horiz) icon becomes [ic_more_vert](https://material.io/icons/#ic_more_vert) and the following format buttons are shown.
- Media
- Headings
- Lists
- Quote
- Bold
- Italic
- Link
- Underline
- Strikethrough
- Horizontal Rule
- More
- HTML

Tapping the [ic_more_vert](https://material.io/icons/#ic_more_vert) in expanded mode returns to the collapsed mode with the [ic_more_horiz](https://material.io/icons/#ic_more_horiz) icon.  See the screenshots below for collapsed (i.e. default) and expanded views.  Note that the collapsed screenshot has the Link button shown even though it should not be included in the collapsed state.

![toolbar_collapsed](https://user-images.githubusercontent.com/3827611/26946723-81b10e0a-4c4d-11e7-9209-7bf12a8c4ce8.png)

![toolbar_expanded](https://user-images.githubusercontent.com/3827611/26946727-84c08c7e-4c4d-11e7-906a-64f040572c37.png)

PR #388, opened 2017-06-16
Issue/370 add toolbar toggle

theck13:
### Fix
Add advanced toolbar mode as described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/370.

### Test
0. Add `aztec:advanced="true"` to [AztecToolbar](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/app/src/main/res/layout-v17/activity_main.xml#L8-L13) view.
1. Launch demonstration app.
2. Notice toolbar contains the following format buttons:
- Media
- Headings
- Lists
- Quote
- Bold
- Italic
3. Tap horizontal ellipsis button.
4. Notice toolbar expands to contain the following format buttons:
- Media
- Headings
- Lists
- Quote
- Bold
- Italic
- Link
- Underline
- Strikethrough
- Horizontal Rule
- More
- HTML
5. Tap vertical ellipsis button.
6. Notice toolbar collapses to contain the original format buttons.


0nko:
The animations look cool! But when I expand the advanced toolbar, it looks like the ellipsis button is just replaced by the strike-through button. I think this may confuse some users who don't know they can scroll the toolbar. If the it smooth-scrolled to the right after the advanced buttons pop up it would show that the toolbar's scrollable and there are more buttons on the right (including the ellipsis). What do you think?

theck13:
After discussing the format button animations with @iamthomasbishop, it was decided that sliding the expanded toolbar format buttons from behind the collapsed toolbar format buttons is a better solution.  It follows [Material guidelines for choreography](https://material.io/guidelines/motion/choreography.html) more closely and hints the view can be scrolled by pushing the expanded toolbar format buttons off the screen.

0nko:
Nice job! :shipit:

PR #393, opened 2017-06-22
Issue/update advanced layout

theck13:
### Fix
Update advanced toolbar layout to stick ellipsis buttons to the right side of the layout when toggling collapsed and expanded views.

***Before***
before

***After***
after

### Test
0. Add `aztec:advanced="true"` to [AztecToolbar](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/app/src/main/res/layout-v17/activity_main.xml#L8-L13) view.
1. Launch demonstration app.
2. Tap ***Expand Toolbar*** button.
3. Notice ***Collapse Toolbar*** button remains on right.

#### Note
This is best tested using a device that is wider than 682dp.

### DIFF
--- a/aztec/src/main/res/layout/aztec_format_bar_advanced.xml
+++ b/aztec/src/main/res/layout/aztec_format_bar_advanced.xml
@@ -27,162 +27,175 @@
         android:paddingRight="@dimen/format_bar_right_margin"
         tools:ignore="RtlHardcoded,RtlSymmetry" >
 
-        <RelativeLayout
+        <LinearLayout
             android:layout_height="match_parent"
-            android:layout_width="wrap_content" >
+            android:layout_width="wrap_content"
+            android:orientation="horizontal" >
 
-            <LinearLayout
-                android:id="@+id/format_bar_button_layout_expanded"
+            <RelativeLayout
                 android:layout_height="match_parent"
-                android:layout_toRightOf="@+id/format_bar_button_layout_collapsed"
-                android:layout_width="wrap_content"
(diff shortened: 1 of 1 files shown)


khaykov:
There is a weird animation effect with expand button - when you collapse toolbar, it goes to the right, away from the screen and then bounces back. On Nexus 7 (and maybe other tablets) it does not disappear completely before bouncing back, like this:
3fb4ed8aa551981552aad389525a6267

theck13:
> when you collapse toolbar, it goes to the right, away from the screen and then bounces back

That is expected behavior.  The toolbar [scrolls to the leftmost position before collapsing](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt#L613-L614).  Due to the screen width of the Nexus 7, scrolling to the leftmost position is the width of one button which pushes the ellipsis button off the screen momentarily then brings it back when the toolbar collapses.  You can see in your animation that the media button is off the left of the screen before tapping the collapse ellipsis button.  When the collapse button is tapped, the media button scrolls into view and the toolbar collapses bringing the expand ellipsis button into view.

khaykov:
Oo. I got how it works now! Makes sense. Thanks for explaining!
Anyway, this was my only comment, PR looks good! 👍

0nko:
Is this intended to ship with the open beta release next week? If that's the case, it should have been merged in the `release/open-beta` first, since we're currently in the code-freeze stage.

PR #395, opened 2017-06-27
Fixing tests

khaykov:
This branch is called "feature/run_ui_tests_with_travis", but after playing for a couple of days with UI tests on remote servers I wasn't able to find a way to run them consistently. 

Emulators on Travis are just too slow, so it's almost impossible to perform any UI actions and confirm their result on time.
Firebase Cloud Testing is better, but because we can't control emulators and tweak long click duration with animation, tests just fail randomly.

For now, I'm just going to push fixed test (they got broken after a couple of recent changes), and encourage everyone to run them before submitting a PR. Espresso is still under development, so there is a good chance that this will change for better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ android:
     - tools
     - build-tools-25.0.2
     - android-25
+    - sys-img-armeabi-v7a-android-25
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


0nko:
:shipit:

PR #400, opened 2017-06-30
Issue/toolbar state triggers

theck13:
### Fix
Switch collapse and expand toolbar click listener triggers to the correct toolbar action buttons.


0nko:
Is this PR intended to ship on Monday with the open beta?

0nko:
:shipit:

Issue #394, opened 2017-06-26
Uploading an image makes the editor too jumpy - unusable


mzorz:
Possibly related: #313 
First observed by @catehstn 

### Expected
The user should be able to keep writing / editing while an image is being uploaded

### Observed
Instead, it's too jumpy and it's impossible to keep the cursor where you want to write - even tapping on the Title field seems to be losing focus immediately.

Video here:
https://cloudup.com/cJmqh2lK6UK

### Reproduced
1. Start a new draft
2. Pick an image to upload
3. try to write something either in the title or before/after the image

#### Tested
[Google Pixel] on [7.1.2], observed this happening on latest `alpha-61` and `develop` in WPAndroid, and is consistently reproduceable

khaykov:
I have a feeling we fixed it in the demo app (#349) but forgot to do it in wpandroid. 🤦‍♂️

khaykov:
Fixed in #6208

PR #401, opened 2017-07-03
Issue/394 reset media span by attributes

khaykov:
### Aztec side fix for #394

When updating the progress of uploading media spans, instead of exposing inserted spans we are going to use `AttributePredicate`, that is available at client side to locate and reset them.

### Test
1. Open demo app.
2. Add image to the editor.
3. While the image is being uploaded make sure that cursor is not moving.

On client side instead of calling `refreshText()` during media upload events, we will call `resetAttributeMediaSpan()`.


khaykov:
Corresponding PR on client side:
https://github.com/wordpress-mobile/WordPress-Android/pull/6208

0nko:
:shipit:

PR #402, opened 2017-07-06
Issue/396 fixing image click detection

### LINKED ISSUE #396: Having multiple images doesn't trigger the tap listener for the first image
### Expected
When the user inserts multiple images on a given Post, tapping on any of the images should be handled correctly (i.e. should show the  "stop uploading?"  popup if it's uploading or enqueued, or show the image settings modal screen if it's already uploaded).

### Observed
When inserting 2 images, tapping on the first image doesn't do anything, and debugging shows `onImageTapped()` is not getting called.

### Reproduced
1. Start a new draft
2. Start the media picker to include an image, by tapping on the "+" icon in the toolbar
3. insert 2 images, either by choosing one each time, or by selecting 2 images in the same image picker action 
4. tap on the first image
5. observe that nothing happens

#### Tested
[Google Pixel] on [7.1.2] with [WPAndroid `alpha-61`, `release/7.7` and `develop`]

khaykov:
### Fixes #396 

Improves hit detection for images without spacing between them.

To test:

1. Load empty editor.
2. Add image.
3. Add another image, that is big enough to render below the first image, like this:
tworobots
4. Make sure both images are clickable.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/EnhancedMovementMethod.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/EnhancedMovementMethod.kt
@@ -1,7 +1,6 @@
 package org.wordpress.aztec
 
 import android.graphics.Rect
-import android.text.Selection
 import android.text.Spannable
 import android.text.method.ArrowKeyMovementMethod
 import android.text.style.ClickableSpan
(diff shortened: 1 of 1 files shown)


0nko:
:shipit:

Issue #331, opened 2017-04-26
Refactor the architecture to support pre/post-processing plugins


0nko:
The core Aztec code currently contains features which are WordPress-specific:

- special comments
- Calypso paragraphs

[Shortcodes](https://github.com/wordpress-mobile/AztecEditor-Android/issues/17), which are also WP-specific are of a high-priority because other features are blocked by its support in Aztec.

PR #409, opened 2017-07-11
WordPress Comments Plugin

0nko:
This is the first major block of the Aztec architecture redesign, which will address #331. There are 4 main changes:

1. Implementation of a "plugin architecture", which will allow to add new functionality to Aztec without changing the core library. This PR introduces these types of plugins:

- [Comment Handler](https://github.com/wordpress-mobile/AztecEditor-Android/blob/feature/wp-comments-plugin/aztec/src/main/kotlin/org/wordpress/aztec/plugins/IAztecCommentHandler.kt)
- [Toolbar Button](https://github.com/wordpress-mobile/AztecEditor-Android/blob/feature/wp-comments-plugin/aztec/src/main/kotlin/org/wordpress/aztec/plugins/IAztecToolbarButton.kt)

2. The `Page` and `More` special comments have been extracted and moved to a [separate plugin](https://github.com/wordpress-mobile/AztecEditor-Android/blob/feature/wp-comments-plugin/wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/WordPressCommentsPlugin.kt)
3. A unified [Aztec configuration class](https://github.com/wordpress-mobile/AztecEditor-Android/blob/feature/wp-comments-plugin/aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt), that allows to set listeners and add plugins during the initialization, using the fluent API style
4. General refactoring

The subsequent plugin types (tag handler, text watcher, text preprocessor/postprocessor) will build upon this groundwork.


khaykov:
@0nko It's beautiful! ✨  Simple and easy. 👍 I haven't found any obvious issues so far.
One thing I was thinking about - we probably won't be working on Aztec forever, and for the sake of future contributors, we should add a bit more comments/documentation, especially to the abstract elements. Some methods like `canHandle` are quite ambiguous if you don't know where are they called from.

Another thought - for future plugins we will probably need a way to add style specific watchers and make sure that they are firing in the correct order.

Again, great work! Aztec start's to look like proper library :)

khaykov:
Forgot one thing - with `minSdkVersion` set to 16, we will need to update the "Supports Android 4.0+" line in readme :)

0nko:
Thanks, @khaykov! 😄 

I added documentation to plugin interfaces. Good idea 👍. 

> Another thought - for future plugins we will probably need a way to add style specific watchers and make sure that they are firing in the correct order.

What do you mean by style specific watchers?

khaykov:
> What do you mean by style specific watchers?

For example, List plugin will need a way to install `ListItemHandler` and `ListHandler`.

0nko:
Oh, yeah. I've removed other types of plugins that were not relevant to this PR. We'll add more as we need them or when we think of a use case.

khaykov:
:shipit: 🎉 🕺

PR #421, opened 2017-07-19
Try to reduce memory usage with images

### LINKED ISSUE #420: Aztec consumes too much memory with a lot of images in it
It seems there is a very high memory usage when a lot of pictures are added to the editor. 
From tests I run comparing the old visual editor with Aztec, it seems that the "old" works better in terms of memory consumption, and doesn't crash the app, instead Aztec seems to be more prone to OOM errors.

I investigated it a bit, and found out that `AztecDynamicImageSpan` do not use Weak references to the original drawable. Instead, its parent class `DynamicDrawableSpan` does use weak reference.
This could probably lead to some memory leak (Take it with a grain of salt) and resulting in high resource consumption. 
Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/6283#issuecomment-315759297

daniloercoli:
### Fix #420 
by using weak references to images added to the editor.  The ticket has better description about the issue, and why we're trying this way of retaining drawables.

### Test
1. Add a lot of pictures to the editor.
2. Scroll up and down, add other pictures.
3. Do not close the app, but exit the editor. 
4. Repeat steps above and check that the app doesn't crash.

There is a wp-android branch that use this - https://github.com/wordpress-mobile/WordPress-Android/tree/danilo/aztec-media-testing

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecDynamicImageSpan.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecDynamicImageSpan.kt
@@ -10,6 +10,7 @@ import android.text.Layout
 import android.text.style.DynamicDrawableSpan
 import android.view.View
 import org.wordpress.aztec.AztecText
+import java.lang.ref.WeakReference
 
 abstract class AztecDynamicImageSpan(val context: Context, protected var imageDrawable: Drawable?) : DynamicDrawableSpan() {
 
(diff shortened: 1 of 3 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecDynamicImageSpan.kt
```diff
@@ -10,15 +10,18 @@ import android.text.Layout
...
-abstract class AztecDynamicImageSpan(val context: Context, protected var imageDrawable: Drawable?) : DynamicDrawableSpan() {
+abstract class AztecDynamicImageSpan(val context: Context, imageDrawable: Drawable?) : DynamicDrawableSpan() {
 
     var textView: AztecText? = null
     var originalBounds = Rect(imageDrawable?.bounds ?: Rect(0, 0, 0, 0))
     var aspectRatio: Double = 1.0
 
     private var measuring = false
 
+    protected var mDrawableRef: WeakReference<Drawable>? = WeakReference<Drawable>(imageDrawable)
```
0nko: We tried to avoid using the Hungarian notation in Aztec.

0nko:
Nice job! :shipit:

PR #423, opened 2017-07-19
Fix OOB in AztecMediaSpan

daniloercoli:
### Fix422
by checking the requested overlay ID before setting the new level to it.


0nko:
:shipit:

Issue #418, opened 2017-07-18
Crash report 7.8: RuntimeException in BlockFormatter.toggleQuote


maxme:
```
Fatal Exception: java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:638)
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:626)
       at org.wordpress.aztec.handlers.BlockHandler$Companion.set(BlockHandler.kt:112)
       at org.wordpress.aztec.formatting.BlockFormatter.applyBlock(BlockFormatter.kt:423)
       at org.wordpress.aztec.formatting.BlockFormatter.applyBlockStyle(BlockFormatter.kt:362)
       at org.wordpress.aztec.formatting.BlockFormatter.applyBlockStyle$default(BlockFormatter.kt:329)
       at org.wordpress.aztec.formatting.BlockFormatter.toggleQuote(BlockFormatter.kt:46)
       at org.wordpress.aztec.AztecText.toggleFormatting(AztecText.kt:613)
       at org.wordpress.aztec.toolbar.AztecToolbar.onToolbarAction(AztecToolbar.kt:430)
       at org.wordpress.aztec.toolbar.AztecToolbar.access$onToolbarAction(AztecToolbar.kt:24)
       at org.wordpress.aztec.toolbar.AztecToolbar$initView$1.onClick(AztecToolbar.kt:338)
       at android.view.View.performClick(View.java:4858)
       at android.widget.CompoundButton.performClick(CompoundButton.java:120)
       at android.view.View$PerformClick.run(View.java:20329)
       at android.os.Handler.handleCallback(Handler.java:815)
       at android.os.Handler.dispatchMessage(Handler.java:104)
       at android.os.Looper.loop(Looper.java:194)
       at android.app.ActivityThread.main(ActivityThread.java:5950)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:987)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:782)
```

PR #425, opened 2017-07-20
Disable styling of multiline text with lingering/trailing newlines.

khaykov:
This is a temporary fix for #418 to prevent future crashes until the issue is properly resolved.

Allowing styling of empty lines resulted in series of crashes in scenarios where styling didn't work before.
This PR adds a temporary try/catch block so the interactions that didn't work before will continue not to work instead of crashing.

😭


0nko:
We'll kill that 🐛 soon!

:shipit:

PR #427, opened 2017-07-21
Use default WordPress video shortcode for quick testing

daniloercoli:
This PR basically changes the HTML produced by `AztecVideoSpan`, and prints out the `[video]` default WP shortcode Ref: https://codex.wordpress.org/Video_Shortcode instead of the HTML `video` tag.

This is a quick fix to make video upload working in wp-android, to be used until a full shortcode support is completed in Aztec.

### DIFF
--- a/aztec/src/main/java/org/wordpress/aztec/Html.java
+++ b/aztec/src/main/java/org/wordpress/aztec/Html.java
@@ -225,6 +225,7 @@ class HtmlToSpannedConverter implements ContentHandler, LexicalHandler {
         reader.setContentHandler(this);
         try {
             reader.setProperty(Parser.lexicalHandlerProperty, this);
+            source = source.replaceAll("\\[video([^\\]]*)\\]", "<video$1/>");
             reader.parse(new InputSource(new StringReader(source)));
         } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


0nko:
This works fine when you first add a video but it only shows `[video...]` afterwards.

If you add `source = source.replaceAll("\\[(video.*)\\]", "<$1>");` before the [parse()](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/java/org/wordpress/aztec/Html.java#L228) call, the video span will render properly in visual mode for the existing videos, too.

daniloercoli:
Cool thanks, will add it shortly!

daniloercoli:
It seems that the correct regexp is the following: `source = source.replaceAll("\\[video([^\\]]*)\\]", "<video$1/>");`

The one proposed above doesn't work if you have more than one video, and doesn't close the video tag, duplicating all the videos if you switch between visual mode and text mode multiple times.

daniloercoli:
This should be ready for another round of testing.

0nko:
Good catch, I just quickly tried that regex :).

:shipit:

Issue #424, opened 2017-07-20
Crash report 7.8: RuntimeException in AztecText.paste


maxme:
```
Fatal Exception: java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:688)
       at android.text.SpannableStringBuilder.change(SpannableStringBuilder.java:434)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:547)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:503)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:502)
       at org.wordpress.aztec.AztecText.paste(AztecText.kt:974)
       at org.wordpress.aztec.AztecText.onTextContextMenuItem(AztecText.kt:920)
       at android.widget.Editor$TextActionModeCallback.onActionItemClicked(Editor.java:3315)
       at com.android.internal.policy.PhoneWindow$DecorView$ActionModeCallback2Wrapper.onActionItemClicked(PhoneWindow.java:3617)
       at com.android.internal.view.FloatingActionMode$3.onMenuItemSelected(FloatingActionMode.java:85)
       at com.android.internal.view.menu.MenuBuilder.dispatchMenuItemSelected(MenuBuilder.java:766)
       at com.android.internal.view.menu.MenuItemImpl.invoke(MenuItemImpl.java:152)
       at com.android.internal.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:909)
       at com.android.internal.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:899)
       at com.android.internal.view.FloatingActionMode$4.onMenuItemClick(FloatingActionMode.java:111)
       at com.android.internal.widget.FloatingToolbar$FloatingToolbarMainPanel$1.onClick(FloatingToolbar.java:1018)
       at android.view.View.performClick(View.java:5265)
       at android.view.View$PerformClick.run(View.java:21534)
       at android.os.Handler.handleCallback(Handler.java:815)
       at android.os.Handler.dispatchMessage(Handler.java:104)
       at android.os.Looper.loop(Looper.java:207)
       at android.app.ActivityThread.main(ActivityThread.java:5728)
       at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #428, opened 2017-07-21
Log more context for the past crash

0nko:
Does not fix the #424 issue, but adds more context to find the culprit.


maxme:
:shipit:

PR #432, opened 2017-07-24
Fix drawables keep disappearing

### LINKED ISSUE #431: Drawables keep disappearing
The images (drawables) behave inconsistently and sometimes they disappear. 

To reproduce, run the demo app and notice the image at the top is not visible (kill the app first if you can't see it). Switch to HTML and back and images reappear. It also affects video span, HR and special comments.

It is caused by #421 (the weak link memory optimization). @daniloercoli, could you take a look at this please?

daniloercoli:
Fix #431 by reverting wordpress-mobile/AztecEditor-Android#421

Didn't find an easy way to fix this, except on reverting #431.
We should probably not pass a drawable in the constructor of `AztecDynamicImageSpan`, but a reference instead, and create the drawable when it's required (image is on the screen). Dispose the drawable when it's in not on the screen. Re-create the drawable when needed.

See `DynamicDrawableSpan`.


kwonye:
@0nko can we get a milestone on this please? Tyty

0nko:
:shipit:

PR #433, opened 2017-07-25
[Not ready to be merged] Possible fix for OOM issues with pictures

daniloercoli:
This is just a preliminary test on removing any hard reference to the Bitmaps drawn on the screen, and use path instead.  Ref: #420 

Basically <del>`wp-android`</del> the main app will not add Bitmap/BitmapDrawable to the editor, but will just pass a `resID` or a `String` with the path to the image on disk or on the Internet. The editor should be ready to load the correct picture by itself, and dispose the object if thery're not on the screen anymore. Since we have the Path of each pictures we can re-create them when needed.

Unfortunately this PR changes a lot of files though, but the main changes are here https://github.com/wordpress-mobile/AztecEditor-Android/compare/issue/420-danilo-testing--fix-oom-with-media?expand=1#diff-1254219b4f1ac2f85c98f60b1440f70bR194
and here: https://github.com/wordpress-mobile/AztecEditor-Android/compare/issue/420-danilo-testing--fix-oom-with-media?expand=1#diff-7ca71cd1f009ab8438e0901205001b14R722

Still need to add the code for videos.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -172,7 +172,7 @@ class MainActivity : AppCompatActivity(),
                     options.inDensity = DisplayMetrics.DENSITY_DEFAULT
                     bitmap = BitmapFactory.decodeFile(mediaPath, options)
 
-                    insertImageAndSimulateUpload(bitmap, mediaPath)
+                    insertImageAndSimulateUpload(mediaPath)
                 }
                 REQUEST_MEDIA_PHOTO -> {
                     mediaPath = data?.data.toString()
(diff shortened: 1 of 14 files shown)


0nko:
I should have mentioned this earlier - Aztec is meant to be a general-purpose library that is not designed to be only used with WordPress. That's why I've been working on the plugin architecture and I extracted the special comments from it and put them in a separate module. 

That's why we don't want to introduce WP dependencies, such as (`compile 'org.wordpress:utils:1.16.0'`). Also, the changes that this PR introduces should make sense even if used outside of the WordPress app. Shortcodes, for example, are also WP-specific and will be in a separate plugin, as well. The video shortcode changes that were merged recently are only meant to be temporary and will be removed once the plugin is finished.

daniloercoli:
Thanks for you comment @0nko, although `'org.wordpress:utils:1.16.0'` is just a utility library we can use everywhere. It's not tailored around WP. It *may* (and probably) contains methods that only have meaning for WP development, but we can re-use it in Aztec IMO. Maybe @maxme has better overview than me.

daniloercoli:
As per chat convo we're closing this PR, and try another approach.

PR #436, opened 2017-07-27
Shortcode plugins

0nko:
This PR adds new types of plugins and implements the `[caption]`, `[video]` and `[audio]` shortcode support. Other shortcodes do not require parsing or spans and are displayed as simple plain text.

The new plugins are:
- Regex plain text handler ([`IHtmlTextHandler`](https://github.com/wordpress-mobile/AztecEditor-Android/blob/b533484c8612c931b6f21062b53e1bbb0943fa01/aztec/src/main/kotlin/org/wordpress/aztec/plugins/html2visual/IHtmlTextHandler.kt) interface)
- HTML preprocessor, which is called before the code is parsed to spans ([`IHtmlPreprocessor`](https://github.com/wordpress-mobile/AztecEditor-Android/blob/b533484c8612c931b6f21062b53e1bbb0943fa01/aztec/src/main/kotlin/org/wordpress/aztec/plugins/html2visual/IHtmlPreprocessor.kt) interface)
- HTML postprocessor, which is called after the spans are serialized to HTML ([`IHtmlPostprocessor`](https://github.com/wordpress-mobile/AztecEditor-Android/blob/b533484c8612c931b6f21062b53e1bbb0943fa01/aztec/src/main/kotlin/org/wordpress/aztec/plugins/visual2html/IHtmlPostprocessor.kt) interface)


aztec/src/main/java/org/wordpress/aztec/Html.java
```diff
@@ -651,6 +655,24 @@ public void characters(char ch[], int start, int length) throws SAXException {
             }
         }
 
+        if (plugins != null) {
```
khaykov: What are use cases for this block (and `IHtmlTextHandler`) will be?

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -98,7 +104,7 @@ class MainActivity : AppCompatActivity(),
                 "}" +
                 "</pre>"
         private val CODE = "<code>if (value == 5) printf(value)</code><br>"
-        private val IMG = "<img src=\"https://examplebloge.files.wordpress.com/2017/02/3def4804-d9b5-11e6-88e6-d7d8864392e0.png\" />"
+        private val IMG = "[caption align=\"alignright\"]<img src=\"https://examplebloge.files.wordpress.com/2017/02/3def4804-d9b5-11e6-88e6-d7d8864392e0.png\" />Caption[/caption]"
         private val EMOJI = "&#x1F44D;"
         private val LONG_TEXT = "<br><br>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. "
         private val VIDEO = "<video src=\"https://examplebloge.files.wordpress.com/2017/06/d7d88643-88e6-d9b5-11e6-92e03def4804.mp4\" />"
```
khaykov: Do you think we should include `[video]` and `[audio]` shortcodes into the demo app?

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -22,21 +22,30 @@
...
-class AztecTagHandler : Html.TagHandler {
+class AztecTagHandler(val plugins: List<IAztecPlugin> = ArrayList()) : Html.TagHandler {
 
     private var order = 0
 
     override fun handleTag(opening: Boolean, tag: String, output: Editable,
                            context: Context, attributes: Attributes,
                            nestingLevel: Int): Boolean {
 
+        if (processTagHandlerPlugins(tag, opening, output, attributes, nestingLevel)) {
```
0nko: That's true.

aztec/src/main/java/org/wordpress/aztec/Html.java
0nko: Originally I thought I'd use it for the shortcodes (since they're essentially plaintext). However, with this approach I needed to handle the start & end tag and parse the attributes manually, which was a pain. But then I realized if I convert shortcodes to HTML elements the parser does all of this for free so I used the pre/post-processing plugins instead. I already had the text handler, though, so I thought I'd leave it here in case someone wants to create a plugin for plaintext processing, since it's already working.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
0nko: You're right. The `[video]` shortcode is sort of already there, since the `<video>` tag gets converted to it. But I'll change it to be consistent with the caption and I'll add an audio example, as well.

0nko:
Yeah, the newlines around the caption is expected, although it's a side effect of being considered a block element. I made it a block span because the nesting, appending text & splitting worked out of the box. Plus, we want the caption to be on a separate line anyway.

khaykov:
Looks good now! 👍 

Regarding captions, I don't think it will be a huge problem. But in the future, we might want to tweak captions logic (strip captions that do not have a plain text inside? Calypso does some things like this.)

PR #439, opened 2017-07-28
Make sure to use the correct loading drawable for Images and Video

### LINKED ISSUE #438: Custom image and video loading icons are not used in the editor
Tried to setup up custom loading indicators (drawables), for videos and images, but those settings are not respected. Always the default loading indicator is used.

daniloercoli:
Fix #438 by checking that passed custom loading indicator is not null, and using it. Use default value otherwise.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -733,7 +733,7 @@ class AztecText : AppCompatAutoCompleteTextView, TextWatcher, UnknownHtmlSpan.On
                 }
 
                 override fun onImageLoading(drawable: Drawable?) {
-                    replaceImage(ContextCompat.getDrawable(context, drawableLoading))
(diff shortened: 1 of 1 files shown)


0nko:
It's true those callback parameters weren't used there, so it's correct to check if they're not null & use them. However, I think there is another problem. When you look [here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L182) you'll see that the image loading placeholder is passed as an XML attribute in the layout. This attribute is already [being used in WordPress](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L182), so it should have been already working even without this PR.

daniloercoli:
Yes, it was using the default one in WP. But I wanted to set custom icons for videos and images, and pass drawable with different dimensions. I'm going to submit a PR on the other project soon.

0nko:
:shipit:

PR #444, opened 2017-08-03
Missing hint fix

### LINKED ISSUE #296: Hint Text is Not Shown in Empty Text Field
### Expected
Hint text is shown when text field is empty.

### Observed
Hint text is not shown when text field is empty.  The text field is never truly empty since there is a ZWJ even if no visible text exists.

### Reproduced
1. Open demonstration app.
2. Tap ***HTML*** format button.
3. Delete all text.
4. Tap ***HTML*** format button.
5. Notice text field is empty, but hint text is not shown.

0nko:
Fixes #296.


khaykov:
Looks good! I haven't encountered any issues so far. 👍

Issue #447, opened 2017-08-03
Non-latin characters are copied as HTML-encoded entities


frosty:
Reported by @rachelmcr:

> When copying non-Latin text from the beta editor and pasting it elsewhere, the text is encoded. I reproduced this by copying Japanese text from Aztec and pasting it into Keep; the HTML encoded text is what appears in Keep.

The same happens for emoji, in both the main text editor and the title field.

For example:

> Êtest 😊😃
> ​测试一个

becomes:

`&#202;test &#128522;&#128515;<br>&#27979;&#35797;&#19968;&#20010;`

Issue #285, opened 2017-03-13
Support for copy/pasting of formatted text


0nko:
Add support for for copy/pasting of formatted (such as newlines) text outside the Aztec editor (from Simplenote, for instance). Currently the text is pasted as one large paragraph, even if the original text was split up.

annezazu:
Related: User in 204077-hs suggested adding the ability to clear out formatting after pasting from Word.

0nko:
Fixed in #454.

PR #454, opened 2017-08-07
Copy/paste Fix & Improvement

### LINKED ISSUE #424: Crash report 7.8: RuntimeException in AztecText.paste

0nko:
Fixes #424, #447 and #285.

Summary of things this PR addressed:

- Fixes block element style copying: Headings, lists, quotes are now preserved.
- Copying now works across multiple list hierarchies: When copying inside a nested list, only the minimum required level gets copied. For example, copying a list item nested under 3 levels of lists will copy just the last one. 
- When pasting text next to an inline style, the pasted text is affected by the original text next to it.
- Copy/pasting of exotic unicode characters, such as non-Latin characters and emoji works now (those characters are not escaped anymore during parsing). 
- Copied Calypso paragraphs are also preserved now. 
- The paste logic now tries to look at the type of data being pasted. If it detects styled text (`Spanned`) or HTML, the data is parsed first and the styling is preserved. That means copying text from a website copies the HTML and all the headings, lists, typefaces are copied along.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Fixed
 - Image/video loading placeholder drawable usage
 - The missing hint bug if text is empty
+- Crash when copy/pasting lists
+- Copy/pasting of non-latin unicode characters
 
 ### Added
 - This changelog
@@ -14,6 +16,7 @@
 - Plugin interface for HTML preprocessor
 - Plugin interface for HTML postprocessor
 - Plugin module with `[video]`, `[audio]` and `[caption]` WordPress shrotcode support
+- Copy/pasting of block element styles
 
 ## [1.0-beta.6](https://github.com/wordpress-mobile/AztecEditor-Android/releases/tag/v1.0-beta.6) - 2017-07-25
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@
 ### Fixed
 - Image/video loading placeholder drawable usage
 - The missing hint bug if text is empty
+- Crash in the URL dialog caused by the non plain text content of a clipboard
 - Crash when copy/pasting lists
 - Copy/pasting of non-latin unicode characters
-- Crash in the URL dialog caused by the non plain text content of a clipboard
 
 ### Added
 - This changelog
@@ -17,8 +17,10 @@
 - Plugin interface for HTML preprocessor
 - Plugin interface for HTML postprocessor
 - Plugin module with `[video]`, `[audio]` and `[caption]` WordPress shrotcode support
-- Copy/pasting of block element styles
 - OnMediaDeletedListener interface, detection and handling
+- Copy/pasting of block element styles
+- Copy/pasting of styled text and HTML from external sources
+
 
 ## [1.0-beta.6](https://github.com/wordpress-mobile/AztecEditor-Android/releases/tag/v1.0-beta.6) - 2017-07-25
 ### Changed
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -313,6 +313,9 @@ class MainActivity : AppCompatActivity(),
         val sourceEditor = findViewById(R.id.source) as SourceViewEditText
         val toolbar = findViewById(R.id.formatting_toolbar) as AztecToolbar
 
+        visualEditor.setCalypsoMode(false)
+        sourceEditor.setCalypsoMode(false)
+
         aztec = Aztec.with(visualEditor, sourceEditor, toolbar, this)
             .setImageGetter(PicassoImageLoader(this, visualEditor))
             .setVideoThumbnailGetter(GlideVideoThumbnailLoader(this))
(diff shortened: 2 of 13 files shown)


khaykov:
Great job! Copypasting feels like one of those rabbit holes. 
I'm still testing different scenarios, but so far a couple of things I noticed (but not sure if they are inside the scope of this PR or not):

- After pasting text the cursor is gone.
- Pasting text (regular text without any block style) ads new line around it.
- Inline formatting is not applied to text that is pasted before or after the styled text.
- History is not affected by pasting.
- Pasting Heading into heading adds an extra newline at the top.
- Pasting anything into the middle of heading splits it in half (this is expected) and the bottom half loses the style (unexpected).

0nko:
> - After pasting text the cursor is gone.

This is caused by the `refershText()` call during image loading. We can probably leave it for another PR.
> - Pasting text (regular text without any block style) ads new line around it.

I couldn't reproduce this. Did you turn off Calypso mode?

> - Inline formatting is not applied to text that is pasted before or after the styled text.

> - History is not affected by pasting.

This is now fixed.

> - Pasting Heading into heading adds an extra newline at the top.

It copies the original heading style, which means it's a "block" style and it is, therefore, surrounded by newlines. So the result is the same as parsing `<h1>Heading<h2>Heading 2</h2></h1>`. This is a tough one, because I'm just pasting HTML into HTML, which means I can't detect if the target location is inside a heading. And then the final HTML has both headings split already. Unless someone starts complaining about this, I'd say we leave this :P.

> - Pasting anything into the middle of heading splits it in half (this is expected) and the bottom half loses the style (unexpected).

I couldn't reproduce this either.

By the way, to test the copy/pasting properly you need to do it on a real device. The emulator messes with the data type description and it's always considered a plain text, even if it's styled or HTML.

khaykov:
I'm testing on my Pixel right now, and issues are still there :( Here are exact steps to reproduce them:

> Pasting text (regular text without any block style) ads new line around it.
Open app with demo content, select word "Lorem" and copy it. Select word "Lorem" again and tap "Paste". "Lorem" will move to a separate line. Also, press "Undo" in history and the word "Lorem" will be gone, without any way to return it.

> Inline formatting is not applied to text that is pasted before or after the styled text.
Open app with demo content, select word "Lorem" and copy it. Put cursor right after word "Bold" and paste the word. Notice that "Lorem" goes to the new line and bold is not applied to it.

> Pasting anything into the middle of heading splits it in half (this is expected) and the bottom half loses the style (unexpected).

Open app with demo content, select word "Lorem" and copy it. Put cursor right after "a" in the "Heading 2", paste the word. Notice above mentioned behavior.

Another issue with history - select word "Lorem" and copy it. Paste it in the middle of the word "Bold". Press undo, notice that there is extra space where the word was pasted.

0nko:
All of the issues above should be addressed with the previous commits. 

A couple of notes:

> Also, press "Undo" in history and the word "Lorem" will be gone, without any way to return it.

Since the original word is replaced with the same one, the text isn't changed so there is no history being recorded.

> Put cursor right after word "Bold" and paste the word. Notice that "Lorem" goes to the new line and bold is not applied to it.

When pasting next to inline styles, the original style is explicitly ignored so that it doesn't affect the pasted text. Say you want to copy a couple of italic words and paste it next to some a bold text. As a user, I'd expect the pasted text to be in the same style as it was copied. I would not expect the 1st word to be bold.

khaykov:
Thank's for the fixes! Everything works as expected now :shipit:

PR #461, opened 2017-08-09
Adding DeleteMediaElementWatcher and new OnMediaDeletedListener

mzorz:
### Fix
This PR adds the ability to detect when a media item has been deleted and issues a call to `OnMediaDeletedListener` if such a listener has been set.
This will be needed in WPAndroid if an operation is going on with the given media item (for example, an upload needs to be cancelled when the user deletes such an item).

### Test
1. start the demo app
2. position the cursor after any media item (i.e. audio, video, or image)
3. tap `backspace` to delete it
4. observe the `media deleted` toast is shown

### Review
@0nko

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Plugin interface for HTML preprocessor
 - Plugin interface for HTML postprocessor
 - Plugin module with `[video]`, `[audio]` and `[caption]` WordPress shrotcode support
+- Added OnMediaDeletedListener interface, detection and handling
 
 ## [1.0-beta.6](https://github.com/wordpress-mobile/AztecEditor-Android/releases/tag/v1.0-beta.6) - 2017-07-25
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Plugin interface for HTML preprocessor
 - Plugin interface for HTML postprocessor
 - Plugin module with `[video]`, `[audio]` and `[caption]` WordPress shrotcode support
-- Added OnMediaDeletedListener interface, detection and handling
+- OnMediaDeletedListener interface, detection and handling
 
 ## [1.0-beta.6](https://github.com/wordpress-mobile/AztecEditor-Android/releases/tag/v1.0-beta.6) - 2017-07-25
 ### Changed
--- a/aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
@@ -19,10 +19,10 @@ class DeleteMediaElementWatcher(aztecText: AztecText) : TextWatcher {
         var deletedMedia = count > 0  && containsMediaChars(text, start, count)
 
         if (deletedMedia) {
-            var mediaSpanList = findAllMediaSpansWithinRange(start, count)
-            mediaSpanList?.forEach {
-                it.onMediaDeleted()
-            }
+            aztecTextRef.get()?.text?.getSpans(start, start + count, AztecMediaSpan::class.java)
+                    ?.forEach {
+                        it.onMediaDeleted()
+                    }
         }
     }
 
@@ -41,10 +41,6 @@ class DeleteMediaElementWatcher(aztecText: AztecText) : TextWatcher {
             }
         }
         return false
-    }
-
-    fun findAllMediaSpansWithinRange(start: Int, count: Int): Array<AztecMediaSpan>? {
-        return aztecTextRef.get()?.text?.getSpans(start, start + count, AztecMediaSpan::class.java)
     }
 
     companion object {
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Plugin interface for HTML preprocessor
 - Plugin interface for HTML postprocessor
 - Plugin module with `[video]`, `[audio]` and `[caption]` WordPress shrotcode support
+- Added OnMediaDeletedListener interface, detection and handling
```
0nko: Excellent, you remembered the changelog! :) Just a nitpick - I've been leaving out the verb for clarity (No "Added", "Fixed", "Changed), since it's already obvious by the section title.

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
```diff
@@ -0,0 +1,46 @@
...
+
+class DeleteMediaElementWatcher(aztecText: AztecText) : TextWatcher {
+
+    private val aztecTextRef: WeakReference<AztecText?> = WeakReference(aztecText)
+
+    private var deletedMedia: Boolean = false
+    private var mediaSpan: AztecMediaSpan? = null
+
+    override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) {
+        deletedMedia = count > 0 && text[start + count - 1] == Constants.IMG_CHAR
```
0nko: This won't work if you delete the media along with other characters. Try surrounding an image with text, select it all, delete it and notice the listener's not called.

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
```diff
@@ -0,0 +1,46 @@
...
+    override fun onTextChanged(text: CharSequence, start: Int, before: Int, count: Int) {
+        // no op
+    }
+
+    override fun afterTextChanged(text: Editable) {
+        if (aztecTextRef.get()?.isTextChangedListenerDisabled() ?: true) {
+            return
+        }
+
+        if (deletedMedia) {
```
0nko: Another thing you'll notice is that the listener isn't called when deleting audio media. The problem is that you save the `deletedMedia` flag in `beforeTextChanged()`. However, another text manipulation is triggered (some parser processing) before the `afterTextChanged()` gets called and it changes your `deletedMedia` flag value to `false`. The solution is would be not to store any instance variables in the watcher class but rather save the information somewhere else (maybe keep a list of deleted spans in `AztecText`). I know some of the other watchers have the same issue, I'll need to go back and fix them.

CHANGELOG.md
mzorz: addressed in c76d167

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
mzorz: I think this one in particular could be fixed in a much simpler way by moving detection directly to `beforeTextChanged()` and checking whether the textChangedListener is enabled before continuing, like this:
```
 if (aztecTextRef.get()?.isTextChangedListenerDisabled() ?: true) {
            return
        }
```
addressed in 7008878

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
mzorz: nice catch - addressed in 8aec182. I'm now also considering there might be mutiple media items within the selected text the user is trying to delete, thus triggering each of their listeners as well.

mzorz:
Ready for another round @0nko 🙇

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) {
+        if (aztecTextRef.get()?.isTextChangedListenerDisabled() ?: true) {
+            return
+        }
+
+        var deletedMedia = count > 0  && containsMediaChars(text, start, count)
+
+        if (deletedMedia) {
+            var mediaSpanList = findAllMediaSpansWithinRange(start, count)
```
0nko: That works. But you can simplify the whole thing with just
```
aztecTextRef.get()?.text?.getSpans(start, start + count, AztecMediaSpan::class.java)
        ?.forEach {
            it.onMediaDeleted()
        }
```

mzorz: sure - wanted to have that separated but it surely looks slimmer this way. Addressed in 02c1723

0nko:
Looking good! :shipit:

Issue #237, opened 2017-02-08
Can't remove styling when writing text


daniloercoli:
### Expected
To be able to remove strikethrough and keep writing in "plain" mode.

### Observed
https://cloudup.com/cQCMPkCJvZ2

### Reproduced
1. Write some text. 
2. Apply strikethrough and write some text.
3. Remove strikethrough, and wait a bit.
4. Continue to write without clicking on the space bar.
5. New text has still the strikethrough applied to it.

If you use the backspace char, as soon as type on it, the strikethrough is highlighted in the button bar. Also noticed that the first character is not deleted.

#### Tested
Nexus 5X on 7.1.1 with Integration Branch

theck13:
This is related to suggestions.  Adding `android:inputType="textNoSuggestions"` to the [`AztecText`](https://github.com/wordpress-mobile/WordPress-Aztec-Android/blob/develop/app/src/main/res/layout/activity_main.xml#L26) view eliminates the issue.

PR #466, opened 2017-08-14
Issue/464 missing suggestions

### LINKED ISSUE #464: Missing suggestions
The editor should always display suggestions/autocorrect (if not disabled). On some devices it's not the case and while it works in other apps, it doesn't in Aztec.

Tested and reproduced on HTC Incredible S.

0nko:
Fixes #464. This brings back #237.


khaykov:
There is another, a bit more serious side effect of #237 - when you delete the last character of styled text the span is removed from it. Open the app with demo content, put cursor behind "Italic" and press backspace on the software keyboard.

I researched this problem a long time ago, and it is indeed caused by autocorrect/suggestion (the suggested text becomes Underlined and all the other spans are removed from it). That's why we reapplying the spans [here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/watchers/InlineTextWatcher.kt#L52). But apparently it did not work :) I'll take a look at #237 - maybe it will be a quick fix.

khaykov:
Another issue - type some letters, toggle inline style, type some more letters, press space. Notice the inline style disappeared.

0nko:
Hmm, that's pretty serious. Thanks for pointing that out, Klym.

maxme:
@khaykov I can't reproduce issues you described. Is this specific to a certain android version or device? I tested on this branch, I used Android 5.0 and 7.0.

khaykov:
@maxme Yes, those issues are from pre 5.0.0 devices.

From the top of my head here is a couple of issues you can reproduce on 5.0.0+ devices (tested on Pixel_API_25 emulator):

Load empty editor, type "h" toggle bold, type "ello" notice that after you finish typing only the last letter will have a bold style applied to it.

Toggle bold, and type some incorrect word, like "herro", press space so it will be autocorrected to "hero", press backspace to cancel autocorrect and notice that the bold style is also gone.

Also, autocorrect only works with soft keyboard, so make sure to use it.

khaykov:
Found another issue that I can only reproduce on API 26 so far:
Load empty editor, type "a" toggle Bold style, type "b". Put the cursor between "a" and "b" and press space. Notice that "b" became unstyled.

:(

mzorz:
- tests run alright for me
- Tested behaviour described in #237 on Google Pixel, Android 8, works fine.
- Tested on Nexus 6P, API 25, works fine
- Tested on a geny image of Samsung Galaxy S3, Android 4.2.2, the behavior described in #237 is not observed here, but a curious problem appears, that I did not observe on the base branch (`develop`): there are no newlines ,it seems everything is on the same line, and tapping `return` on the soft keyboard doesn't have any effects. See the gif below showing that I navigate the contents by moving the cursor to the right only:

aztec_horizontal

mzorz:
Just for the record: this comment https://github.com/wordpress-mobile/AztecEditor-Android/pull/466#issuecomment-344315573 in this PR doesn't block the review/merge.

mzorz:
@0nko I confirm I can reproduce this issue here https://github.com/wordpress-mobile/AztecEditor-Android/pull/466#issuecomment-338112619 

We can merge this one once the conflicts are resolved, going to open new issues with whatever is found (that comment for example, plus the all-content-in-one-line issue from 4.2.2 )

mzorz:
:shipit:

PR #467, opened 2017-08-15
Gradle upgrade

0nko:
Since Android Studio 3 is close to being released I think we can start using it (I have for a while now) and we can update the Gradle tools to take advantage of some of the improvements. 

This PR upgrades the Gradle wrapper to version 4.1 and Gradle plugin to 3.0.0-beta2. Some of the dependencies have been updated and and I added some small refactoring.

Main benefits of the new Gradle version include
- New `implementation` dependency configuration that removes transitive dependencies from the library clients (this allows us to drop the `exclude module: 'aztec'` from wpandroid)
- New `api` configuration that explicitly includes the transitive dependency
- Fixed some bugs that allow us to drop workarounds (like `'org.khronos:opengl-api:gl1.1-android-2.1_r1'`)
- Improved build speed


0nko:
Ready for a check! :)

0nko:
It looks like it's a problem with Jitpack, it's not including the `implementation` dependencies in the runtime classpath. When I tried using Jitpack references in the demo app it stopped working and when I published Aztec into a local Maven repo it started working again. So we can just add the missing dependencies in wpandroid until they fix it.

khaykov:
:shipit:

PR #468, opened 2017-08-16
Aztec constructor without source editor

### LINKED ISSUE #446: Add option to use Aztec without source editor
The current configuration API assumes the client wants to use both the visual and source code editor. However, this might not be the case and we need to add an option to initialize just the visual editor by itself in `Aztec`.

mzorz:
Fixes #446 

This PR aims at making available a simple constructor for an instance that doesn't need a `SourceViewEditText`, `AztecToolbar`  nor `IAztecToolbarClickListener` for it to exist.

### Review
@0nko


0nko:
You've mentioned you wanted headless Aztec, but you're still using an instance of `AztecText`, which is a UI view. For the purpose of just parsing HTML, you can use an instance of `AztecParser` (which is used internally by `AztecText`). But that's just FYI, we do want to be able to use `AztecText` without the toolbar or `SourceViewEditText` ;).

mzorz:
Thank you for the explanation and more elsewhere! Changing the name of this PR to better describe it :)

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -44,6 +51,10 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
                  toolbar: AztecToolbar, toolbarClickListener: IAztecToolbarClickListener) : Aztec {
             return Aztec(visualEditor, sourceEditor, toolbar, toolbarClickListener)
         }
+
+        fun with(visualEditor: AztecText) : Aztec {
```
0nko: I think it would be useful to add an argument for the toolbar and toolbar listener (with default null value), so that you can use the visual editor with or without a toolbar.

mzorz: addressed in f1fa2b5

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -104,20 +133,20 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
...
     fun initHistory() {
-        sourceEditor.history = visualEditor.history
+        sourceEditor?.history = visualEditor.history
     }
 
     private fun initToolbar() {
-        toolbar.setEditor(visualEditor, sourceEditor)
-        toolbar.setToolbarListener(toolbarClickListener)
-        visualEditor.setToolbar(toolbar)
+        toolbar?.setEditor(visualEditor, sourceEditor!!)
```
0nko: `sourceEditor` can be null, this call can result in NPE.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -104,20 +133,20 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
...
+        toolbar?.setToolbarListener(toolbarClickListener!!)
```
0nko: Same here.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
0nko: The `setEditor()` function will need to be modified so that it can handle the case without the source editor.

mzorz:
thank you for the comments! addressed them in cbcaef1 - ready for another round @0nko 👍

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -312,7 +312,7 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
         return editor != null && editor is AztecText
     }
 
-    fun setEditor(editor: AztecText, sourceEditor: SourceViewEditText) {
+    fun setEditor(editor: AztecText, sourceEditor: SourceViewEditText?) {
```
0nko: Unfortunately this isn't enough. You'll need to change all of the `sourceEditor!!` to `sourceEditor?`.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -10,8 +10,8 @@ import org.wordpress.aztec.toolbar.AztecToolbar
 import org.wordpress.aztec.toolbar.IAztecToolbarClickListener
 import java.util.*
 
-open class Aztec private constructor(val visualEditor: AztecText, val sourceEditor: SourceViewEditText,
-                                     val toolbar: AztecToolbar, val toolbarClickListener: IAztecToolbarClickListener) {
+open class Aztec private constructor(val visualEditor: AztecText, val toolbar:
+                            AztecToolbar? = null, val toolbarClickListener: IAztecToolbarClickListener? = null) {
```
0nko: I've realized that it actually doesn't make sense to have a visual editor without a toolbar, does it? If you agree then the `toolbar` and `toolbarClickListener` can actually be made non-nullable (there is no `with()` function that allows this anyway). Sorry for the confusion... 

We can also make `toolbarClickListener` private, since listener getters are not needed.

Nitpick: For readability I think it's better to put the entire parameter declaration on one line (i.e. the `toolbar` parameter)

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -23,17 +23,29 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
...
     private var plugins: ArrayList<IAztecPlugin> = visualEditor.plugins
+    var sourceEditor: SourceViewEditText? = null
 
     init {
         initHistory()
         initToolbar()
     }
-    
+
     constructor(activity: Activity, @IdRes aztecTextId: Int,
```
0nko: All constructors should be private as we only want to use the factory to create an instance.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -23,17 +23,29 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
...
                 @IdRes sourceTextId: Int, @IdRes toolbarId: Int,
                 toolbarClickListener: IAztecToolbarClickListener) : this(activity.findViewById(aztecTextId) as AztecText,
             activity.findViewById(sourceTextId) as SourceViewEditText, activity.findViewById(toolbarId) as AztecToolbar, toolbarClickListener)
 
+
+    constructor(activity: Activity, @IdRes aztecTextId: Int,
```
0nko: Here, as well.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -23,17 +23,29 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
...
+                @IdRes toolbarId: Int,
+                toolbarClickListener: IAztecToolbarClickListener) : this(activity.findViewById(aztecTextId) as AztecText,
+            activity.findViewById(toolbarId) as AztecToolbar, toolbarClickListener)
+
+    constructor(visualEditor: AztecText, sourceEditor: SourceViewEditText,
```
0nko: Here, as well.

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -55,7 +55,7 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
         initView(attrs)
     }
 
-    fun setToolbarListener(listener: IAztecToolbarClickListener) {
+    fun setToolbarListener(listener: IAztecToolbarClickListener?) {
```
0nko: Not necessary if we always need a toolbar.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -23,17 +23,29 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
...
+                toolbar: AztecToolbar, toolbarClickListener: IAztecToolbarClickListener) : this (visualEditor, toolbar, toolbarClickListener) {
```
0nko: Nitpick: Should be `this(visualEditor...`

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
mzorz: addressed in 3788d02

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
mzorz: All of those occurrences where in `toggleEditorMode`, so I decided to just check for `sourceEditor` presence right away on that method, here 31fa909

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
mzorz: 👍  addressed in 55a626b

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
mzorz: addressed in c245dc7

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
mzorz: removed extra space in 8525691

mzorz:
Ready for another round @0nko 🙇 thanks for all the tips!

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -104,19 +120,19 @@ open class Aztec private constructor(val visualEditor: AztecText, val sourceEdit
         plugins.add(plugin)
 
         if (plugin is IToolbarButton) {
-            toolbar.addButton(plugin)
+            toolbar?.addButton(plugin)
```
0nko: You can change all occurrences of `toolbar?.` in this class to just `toolbar.`.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -673,7 +673,7 @@ class AztecText : AppCompatAutoCompleteTextView, TextWatcher, UnknownHtmlSpan.On
         }
     }
 
-    fun setToolbar(toolbar: AztecToolbar) {
+    fun setToolbar(toolbar: AztecToolbar?) {
```
0nko: This does not need to be nullable, since this call will always be called with a non-null value.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -340,11 +340,11 @@ class MainActivity : AppCompatActivity(),
 
         // initialize the text & HTML
         if (!isRunningTest) {
-            aztec.sourceEditor.displayStyledAndFormattedHtml(EXAMPLE)
+            aztec.sourceEditor?.displayStyledAndFormattedHtml(EXAMPLE)
         }
 
         if (savedInstanceState == null) {
-            aztec.visualEditor.fromHtml(aztec.sourceEditor.getPureHtml())
+            aztec.visualEditor.fromHtml(aztec.sourceEditor?.getPureHtml()!!)
```
0nko: This may crash if we don't initialize the source editor (the effect of using `?.` is that a `null` value is returned if  `sourceEditor` is `null`). Since this is just a demo and we know that we will have a source editor, this fine. But using `!!` is dangerous and it should only be used when you are certain the expression can never be `null`.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
mzorz: addressed in eaa4a44

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mzorz: addressed in df77ea4

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
mzorz: thanks for the explanation! Learning a lot from your reviews 🙇 
What would we need to do, if we wanted to do something about it?

0nko: No problem ;). 

You'd have to add a check so that `fromHtml()` method is not called with a `null` value (or change it so that it can handle it). In this case it wouldn't we'd need to initialize the visual editor with `EXAMPLE`.

0nko:
Seems like Travis is complaining about the generic `findViewById()`. I'm not sure why since you're using support lib v25. Try merging in from `develop` and see if it gets fixed.

0nko:
Good job! 🥇

PR #470, opened 2017-08-25
Fix broken HTML button

0nko:
The primary `Aztec` constructor is called before `sourceEditor` gets initialized by a factory method, so it's always `null` in `AztecToolbar`. This PR fixes this.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
@@ -44,7 +44,7 @@ open class Aztec private constructor(val visualEditor: AztecText, val toolbar: A
         this.sourceEditor = sourceEditor
 
         initToolbar()
-        initHistory()
+        initSourceEditorHistory()
     }
 
     companion object Factory {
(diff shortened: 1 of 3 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -43,6 +42,9 @@ open class Aztec private constructor(val visualEditor: AztecText, val toolbar: A
     private constructor(visualEditor: AztecText, sourceEditor: SourceViewEditText,
                 toolbar: AztecToolbar, toolbarClickListener: IAztecToolbarClickListener) : this(visualEditor, toolbar, toolbarClickListener) {
         this.sourceEditor = sourceEditor
+
+        initToolbar()
```
mzorz: shouldn't we remove this line? As it's already executed in the `init{}` block

0nko: No, that's the fix. The first time `initToolbar()` is callled `sourceEditor` is not initialized yet.

mzorz: oh what a fail from my side! 😂 I thik I'd better leave this for someone else to review

khaykov:
Looks good! 👍

PR #474, opened 2017-09-01
Try to fix OOM errors with several pictures

daniloercoli:
### Fix420

This is a tentative to fix OOM issues with several pictures by using lazy image provider for drawables, and disposing drawables that are not in the viewport.

Didn't test it extensively on a good range of devices, and there are still some optimization to add to this code, but the general idea is already there.

### Review
@0nko could you please take a look at this PR and check if it's going in the right direction, and works for you?


aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecMediaSpan.kt
```diff
@@ -1,35 +1,50 @@
...
+    }
 
-        setInitBounds(newDrawable)
+    fun setDrawable(newDrawable: Drawable?, isPlaceholder: Boolean = false) {
+        if (!isPlaceholder) {
+            innerPlaceholder = drawableToBitmap(newDrawable)
+        }
 
-        computeAspectRatio()
+        if (!isPlaceholder || innerPlaceholder == null) {
```
daniloercoli: @0nko I'm not sure the ` innerPlaceholder == null` is required here. Maybe it's not used in our projects, but we can change/switch the drawable of a `MediaSpan` resulting in an invalid `innerPlaceholder` since it was computed on the previous picture.

0nko: ok :)

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecMediaSpan.kt
```diff
@@ -1,35 +1,50 @@
...
-                              editor: AztecText? = null) : AztecDynamicImageSpan(context, drawable), IAztecAttributedSpan {
+                              editor: AztecText? = null) : AztecDynamicImageSpan(context, imageProvider), IAztecAttributedSpan {
 
     abstract val TAG: String
 
     private val overlays: ArrayList<Pair<Drawable?, Int>> = ArrayList()
 
+    private var innerPlaceholder: Bitmap? = null
+    private var isDirty = false
+    private val EXTRA_LOADING_SIZE = 500
```
daniloercoli: What's about increasing this to 1000?

0nko: That would mean 2000px + the size of the edit text worth of images will be loaded. Can the most low-end devices handle it? cc: @mzorz

0nko:
Hey @khaykov! Since both @daniloercoli and I worked on this PR, could you please take a look at it as an unbiased observer? :)

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -257,6 +258,18 @@ class AztecText : AppCompatAutoCompleteTextView, TextWatcher, UnknownHtmlSpan.On
...
         isViewInitialized = true
+
+        viewTreeObserver.addOnScrollChangedListener {
+            if (this@AztecText.visibility == View.VISIBLE) {
+                if (invalidateMediaRunnable != null) {
+                    invalidateMediaHandler.removeCallbacks(invalidateMediaRunnable)
+                }
+                if (this@AztecText.text.getSpans(0, text.length, AztecMediaSpan::class.java).isNotEmpty()) {
+                    invalidateMediaRunnable = Runnable { this@AztecText.refreshText() }
+                    invalidateMediaHandler.postDelayed(invalidateMediaRunnable, 100L)
```
khaykov: Why are we refreshing text during scrolling? This causes the cursor to reset, the autocorrect rechecks whole text again, images will get reset too, which sometimes causes the scroll to"jump". Context menus are dismissed too.

daniloercoli: Yeah, sorry. It was a short and dirty way to call a redraw over all pictures.
We needed a way to call a redraw over all pictures in the editor. Pictures outside the bounds are set to null and the other redrawn. I guess there is a better way to do that (looping over all AztecMediaSpan??) Probably @0nko knows a better solution to the problem.

khaykov: How about just reseting spans? Ideally, we will only do it for spans that need it. Something like this:
```
editableText.getSpans(0, text.length, AztecMediaSpan::class.java).filter { it.isDirty }.forEach {
                    val spanStart = editableText.getSpanStart(it)
                    val spanEnd = editableText.getSpanEnd(it)
                    val flags = editableText.getSpanFlags(it)

                    invalidateMediaRunnable = Runnable { this@AztecText.editableText.setSpan(it, spanStart, spanEnd, flags) }
                    invalidateMediaHandler.postDelayed(invalidateMediaRunnable, 100L)
                }
```

daniloercoli: I've fixed it in https://github.com/wordpress-mobile/AztecEditor-Android/pull/474/commits/26b6b4d64b856218e633965426e68c2bfa110588 

Tested it and seems to work. Will let @0nko to run another round of testing with other devices.

0nko: Unfortunately, this approach seems to be small jumps in spans when scrolling and it is most visible on the block quote:

screenshot_1504774806

khaykov: I think it's because we are jamming all the span transactions into one Runnable. It works ok for me without it.

daniloercoli: I've tried different solutions but none of them completely fix the issues. 

- Setting the spans without using runnable isn't a viable solutions since the `OnScrollChangedListener` is called so many times, and there are a lot of redrawing involved while scrolling making it shuttering on lower devices or emulators.

- using 1 runnable for 1 span doesn't fix the issue. since the redraw is performed at the end of the scrolling and you will see the jump. Redraw is performed at the end of the scrolling since we remove pending runnable with `invalidateMediaHandler.removeCallbacks(null)`.
If we don't remove old callbacks the problem is then fixed but we're falling on performance issues like above, since too many runnables are created. 

I've then decided to use a "mixed" solution, re-set spans inline, each 10 cycle, and use runnable to make the final set/redraw. https://github.com/wordpress-mobile/AztecEditor-Android/commit/aa0942e052ec7c80f0ddd6e09962ec05ee3ffcf8

*Note that the problem is not in the image optimization code, I've disabled it and saw the jump on the first layout/redraw^. The underlying problem lies in the code below:
```
editableText.getSpans(0, text.length, AztecMediaSpan::class.java).forEach {
                                val spanStart = editableText.getSpanStart(it)
                                val spanEnd = editableText.getSpanEnd(it)
                                val flags = editableText.getSpanFlags(it)
                                editableText.setSpan(it, spanStart, spanEnd, flags)
                            }```

Don't know why for some strage reason some spans (it seems underline and quote in my case) change their dimensions. You can see this noticeable on the background box of the quote, that sometimes changes its size and there is 1px outside the correct position.


Anyway, this seems to be to most reliable and less hack-ish solution I came up with. We can change the cycle to something less than 10 if you still see the jump.

0nko:
I'm going to close this for now as @daniloercoli has made some improvements without this PR and the performance is acceptable. We can always come back and reopen it.

PR #476, opened 2017-09-05
Issue/fixing spellcheck removing styling

khaykov:
This PR should fix inline styling issues when suggestions enabled.

I tried unsuccessfully to find a way around the problem, but the only solution that kind of works is an improvement on previous suggestion event detection functionality (```carryOverInlineSpans/reapplyCarriedOverInlineSpans```).

Pre lollipop devices have all the issues related to suggestion framework, and post-Lolipop have only a subset of issues. So it makes sense to test it on devices with Android OS < 5.0.0

To test - open editor and type/delete styled and mix-styled text for some time to make sure styles are not disappearing or changing to something different.

~~There is one known issue on **pre 5.0.0** devices:~~

~~- Type incorrect word that will be a target for the suggestion framework.~~
~~- Move the cursor into the middle of the word (make sure the word is still marked for a suggestion - this doesn't happen every time).~~
~~- Press backspace and notice that cursor moves to the end of the word and deletes the last character.~~


0nko:
I found some quirky style behavior:

1. Select **Bold** and type "a"
2. Unselect **Bold**, select **Italic** and type "b"
3. Add a single space (**Italic** style is still selected)
4. Add **Bold** style and type "c" (the thext should look like "**a**_b_ _**c**_")
5. Add a single space
6. Notice that "b" becomes bold ("**a**_**b**_ _**c**_ ")

0nko:
Also, don't forget to update the changelog ;)

0nko:
Excellent! 🚢

PR #479, opened 2017-09-08
Updated UI tests

MatisseHack:
* Updated UI tests to use the page object pattern
* Added image tests
* Added mixed list test

### Test
1. Run all tests in androidTest


0nko:
When I run the unit tests, `testMixedList()` is failing because the expected value doesn't reflect the actual behavior. When you tap a single item in a list and change the list type, the entire list changes, which is the correct way.

MatisseHack:
I was under the impression that only the list item(s) selected should change type. If the whole list should change, then that is inconsistent with iOS. Not sure which is correct, but it's an easy fix either way.

cc @aerych

aerych:
Thanks @0nko.  @MatisseHack if its no trouble could we update the test as @0nko described?

MatisseHack:
No trouble at all

0nko:
Thanks @MatisseHack! Looks good. I'll go ahead and merge this.

:shipit:

aerych:
Thank you @0nko & @MatisseHack :)

PR #483, opened 2017-09-26
Issue/480 missing paragraph

### LINKED ISSUE #480: Paragraph tag is getting removed
The `<p>` tag is  being removed during parsing. It should be preserved as it may contain important attributes and Aztec should not mess with the HTML in general.

This keeps happening only when Calypso mode is enabled.

0nko:
Fixes #480.


khaykov:
I found an issue with paragraph extending on a new line.

Load editor with something like this `<p class="super">paragraph</p>`
Press enter at the end of the line.
Type "test".
Switch to HTML mode and notice that paragraph tag extended to the new line, like this:
```
<p class="super">paragraph

test</p>
```
Switch back to visual and to HTML again, notice HTML became:
```
<p class="super">paragraph</p>
test
```

0nko:
Ready for a new pass.

aztec/src/main/kotlin/org/wordpress/aztec/source/Format.kt
```diff
@@ -324,20 +325,36 @@ internal object Format {
                 val spanEnd = text.getSpanEnd(it)
 
                 if (text[spanStart] == '\n' && text.getSpans(spanEnd, spanEnd + 1, IAztecParagraphStyle::class.java)
-                        .filter { it !is ParagraphSpan && text.getSpanStart(it) == spanEnd }.isEmpty()) {
+                        .filter { (it !is ParagraphSpan || !it.attributes.isEmpty()) && text.getSpanStart(it) == spanEnd }.isEmpty()) {
```
maxme: Just to double check, this only retains the `<p>` if it contains at least one attribute, right?

0nko: That's right.

maxme: Ok, I updated `testRetainParagraphFormatting` in 90b4b7e

maxme:
Looks good, :shipit:

PR #484, opened 2017-09-26
Resize pictures to smaller size (800px)

daniloercoli:
With this PR pictures added to the editor are resized to 800px wide (maximum) before inserting them.
Creating bitmaps at 800px wide saves a lot of memory and the editor is way much more responsive if you try to load several pictures into it. This is a first step for #420 

Note: Bitmap are downloaded from the network at 800px maximum, (resized to 800px max if taken from the camera) but the drawing code keeps changing the density 
```
 override fun onBitmapLoaded(bitmap: Bitmap?, from: Picasso.LoadedFrom?) {
                bitmap?.density = DisplayMetrics.DENSITY_DEFAULT
                callbacks.onImageLoaded(BitmapDrawable(context.resources, bitmap))
                targets.remove(source)
            }
```
and then the bitmap are drawn to fill the entire width of the screen.  You can't even tell that it was resized to 800px. 

### Review
@0nko @khaykov

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -240,7 +240,7 @@ class AztecText : AppCompatAutoCompleteTextView, TextWatcher, UnknownHtmlSpan.On
         styles.recycle()
 
         // set the pictures max size to the min of screen width/height and DEFAULT_IMAGE_WIDTH
-        var minScreenSize = Math.min(context.resources.displayMetrics.widthPixels,
+        val minScreenSize = Math.min(context.resources.displayMetrics.widthPixels,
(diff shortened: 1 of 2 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -235,6 +239,11 @@ class AztecText : AppCompatAutoCompleteTextView, TextWatcher, UnknownHtmlSpan.On
 
         styles.recycle()
 
+        // set the pictures max size to the min of screen width/height and DEFAULT_IMAGE_WIDTH
+        var minScreenSize = Math.min(context.resources.displayMetrics.widthPixels,
```
0nko: You can make this a `val`.

0nko:
:shipit:

PR #486, opened 2017-09-28
Use GlideImageLoader in demo app

daniloercoli:
This PR use GlideImageLoader and get rid of Picasso Image Loader.


0nko:
Looks good 👍

PR #495, opened 2017-10-13
Use a single instance of loading drawable

daniloercoli:
This PR optimizes memory usage when loading remote posts with several pictures.

- Makes sure that AztecTagHandler uses a single instance of loading drawable for all the media tag. No need to re-create a new loading drawable each time. This drawable are used just for a small amount of time, before Glide kicks in.

- This PR also makes sure that AztecText uses a single instance of loading drawable on media. When starting loading remote URLs, we're now passing a single instance of the loading drawable, instead before we were re-creating it for each image.


khaykov:
Looks good to me 👍

PR #504, opened 2017-10-18
Update README: simplify command to run connected tests

maxme:
@0nko I'm not sure why we use `adb shell` directly to run connected tests, maybe I'm missing something.


0nko:
I can't remember. I think we just copied it when we first started and it worked. `gradlew cAT` is definitely simpler :). :shipit:

PR #506, opened 2017-10-19
New tests to reproduce known issues and minor fixes in the test suite

maxme:
I bumped Espresso version to 3.0.1 so we can use [GrantPermissionRule](https://developer.android.com/reference/android/support/test/rule/GrantPermissionRule.html). `testAddPhoto` was failing locally on Android >= 6.0, because permissions were reset every time the test app was installed (before the tests are run).

I also dropped usage of `Optional` in `BetterClickAction`, not because I don't want it, but because Espresso 3.0.1 doesn't come with guava anymore and I don't want to introduce a new test dependency just for one field.

Note: I removed `Created by` comments since we don't use them in our apps or libraries.


app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/SimpleTextFormattingTests.kt
```diff
@@ -310,4 +306,37 @@ class SimpleTextFormattingTests : BaseTest() {
...
+                .verifyHTML(html)
+    }
+
+    // Test reproducing the issue described in
+    // https://github.com/wordpress-mobile/AztecEditor-Android/pull/466#issuecomment-322405856
+    @Test
+    fun testInlineStyleAndSpace() {
+        val text1 = "some"
+        val text2 = "text"
+        val text3 = " "
```
rachelmcr: Any reason not to combine `text2` and `text3` into a single string `text ` (with the space at the end of that string, rather than entering the space separately)?

maxme: Keyboard is dismissed between two calls to `insertText`, honestly I'm not sure if that changes anything. I tried both methods and they both pass.

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
```diff
@@ -107,4 +107,23 @@ class MixedTextFormattingTests : BaseTest() {
                 .verifyHTML(text)
     }
 
+    // Test reproducing the issue described in
+    // https://github.com/wordpress-mobile/AztecEditor-Android/pull/476#issuecomment-327762497
+    @Test
+    fun testMixedBoldItalics() {
```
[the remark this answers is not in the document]
maxme: Ah right I didn't notice this, but this is strange: `testBoldAndItalicFormatting` passes on all device I tested, and the expected HTML tags are always in that order `<b><i>$text3</i></b>` (italics _inside_ bold). That's not the case for `testMixedBoldItalics`, on API 16 I get `<i><b>c </b></i>` and API > 16 I get `<b><i>c </i></b>` (and that's why I used a Regex here).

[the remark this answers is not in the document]
rachelmcr: Hmmm, odd. I consistently get `<b><i>c </i></b>` even on API 16. Since I can't reproduce the issue on my end I'm not sure what's going on, but I wonder if that's related to the separate spaces/the keyboard being dismissed between the calls to `insertText`? That's the only significant difference I see in the test steps.

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/SimpleTextFormattingTests.kt
rachelmcr: Ok. Not dismissing the keyboard seems closer to the behavior of a user entering text in the editor, but given that it doesn't make a difference to the test I don't have a strong opinion about it. :)

maxme: Fixed in ee5dde0

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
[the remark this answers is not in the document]
maxme: Dropped this test and moved the comment in 1ff934b

maxme:
@rachelmcr I ran the tests in Firebase and found this:

```
android.support.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError: 'with text: is "<b>so</b><i>me </i><b><i>text </i></b>"' doesn't match the selected view.
Expected: with text: is "<b>so</b><i>me </i><b><i>text </i></b>"
     Got: "SourceViewEditText{id=2131231019, res-name=source, visibility=VISIBLE, width=1080, height=1458, has-focus=true, has-focusable=true, has-window-focus=true, is-clickable=true, is-enabled=true, is-focused=true, is-focusable=true, is-layout-requested=false, is-selected=false, layout-params=android.widget.FrameLayout$LayoutParams@b5447ad, tag=null, root-is-layout-requested=false, has-input-connection=true, editor-info=[inputType=0xa0001 imeOptions=0x50000006 privateImeOptions=null actionLabel=null actionId=0 initialSelStart=0 initialSelEnd=0 initialCapsMode=0x0 hintText=Codex is the source label=null packageName=null fieldId=0 fieldName=null extras=null hintLocales=null contentMimeTypes=null ], x=0.0, y=0.0, text=<b>so</b><i>me </i><i><b>text </b></i>, hint=Codex is the source, input-type=655361, ime-target=true, has-links=false}"
```

So the original test you wrote (mixing bold and italics) apparently has the same issue (bold inside italics instead of italics inside bold), I'm not sure why there is this difference. It doesn't matter, so I'll update the check with a Regex.

PR #507, opened 2017-10-19
Fix Multi-digit ordered list items rendering issue

### LINKED ISSUE #496: Multi-digit ordered list items not properly displayed
### Expected
To be able to visualize an ordered list of arbitrary size

### Observed
Any list item >9 is cut off.  
It doesn't show up in the screenshot below, but on the live app if you look _really_ closely you can see a single pixel of the first digit displayed, but that's it.  The problem exists in both landscape and portrait.

image

### Reproduced
1. Start a new post
1. Create an ordered list
1. Add at least 10 elements

#### Tested
Nexus 6P on Android 8.0.0, WPAndroid alpha-79 (with Beta editor turned on)

daniloercoli:
### Fix #496 
by reducing the text size used to draw the item number.

It feels like an hacking solution but I didn't find a way to adapt the leading margin to the text being drawn on the screen. `getLeadingMargin` doesn't have any clue of the text being drawn on the screen, and thus we can't change the margin by adapting it to text. We should probably calculate the required space for the margin before, and pass it to `AztecOrderedListSpan` in the listStyle parameter.  

### Test
Change the demo app, and add a bunch of `li` in a `ol`.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecOrderedListSpan.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecOrderedListSpan.kt
@@ -28,7 +28,7 @@ class AztecOrderedListSpan(
         override var nestingLevel: Int,
         override var attributes: AztecAttributes = AztecAttributes(),
         var listStyle: BlockFormatter.ListStyle = BlockFormatter.ListStyle(0, 0, 0, 0, 0)
-    ) : AztecListSpan(nestingLevel, listStyle.verticalPadding) {
+) : AztecListSpan(nestingLevel, listStyle.verticalPadding) {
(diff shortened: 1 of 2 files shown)


iamthomasbishop:
Would simply adding more margin-left on all lists fix this issue? I'm thinking we could shift all lists ~5-8dp to the right and also tighten the spacing between bullet/num and list item text, but that's subjective.

daniloercoli:
That could be a solution, but if you change the text size in device settings (or your device has a bigger text size set by default) we can fall on the same problem.

Btw, I can try to shift them 5-8dp, in addition to the proposed solution above.

daniloercoli:
Closing this PR since we've fixed the issue in a nicer way in #518

PR #510, opened 2017-10-23
Small images cleanup

daniloercoli:
This PR doesn't fix anything in particular, but removes image related code that I don't think it's necessary. Referring to `calculateWantedWidth` routine.

Maybe @khaykov has some more background/details about why it was introduced.


khaykov:
Good catch, @daniloercoli ! I think `calculateWantedWidth` is just an artifact from the past.

PR #518, opened 2017-11-08
Issue/496 experiment fix

### LINKED ISSUE #496: Multi-digit ordered list items not properly displayed

0nko:
Fixes #496.

The solution shifts the numbers to the right if it can't fit all digits. The leading margin is shifted too, so that the numbers are all right-aligned with the items' text.

image

image

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecOrderedListSpan.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecOrderedListSpan.kt
@@ -32,6 +32,7 @@ class AztecOrderedListSpan(
     override val TAG = "ol"
 
     private var horizontalShift = 0
+    private var maxWidth = 0f
 
     override fun getLeadingMargin(first: Boolean): Int {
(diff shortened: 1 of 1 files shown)


daniloercoli:
I think this is a good solution. Even though on huge lists the first items are shifted far to the right (or left, depending on the language orientation). I guess this is not a problem since it's a edge case. cc @iamthomasbishop  (Maybe we should dispose the right alignment of bullets in the list?)

With the merge of this PR, I'm also going to close the other PR (https://github.com/wordpress-mobile/AztecEditor-Android/pull/507) that does reduce the size of the numbered bulled. cc /

Issue #306, opened 2017-04-07
Image upload crash


0nko:
Copy of https://github.com/wordpress-mobile/WordPress-Android/issues/5551.

Issue #299, opened 2017-03-28
Crash report wpandroid alpha-42: Undo while uploading image crashes app


maxme:
```
Caused by java.lang.NullPointerException: Attempt to invoke interface method 'int org.xml.sax.Attributes.getLength()' on a null object reference
       at org.ccil.cowan.tagsoup.AttributesImpl.setAttributes(AttributesImpl.java:356)
       at org.ccil.cowan.tagsoup.AttributesImpl.<init>(AttributesImpl.java:83)
       at org.wordpress.android.editor.AztecEditorFragment$AttributesWithClass.<init>(AztecEditorFragment.java:636)
       at org.wordpress.android.editor.AztecEditorFragment.onMediaUploadProgress(AztecEditorFragment.java:590)
       at org.wordpress.android.ui.posts.EditPostActivity.onUploadProgress(EditPostActivity.java:990)
       at org.wordpress.android.ui.posts.EditPostActivity.onMediaUploaded(EditPostActivity.java:2299)
       at java.lang.reflect.Method.invoke(Method.java)
       at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:485)
       at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:479)
       at org.greenrobot.eventbus.HandlerPoster.handleMessage(HandlerPoster.java:67)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:154)
       at android.app.ActivityThread.main(ActivityThread.java:6119)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

0nko:
The attribute retrieval updated & made safer to use (it does not return `null` anymore) by the [attributes PR](https://github.com/wordpress-mobile/AztecEditor-Android/pull/321) and the crash does not occur anymore.

PR #521, opened 2017-11-10
Add high priority and mixed formatting UI tests

rachelmcr:
1. Adds UI tests for high priority issues #306 and #299.
2. Adds more UI tests for mixed formatting. 
3. Strips whitespace between HTML tags when verifying the HTML output (to match the iOS test behavior in https://github.com/wordpress-mobile/AztecEditor-iOS/pull/817).

### To Test

Build and run the instrumentation tests, e.g. with `./gradlew cAT` ([see README for details](https://github.com/wordpress-mobile/AztecEditor-Android/tree/add/high-priority-ui-tests#build-and-test)).

Note that `testRetainParagraphFormatting` currently fails but will be resolved by https://github.com/wordpress-mobile/AztecEditor-Android/pull/483.

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ImageTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ImageTests.kt
@@ -24,8 +24,28 @@ class ImageTests : BaseTest() {
     @JvmField
     val mActivityIntentsTestRule = IntentsTestRule<MainActivity>(MainActivity::class.java)
 
+    // Simple photo test, also tests the issue described in
+    // https://github.com/wordpress-mobile/AztecEditor-Android/issues/306
     @Test
     fun testAddPhoto() {
+        val regex = Regex("<img src=.+>")
+
+        createImageIntentFilter()
+
+        EditorPage()
+                .tapTop()
+                .insertMedia()
+
+        // Must wait for simulated upload
+        Thread.sleep(10000)
+
+        EditorPage()
+                .toggleHtml()
+                .verifyHTML(regex)
+    }
+
+    @Test
+    fun testAddPhotoAndText() {
         var sampleText = "sample text "
         val regex = Regex(".+<img src=.+>.+")
 
@@ -65,6 +85,25 @@ class ImageTests : BaseTest() {
         EditorPage()
                 .toggleHtml()
                 .verifyHTML(regex)
+    }
+
+    // Tests the issue described in
+    // https://github.com/wordpress-mobile/AztecEditor-Android/issues/299
+    @Test
+    fun testUndoImageUpload() {
+        createImageIntentFilter()
+
+        EditorPage()
+                .tapTop()
+                .insertMedia()
+
+        // Must wait for simulated upload to start
+        Thread.sleep(1000)
+
+        EditorPage()
+                .undoChange()
+                .toggleHtml()
+                .verifyHTML("")
     }
 
     private fun addPhotoWithHTML() {
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
@@ -108,4 +108,84 @@ class MixedTextFormattingTests : BaseTest() {
                 .toggleHtml()
                 .verifyHTML(text)
     }
+
+    @Test
+    fun testTwoHeadings() {
+        val text = "some text"
+        val linebreak = "\n"
+        val html = "<h1>$text</h1>$linebreak<h2>$text</h2>"
+
+        EditorPage()
+                .makeHeader(EditorPage.HeadingStyle.ONE)
+                .insertText(text)
+                .insertText(linebreak)
+                .makeHeader(EditorPage.HeadingStyle.TWO)
+                .insertText(text)
+                .toggleHtml()
+                .verifyHTML(html)
+    }
+
+    @Test
+    fun testEndHeadingFormatting() {
+        val text = "some text"
+        val linebreak = "\n"
+        val html = "<h1>$text</h1>$linebreak$text"
+
+        EditorPage()
+                .makeHeader(EditorPage.HeadingStyle.ONE)
+                .insertText(text)
+                .insertText(linebreak)
+                .insertText(text)
+                .toggleHtml()
+                .verifyHTML(html)
+    }
+
+    @Test
+    fun testEndQuoteFormatting() {
+        val text = "some text"
+        val linebreak = "\n"
+        val html = "<blockquote>$text</blockquote>$linebreak$text"
+
+        EditorPage()
+                .toggleQuote()
+                .insertText(text)
+                .insertText(linebreak)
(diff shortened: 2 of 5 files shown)


app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ImageTests.kt
```diff
@@ -24,8 +24,28 @@ class ImageTests : BaseTest() {
...
+        val regex = Regex("<img src=.+>")
+
+        createImageIntentFilter()
+
+        EditorPage()
+                .tapTop()
+                .insertMedia()
+
+        // Must wait for simulated upload
+        Thread.sleep(10000)
```
brbrr: 10s is quite a large delay. There no way to dynamically check if the upload is done?

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
```diff
@@ -108,4 +108,84 @@ class MixedTextFormattingTests : BaseTest() {
...
+    }
+
+    @Test
+    fun testQuotedListFormatting() {
+        var text = "some text\n"
+        val regex = Regex("<blockquote>\\s+<ul>[\\S\\s]+</ul>\\s+</blockquote>")
+
+        for (i in 1..3) {
+            text += text
+        }
```
brbrr: Is this `for` cycle is needed? I think a simple string with multiple `\n` would make the test easier to read.

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
```diff
@@ -108,4 +108,84 @@ class MixedTextFormattingTests : BaseTest() {
...
+                .insertText(text)
+                .toggleQuote()
+                .toggleHtml()
+                .verifyHTML(html)
+    }
+
+    @Test
+    fun testQuotedListFormatting() {
+        var text = "some text\n"
+        val regex = Regex("<blockquote>\\s+<ul>[\\S\\s]+</ul>\\s+</blockquote>")
```
brbrr: I'm not aware how Regex is working in Kotlin/Java but I've able to make it work after removing escaped `\`: 
`<blockquote>\s+<ul>[\S\s]+<\/ul>\s+<\/blockquote>`

was matched:
```
<blockquote>
  <ul>
    <li>Wee</li>
    <li>Add</li>
    <li>Xox</li>
  </ul>
</blockquote>
```

Tested here: https://regexr.com/

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
```diff
@@ -108,4 +108,84 @@ class MixedTextFormattingTests : BaseTest() {
                 .toggleHtml()
                 .verifyHTML(text)
     }
+
+    @Test
+    fun testTwoHeadings() {
```
brbrr: Are there any GH issues or test cases for these new tests? Want to be sure that Android and iOS tests are following the same test cases

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ImageTests.kt
rachelmcr: I was using the same general approach in the previously added image tests with that sleep, but I can take another look to see what options there might be to dynamically check the upload status.

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
rachelmcr: Good thought. I like the `for` loop for keeping things clean when we need a _really_ long string, but I can probably just use a single string here.

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
rachelmcr: The backslashes need to be escaped per the [pattern syntax](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html). I'm not sure why the escaped version wouldn't work for you (although I'm also looking into stripping whitespace characters between HTML tags in our matchers for these tests, which will allow us to remove some of the regex here).

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ImageTests.kt
rachelmcr: The [simulated upload](https://github.com/wordpress-mobile/AztecEditor-Android/blob/b72568f0a087ad94a09629b57039f54068a87f8d/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L288) in the demo app takes 10s, so I'm going to leave this as-is for our Aztec tests. I'll look at a more dynamic approach for the WordPress app tests.

rachelmcr:
Ok, I think the tests in this PR are complete. @maxme could you review this as well?

I'm going to look into adding some better link tests but I'll do that in a followup PR so I don't delay getting these high priority tests merged.

maxme:
Test looks good, but do you mind implementing `describeMismatchSafely` in the Matchers (or maybe use `TypeSafeDiagnosingMatcher`, I'm not familiar with Harmcrest) to change the output of failing tests? It's hard to read the current output with the full view dump:

```
AssertionFailedWithCauseError: 'EditText matches hello world' doesn't match the selected view.
Expected: EditText matches hello world
Got: "SourceViewEditText{id=2131231019, res-name=source, visibility=VISIBLE, width=768, height=928, has-focus=true, has-focusable=true, has-window-focus=true, is-clickable=true, is-enabled=true, is-focused=true, is-focusable=true, is-layout-requested=false, is-selected=false, layout-params=android.widget.FrameLayout$LayoutParams@f80d919, tag=null, root-is-layout-requested=false, has-input-connection=true, editor-info=[inputType=0xa0001 imeOptions=0x58000005 privateImeOptions=null actionLabel=null actionId=0 initialSelStart=0 initialSelEnd=0 initialCapsMode=0x0 hintText=Codex is the source label=null packageName=null fieldId=0 fieldName=null extras=null hintLocales=null contentMimeTypes=null ], x=0.0, y=0.0, text=bye bye world, hint=Codex is the source, input-type=655361, ime-target=true, has-links=false}"
```

It would be great if we could get this instead:

```
AssertionFailedWithCauseError: 'EditText matches hello world' doesn't match the selected view.
Expected: "hello world"
Got: "bye bye world"
```

I can merge that PR and fill a ticket if you prefer.

rachelmcr:
Good idea. I looked at making that change to the output and it isn't working the way I expect it to, so I'll need to investigate how to implement that correctly. Let's go ahead and merge this PR and I'll follow up on the output message enhancement with this ticket: https://github.com/wordpress-mobile/AztecEditor-Android/issues/531

maxme:
:shipit:

Issue #520, opened 2017-11-09
Add external interface to change caption shortcode


0nko:
The Caption shortcode plugin supports captions in HTML but there is no easy way to modify it from Aztec client's perspective.

There should be an easy way to get and set the caption for a particular image based on its ID attribute.

PR #522, opened 2017-11-10
Issue/520 caption editing

0nko:
Addresses #520 and provides a way to edit the caption from the client side.

### DIFF
--- a/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/extensions/CaptionExtensions.kt
+++ b/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/extensions/CaptionExtensions.kt
@@ -10,17 +10,32 @@ import org.wordpress.aztec.spans.IAztecNestable
 import org.wordpress.aztec.util.SpanWrapper
 
 fun AztecText.getImageCaption(attributePredicate: AztecText.AttributePredicate): String {
-    return this.text
-            .getSpans(0, this.text.length, AztecImageSpan::class.java)
-            .firstOrNull {
-                attributePredicate.matches(it.attributes)
-            }
-            ?.getCaption() ?: ""
+    return this.text.getSpans(0, this.text.length, AztecImageSpan::class.java)
+        .firstOrNull {
+            attributePredicate.matches(it.attributes)
+        }
+        ?.getCaption() ?: ""
+}
+
+fun AztecText.getImageCaptionAttributes(attributePredicate: AztecText.AttributePredicate): AztecAttributes {
+    return this.text.getSpans(0, this.text.length, AztecImageSpan::class.java)
+        .firstOrNull {
+            attributePredicate.matches(it.attributes)
+        }
+        ?.getCaptionAttributes() ?: AztecAttributes()
+}
+
+@JvmOverloads
+fun AztecText.setImageCaption(attributePredicate: AztecText.AttributePredicate, value: String, attrs: AztecAttributes? = null) {
+    this.text.getSpans(0, this.text.length, AztecImageSpan::class.java)
+        .firstOrNull {
+            attributePredicate.matches(it.attributes)
+        }
+        ?.setCaption(value, attrs)
 }
 
 fun AztecText.removeImageCaption(attributePredicate: AztecText.AttributePredicate) {
-    this.text
-        .getSpans(0, this.text.length, AztecImageSpan::class.java)
+    this.text.getSpans(0, this.text.length, AztecImageSpan::class.java)
         .firstOrNull {
             attributePredicate.matches(it.attributes)
         }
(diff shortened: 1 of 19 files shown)


wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/extensions/CaptionExtensions.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.wordpress.aztec.plugins.shortcodes.spans.CaptionShortcodeSpan
+import org.wordpress.aztec.spans.AztecImageSpan
+import org.wordpress.aztec.spans.IAztecNestable
+import org.wordpress.aztec.util.SpanWrapper
+
+fun AztecText.getImageCaption(attributePredicate: AztecText.AttributePredicate): String {
+    return this.text
+            .getSpans(0, this.text.length, AztecImageSpan::class.java)
+            .first {
+                attributePredicate.matches(it.attributes)
```
khaykov: This and same part in `setImageCaption` will throw an exception if nothing is found. Maybe we should just ignore cases where we can't find image span?

0nko: Oh yeah, you're right! :+1:

0nko:
I added the functionality for getting and setting caption attributes and also for removing a caption and checking if a it's present. The functions are in `CaptionExtensions`. I also fixed the problem with text appearing right next to a caption and now it's always on a separate line.

I've spent a lot of time with making sure a newline entered around the caption (pushing the newline below it) is handled properly and all the span bounds are set correctly. Another problem was when a newline got deleted before an image. It would make the caption start _after_ an image. All in all this turned out to be the typical newline 💩show 😑... Hopefully it's all working properly now. Let me know if you need anything else.

khaykov:
Thank's for the improvements, @0nko ! You are right, this is one of those 💩 issues. And we touched it, so it starts to smell :) It works perfectly until we try to type around the image - then weird things start to happen.

Load editor with `IMG` tag that has caption around it.

Type something before the image, you will notice that caption span will move from original caption to the text you typed.
[https://gyazo.com/e5f1f4686acb8ffc65e99241859f6bfa](https://gyazo.com/e5f1f4686acb8ffc65e99241859f6bfa)

Put cursor right after the image (on the same line) and type something, you will notice this:
[https://gyazo.com/b937ce1019a7e8b7b419850e68321ccb](https://gyazo.com/b937ce1019a7e8b7b419850e68321ccb)

Put cursor right after the image (on the same line) and type something, this will result in multiline caption:
[https://gyazo.com/f52d6a1b80a105856a000cbc2287fed5](https://gyazo.com/f52d6a1b80a105856a000cbc2287fed5)

Another thing, functionality wise, it would be great if we could remove a caption with backspace.
When you press backspace on empty caption line, the caption will get deleted instead of an image.

Also, a crash:

[https://gyazo.com/1b08276a4f30fb4f1d24559d52160cc5](https://gyazo.com/1b08276a4f30fb4f1d24559d52160cc5)

```
E: FATAL EXCEPTION: main
   Process: org.wordpress.aztec, PID: 20069
   java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:688)
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:677)
       at org.wordpress.aztec.plugins.shortcodes.watchers.CaptionWatcher.onTextChanged(CaptionWatcher.kt:32)
       at android.widget.TextView.sendOnTextChanged(TextView.java:8182)
       at android.widget.TextView.handleTextChanged(TextView.java:8244)
       at android.widget.TextView$ChangeWatcher.onTextChanged(TextView.java:10366)
       at android.text.SpannableStringBuilder.sendTextChanged(SpannableStringBuilder.java:1208)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:578)
       at android.text.SpannableStringBuilder.delete(SpannableStringBuilder.java:230)
       at android.text.SpannableStringBuilder.delete(SpannableStringBuilder.java:229)
       at android.view.inputmethod.BaseInputConnection.deleteSurroundingText(BaseInputConnection.java:251)
       at android.view.inputmethod.InputConnectionWrapper.deleteSurroundingText(InputConnectionWrapper.java:126)
       at com.android.internal.view.IInputConnectionWrapper.executeMessage(IInputConnectionWrapper.java:451)
       at com.android.internal.view.IInputConnectionWrapper$MyHandler.handleMessage(IInputConnectionWrapper.java:91)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:154)
       at android.app.ActivityThread.main(ActivityThread.java:6077)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:865)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
```
So with all this, I wonder, should we just ship the basic caption functionality with extension and treat proper caption handling as a separate issue?

0nko:
Thanks for catching those @khaykov! I was so fixated on getting the newlines and deletion to work that I forgot the normal text 🤦‍♂️. I fixed the crash and the weird insertion behavior around the image.

About the caption deletion, that could be a little tricky so I'd leave it out of this PR and maybe create a separate issue for it.

khaykov:
It's beautiful ✨ :shipit:

Issue #513, opened 2017-10-31
Cursor can not move to image start.


liufsd:
hi~
when i insert a max width image . the Cursor can not move to this line end.
![image](https://user-images.githubusercontent.com/1910405/32223422-4c43c5f2-be0b-11e7-96a9-c28aa738925f.png)
(can not move the cursor to the position after 'n')

how to fix this issue?

liufsd:
and also can not select this line text before image.
![image](https://user-images.githubusercontent.com/1910405/32223695-61f5348e-be0c-11e7-8c2b-f3daeec20ece.png)

(build by last source code)

0nko:
This is a known issue and is related to #327.

daniloercoli:
Fixed in #525

liufsd:
sorry . not fixed.
test:
input "hello"
and then, insert image
you will see this issue again.

liufsd:
```
    fun insertImage(drawable: Drawable?, attributes: Attributes, onImageTappedListener: OnImageTappedListener?,
                    onMediaDeletedListener: AztecText.OnMediaDeletedListener?) {
        val nestingLevel = IAztecNestable.getNestingLevelAt(editableText, selectionStart)
        val span = AztecImageSpan(editor.context, drawable, nestingLevel, AztecAttributes(attributes), onImageTappedListener,
                onMediaDeletedListener, editor)
        insertMedia(span)
    }
```

The 'nestingLevel' return 0. 
you can also test insert a image in some text. such as:
input 'hello', and then insert a image before 'e'.

liufsd:
Any answer about this ?
i think, insert a image in the some text. we must insert a new line before image tag.

Issue #492, opened 2017-10-11
Add a gap between adjacent images


khaykov:
We should add a 1 dp gap between adjacent images to make it clear which image is targeted by the loader.

maxme:
We should probably add a bigger gap between images (or allow it to be configured by the app) so the user can move it's cursor between images.

folletto:
> We should probably add a bigger gap between images (or allow it to be configured by the app) so the user can move it's cursor between images.

Expanding this from a interaction modelling perspective:

1. The text cursor shows either to the left (vertical bar full height with the image), to the right, or selects the image (the whole image should get something like a border of the same color of the cursor).
2. The cursor should never show in between (so, 1px is fine) unless there's a line in between, in which case, the cursor will show as normally on a blank line
3. In this general model, there's no discernible difference (unfortunately) between two images on the same line that happen to wrap, and two images on two different lines. We could however make this more evident by playing with margins: 1px if they are both on the same line and just happen to wrap, while somewhere between 1px and a 1em (single text line height) when they are on separate lines. While this might follow the same difference between line wrap and paragraph spacing, it might require still a special case to sharpen the difference between same-line, multi-line and line-in-between.

daniloercoli:
Fixed in #525

PR #525, opened 2017-11-14
Fix for various image issues

### LINKED ISSUE #327: A variety of issues from not having images on a separate line
Incorrect click handler is triggered when an unknown HTML element is immediately followed by an image. When you tap on the left side of an image the unknown HTML tap is registered, instead of a tap on the image. 

Another problem is when an image is followed by some text without a line break and it overflows to a line below, then the 2nd line has the same height as the image, which results in a large blank space (depending on the image height).

This bug is related to a weird cursor behavior around images, the [same problem](https://github.com/wordpress-mobile/AztecEditor-Android/issues/293) that was solved for `<hr>` and special comments. Once that is solved for media, this problem will go away, as well.

Subissues:
- [x] #492: Add a gap between adjacent images
- [x] #513: Cursor can not move to image start

0nko:
Fixes #327, #492 and #513.

This PR puts every image on a separate line. This is only a visual change and it does not affect the content.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -346,7 +346,7 @@ class MainActivity : AppCompatActivity(),
             .addPlugin(AudioShortcodePlugin())
 
         BlockElementWatcher(visualEditor)
-                .add(CaptionHandler())
+                .add(CaptionHandler(visualEditor))
                 .install(visualEditor)
 
         // initialize the text & HTML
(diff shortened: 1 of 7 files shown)


daniloercoli:
I'm able to put the cursor on the right side of a picture without any issue now!  🎊 
Unfortunately I noticed that it's easy to breaks the `caption` of a picture. Step to repro:
- Start the demo app 
- Tap on the right side of the first picture and place the cursor at line end
- Tap Enter and write few chars
- Switch to HTML mode and you will see that all of the text is outside the `caption` tag.

0nko:
What would be the expected behavior? In hybrid editor when you follow the steps above and switch to HTML the caption disappears.

daniloercoli:
Good question, I'm not quite sure actually.... I think the cursor should jump _after_ the whole image+caption block, but this is questionable.  Or mimic Calypso web, where the enter inserts a new line above the picture. I see the following behavior on the Web:
caption-and-enter

0nko:
I think I found an acceptable solution. If there is a newline detected before a caption it's pushed underneath.

daniloercoli:
LGTM! :shipit:

PR #526, opened 2017-11-15
Add toolbarStateAboutToChange method to IToolbarAction interface

mzorz:
This PR expands the `IToolbarAction` interface so plugins are signalled that the toolbar is going to be enabled/disabled, for plugins to do their required stuff (i.e. disabling the more button, for example).

The problem is that the [`enableMediaMode()` method in `AztecToolbar`](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt#L760) only goes through all "native" Aztec toolbar buttons, but not through the plugin buttons, thus leaving these buttons enabled all the time.
As plugins might have their own purpose, I decided to extend the functionality by augmenting the `IToolbarAction` interface to let the plugin button know when the toolbar state has changed.

### Fix
Check the related WPAndroid PR https://github.com/wordpress-mobile/WordPress-Android/pull/6871

### Test
1. start a draft post
2. tap on the body area of the editor
3. expand the toolbar and check that the `More` option is tappable and works (inserts a More item in the body of the Post)
4. tap on the `+` at the leftmost part of the toolbar to show the mediapicker.
5. once the mediapicker is showing, check in the toolbar that the `More` option is disabled now.
7. exit the media picker (either by pressing back or by actually choosing to insert some media item)
8. observe the `More` item in the toolbar is enabled now again.

### Review
@0nko

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/plugins/IToolbarButton.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/plugins/IToolbarButton.kt
@@ -3,6 +3,7 @@ package org.wordpress.aztec.plugins
 import android.content.Context
 import android.view.KeyEvent
 import android.view.ViewGroup
+import org.wordpress.aztec.toolbar.AztecToolbar
 import org.wordpress.aztec.toolbar.IToolbarAction
 import org.wordpress.aztec.toolbar.RippleToggleButton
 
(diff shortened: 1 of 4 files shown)


0nko:
Looks good.

PR #534, opened 2017-11-19
Issue/196 make images stay inside link

### LINKED ISSUE #196: Image does not stay inside a link
1. Type the following in HTML mode:
```
<a href="http://wordpress.com">
  <img src="https://s.w.org/about/images/logos/wordpress-logo-stacked-rgb.png" />
<a/>
```
2. Switch to visual mode
3. Switch back to HTML
4. Notice the HTML is now:
```
<img src="https://s.w.org/about/images/logos/wordpress-logo-stacked-rgb.png" />
<a href="http://wordpress.com"><a/>
```

khaykov:
Fixes #196 and adds a couple of methods for simplifying working with links attached to media.

The issue was caused by the span order problem, present in some versions of Android. URL span was processed after image tag and discarded because it was empty.
Fix simply detects situations when this happens and swaps spans so that they will be parsed correctly.

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ImageTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ImageTests.kt
@@ -88,6 +88,26 @@ class ImageTests : BaseTest() {
     }
 
     // Tests the issue described in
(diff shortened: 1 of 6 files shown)


daniloercoli:
Looks good! :shipit:

Issue #481, opened 2017-09-26
Crash report 8.1: RuntimeException in AztecParser.markBlockElementsAsParagraphs


khaykov:
So far this issue only reported on Android 7.0 and 4.1.2 on following **Samsung** devices:

- Galaxy S7 Edge
- Galaxy S7
- Galaxy Win

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{org.wordpress.android/org.wordpress.android.ui.posts.EditPostActivity}: java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2114)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2139)
       at android.app.ActivityThread.access$700(ActivityThread.java:143)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1241)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:137)
       at android.app.ActivityThread.main(ActivityThread.java:4960)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:511)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1038)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:805)
       at dalvik.system.NativeStart.main(NativeStart.java)
Caused by java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:600)
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:588)
       at org.wordpress.aztec.util.SpanWrapper.setFlags(SpanWrapper.kt:20)
       at org.wordpress.aztec.AztecParser.markBlockElementsAsParagraphs(AztecParser.kt:255)
       at org.wordpress.aztec.AztecParser.fromHtml(AztecParser.kt:45)
       at org.wordpress.android.editor.AztecEditorFragment.hasMediaItemsMarkedWithTag(AztecEditorFragment.java:1623)
       at org.wordpress.android.editor.AztecEditorFragment.hasMediaItemsMarkedUploading(AztecEditorFragment.java:1613)
[...]

PR #539, opened 2017-11-22
Fix for 'PARAGRAPH span must start at paragraph boundary'

### LINKED ISSUE #501: Cut&Paste Crash - RuntimeException: PARAGRAPH span must start at paragraph boundary
I had the editor with a long text in it. Different styles were applied to the text.
I selected a word and then pasted other chunk of text previously copied in the clipboard.

**Tested on Nexus 5X Android 7**

```
java.lang.RuntimeException: PARAGRAPH span must start at paragraph boundary
 at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:687)
 at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:676)
 at org.wordpress.aztec.util.SpanWrapper.setFlags(SpanWrapper.kt:20)
 at org.wordpress.aztec.AztecParser.markBlockElementsAsParagraphs(AztecParser.kt:255)
 at org.wordpress.aztec.AztecParser.fromHtml(AztecParser.kt:45)
 at org.wordpress.aztec.AztecText.fromHtml(AztecText.kt:737)
 at org.wordpress.aztec.AztecText.paste(AztecText.kt:1053)
 at org.wordpress.aztec.AztecText.onTextContextMenuItem(AztecText.kt:964)
 at android.widget.Editor$TextActionModeCallback.onActionItemClicked(Editor.java:3793)
 at com.android.internal.policy.DecorView$ActionModeCallback2Wrapper.onActionItemClicked(DecorView.java:2301)
 at com.android.internal.view.FloatingActionMode$3.onMenuItemSelected(FloatingActionMode.java:88)
 at com.android.internal.view.menu.MenuBuilder.dispatchMenuItemSelected(MenuBuilder.java:761)
 at com.android.internal.view.menu.MenuItemImpl.invoke(MenuItemImpl.java:152)
 at com.android.internal.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:904)
 at com.android.internal.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:894)
 at com.android.internal.view.FloatingActionMode$4.onMenuItemClick(FloatingActionMode.java:114)
 at com.android.internal.widget.FloatingToolbar$FloatingToolbarPopup$3.onClick(FloatingToolbar.java:394)
 at android.view.View.performClick(View.java:5637)
 at android.view.View$PerformClick.run(View.java:22429)
 at android.os.Handler.handleCallback(Handler.java:751)
 at android.os.Handler.dispatchMessage(Handler.java:95)
 at android.os.Looper.loop(Looper.java:154)
 at android.app.ActivityThread.main(ActivityThread.java:6121)
 at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
```

Related to https://github.com/wordpress-mobile/AztecEditor-Android/issues/481 ?

daniloercoli:
Fix #501 and #481 by not adding new lines to empty spans.

**To Test**
Paste the following HTML code in the visual editor
```
<p></p><span>[caption]</span>
```
It should not crash the app.

There is a change that the same problem can occur for other tags that are not "caption" or "img", but I wan't able to reproduce on those other tags.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt
@@ -3,6 +3,12 @@ package org.wordpress.aztec.util
 import android.text.Spannable
 
 class SpanWrapper<T>(var spannable: Spannable, var span: T) {
+
+    // Copied from SpannableStringBuilder
+    private val START_MASK = 0xF0
+    private val START_SHIFT = 4
+    private val PARAGRAPH = 3
+
     fun remove() {
         spannable.removeSpan(span)
     }
@@ -17,7 +23,18 @@ class SpanWrapper<T>(var spannable: Spannable, var span: T) {
 
     var flags: Int
         get() { return spannable.getSpanFlags(span) }
-        set(flags) { spannable.setSpan(span, start, end, flags) }
+        set(flags) {
+            // Do not set the span if it's a PARAGRAPH that doesn't start at paragraph boundary
+            // Copied from SpannableStringBuilder that throws an exception in this case.
+            val flagsStart = flags and START_MASK shr START_SHIFT
+            if (!isInvalidParagraph(start, flagsStart)) {
+                spannable.setSpan(span, start, end, flags)
+            }
+        }
+
+    private fun isInvalidParagraph(index: Int, flag: Int): Boolean {
+        return flag == PARAGRAPH && index != 0 && index != spannable.length && spannable.get(index - 1) != '\n'
+    }
 
     companion object {
         inline fun <reified T : Any> getSpans(spannable: Spannable, start: Int, end: Int): List<SpanWrapper<T>> {
--- a/aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt
@@ -6,6 +6,7 @@ class SpanWrapper<T>(var spannable: Spannable, var span: T) {
 
     // Copied from SpannableStringBuilder
     private val START_MASK = 0xF0
+    private val END_MASK = 0x0F
     private val START_SHIFT = 4
     private val PARAGRAPH = 3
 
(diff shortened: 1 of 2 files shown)


0nko:
It seems this does not fix the issue entirely. Pasting the full text from https://fabric.io/automattic/android/apps/org.wordpress.android/issues/59a7e418be077a4dcce162ef?time=last-thirty-days still causes a crash.

Maybe we should add a condition check before applying `SPAN_PARAGRAPH` flag in `AztecParser`.

daniloercoli:
Right, it's probably best to add a condition check berore applying `SPAN_PARAGRAPH `.

For reference this is a test code to use to test this other problem:
`<p></p> <span>  [caption ]</span>Test`

aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt
```diff
@@ -17,7 +23,18 @@ class SpanWrapper<T>(var spannable: Spannable, var span: T) {
 
     var flags: Int
         get() { return spannable.getSpanFlags(span) }
-        set(flags) { spannable.setSpan(span, start, end, flags) }
+        set(flags) {
+            // Do not set the span if it's a PARAGRAPH that doesn't start at paragraph boundary
+            // Copied from SpannableStringBuilder that throws an exception in this case.
+            val flagsStart = flags and START_MASK shr START_SHIFT
+            if (!isInvalidParagraph(start, flagsStart)) {
```
0nko: Since we're checking for the correct start bound of a paragraph, should we also check for the correct end (the last char is a newline or end == text.length)?

0nko:
Would the system Log messages show up anywhere if the app doesn't crash? I know you can log caught exceptions in Crashlytics but we'd be avoiding that here.

aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt
daniloercoli: Right, i've omitted that check on purpose, since we're not seeing crashes for it, but now that I think more on it I agree with you, better to have that check in the code.

daniloercoli:
Nope, system Logs are not shown in wp-android app logs. We should use our Logging library `AppLog`, already available in Utils, if we want to add them to the main app log.

Btw, this is ready for another round.

0nko:
Thanks. It looks good 👍. I'll merge this one and we can add logging in another PR, I think it could be useful.

:shipit:

PR #540, opened 2017-11-23
Feature/min image size support

khaykov:
This PR adds interface support for loading images with specific minimum width. 
This should solve potential problems on clients where tiny images are untappable.

For demonstration purposes only, I added an upscaling functionality to Glide, but I'm not sure if the implementation is done in proper "Glide" way or if it's efficient enough. I tried to look for a good way to do it but wasn't able to find anything working or suited specifically for this problem. If anyone has any ideas - let me know :)

### DIFF
--- a/glide-loader/src/main/java/org/wordpress/aztec/glideloader/GlideVideoThumbnailLoader.kt
+++ b/glide-loader/src/main/java/org/wordpress/aztec/glideloader/GlideVideoThumbnailLoader.kt
@@ -18,6 +18,7 @@ import com.bumptech.glide.request.animation.GlideAnimation
 import com.bumptech.glide.request.target.SizeReadyCallback
 import com.bumptech.glide.request.target.Target
 import org.wordpress.aztec.Html
+import org.wordpress.aztec.glideloader.extensions.upscaleTo
 import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
 import java.io.IOException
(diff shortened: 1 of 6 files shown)


glide-loader/src/main/java/org/wordpress/aztec/glideloader/GlideVideoThumbnailLoader.kt
```diff
@@ -39,8 +45,21 @@ class GlideVideoThumbnailLoader(private val context: Context) : Html.VideoThumbn
                         callbacks.onThumbnailFailed()
                     }
 
-                    override fun onResourceReady(resource: GlideDrawable?, glideAnimation: GlideAnimation<in GlideDrawable>?) {
-                        callbacks.onThumbnailLoaded(resource)
+                    override fun onResourceReady(resource: Bitmap?, glideAnimation: GlideAnimation<in Bitmap>?) {
+                        //Upscaling bitmap only for demonstration purposes.
```
daniloercoli: This PR looks nice @khaykov - I've just a minor comment, maybe nitpicking. I see the same used in the image loader. What do you think to put it in common in a Utils class?

khaykov: Yep, you are right! I'll move it to separate class.

glide-loader/src/main/java/org/wordpress/aztec/glideloader/extensions/BitmapExtension.kt
```diff
@@ -0,0 +1,10 @@
+package org.wordpress.aztec.glideloader.extensions
+
+import android.graphics.Bitmap
+import android.graphics.Bitmap.createScaledBitmap
+
+internal fun Bitmap.upscaleTo(desiredWidth: Int): Bitmap {
+    val ratio = this.height / this.width
```
daniloercoli: I've just found problems when resizing pictures whose h/w ratio is less than 1.
We should make sure to use Float variables when calculating the `ratio`, and then cast to int at the end of the routine. Something like the following.

``` 
val ratio : Float = (this.height * 1f) / this.width
val proportionateHeight : Float = ratio * desiredWidth
val finalHeight = Math.rint(proportionateHeight.toDouble()).toInt()

return createScaledBitmap(this, desiredWidth, finalHeight, true)
```

We could also add an additional check on `finalHeight`, ensuring that it's `> 0`

khaykov: Good catch. Original code from wpandroid used floats, but I brainfarted and removed the casts for some reason :) 

Rounding is a nice touch too!

daniloercoli:
:shipit:

PR #548, opened 2017-11-28
UI tests fix

0nko:
This PR enables the Calypso mode for certain formatting tests that are failing since Calypso mode is disabled by default in the demo app.

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/Matchers.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/Matchers.kt
@@ -5,6 +5,7 @@ import android.widget.EditText
 import org.hamcrest.Description
 import org.hamcrest.Matcher
 import org.hamcrest.TypeSafeMatcher
+import org.wordpress.aztec.source.Format
 
 object Matchers {
     fun withRegex(expected: Regex): Matcher<View> {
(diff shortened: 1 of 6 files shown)


app/src/androidTest/kotlin/org/wordpress/aztec/demo/Matchers.kt
```diff
@@ -35,9 +36,9 @@ object Matchers {
 
             public override fun matchesSafely(view: View): Boolean {
                 if (view is EditText) {
-                    val regex = Regex(">\\s+<")
-                    val strippedText = view.text.toString().replace(regex, "><")
-                    return strippedText.equals(expected)
+                    val expectedHtml = Format.removeSourceEditorFormatting(expected, false)
```
rachelmcr: This is working, but I'm a little wary of stripping the formatting like this from both the expected and actual values, in case a bug is introduced into this function in the future. How about updating the expected values (to remove any whitespace) and comparing the unchanged expected values to the stripped actual HTML? (I can do this in another PR if you prefer/agree.)

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/MixedTextFormattingTests.kt
```diff
@@ -1,18 +1,31 @@
...
     @JvmField
     var mActivityTestRule = ActivityTestRule(MainActivity::class.java)
 
+    @Before
+    fun init() {
+        val aztecText = mActivityTestRule.activity.findViewById<AztecText>(R.id.aztec)
+        val sourceText = mActivityTestRule.activity.findViewById<SourceViewEditText>(R.id.source)
+
+        aztecText.setCalypsoMode(true)
+        sourceText.setCalypsoMode(true)
```
rachelmcr: I'm fine with this, but would it be simpler to enable Calypso mode in `MainActivity` when `isRunningTest` is true (similar to how we decide whether to add demo content) instead of here before each test?

0nko: Yeah, that would work, but not for all tests. Paragraph tests, for example, will fail unless Calypso mode is off. But since the majority of tests rely on it being enabled I'll just disable it for paragraphs.

app/src/androidTest/kotlin/org/wordpress/aztec/demo/Matchers.kt
0nko: You're right. Let's do what you suggested in a separate PR.

0nko:
Ready for another pass.

PR #549, opened 2017-11-29
Always close <video> and <audio> tags immediately 

### LINKED ISSUE #472: Shortcode parsing errors
1. Configure the demo app to use the shortcode plugin
2. Start the app and add an `<audio>` (or `<video>`) element, follow by an image in HTML: `<audio>`
2. Switch to visual mode
3. Switch back to HTML
4. Notice there are 2 audio (or video) shortcodes, instead of 1

0nko:
Fixes #472. If the parser finds an opening tag for `<audio>` or `<video>` it moves on until it finds the closing tag or the end of a document, if there isn't one. In that case the span is too long and produces multiple elements during parsing back to HTML. This PR makes sure the tags above are closed as soon as they're encountered.


daniloercoli:
Looks cool! :shipit: 

Just for reference, WordPress handles Video tags in custom way. There is not way to add this standard HTML to a post content
```
<video width="320" height="240" controls>
  <source src="movie.mp4" type="video/mp4">
  <source src="movie.ogg" type="video/ogg">
  Your browser does not support the video tag.
</video>
```

PR #550, opened 2017-11-29
Add logging utils library to Aztec

daniloercoli:
This PR just adds our Logging utils library to Aztec, and use it (`AppLog`) to start logging the PARAGRAPH issue details. 
With this PR in the repo we can send logs into out main app log.

### Review
@0nko

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -43,7 +43,6 @@ dependencies {
     implementation "com.android.support:appcompat-v7:$supportLibVersion"
     implementation "org.wordpress:utils:$wordpressUtilsVersion"
 
-
     androidTestImplementation "com.android.support.test.espresso:espresso-core:$espressoVersion", {
         exclude group: 'com.android.support', module: 'support-annotations'
     }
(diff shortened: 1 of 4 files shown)


0nko:
Looks good 👍. I just realized that maybe we should also check and log invalid paragraphs [here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/64c7d6dae7386322f7ff8960c18331f31190b49a/aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt#L18) and [here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/64c7d6dae7386322f7ff8960c18331f31190b49a/aztec/src/main/kotlin/org/wordpress/aztec/util/SpanWrapper.kt#L22), since those places could potentially also become the source of errors.

daniloercoli:
Right! I've updated this PR by checking the boundaries in those other methods too.

0nko:
Thanks, everything looks good, there are just some ktlint errors.

PR #551, opened 2017-11-30
Disable the Calypso mode before loading the sample text

0nko:
Oops, seems like I broke the demo app with my last PR.


maxme:
:shipit:

Issue #445, opened 2017-08-03
History not reliable


0nko:
The history does not work reliably (undo, redo does not bring back the right state) for:

- Copy/pasting 
- Adding newlines inside blockquotes
- Deleting of empty list items with a backspace
- Adding an image, pressing undo, then redo (does not bring the image back)
- Deleting the demo text, then undoing loses all styling

We need unit tests for all the different scenarios of using the history.

A nice feature would be to be able to define a timeout and only save after X seconds of inactivity, as opposed to after every key stroke.

maxme:
I de-prioritized this issue for now. We can eventually ship Aztec v1 without undo/redo since the Hybrid editor doesn't have it.

PR #552, opened 2017-11-30
Fixed History Reliability Issues and Added Tests

AmandaRiu:
### Fix
This PR fixes the issues outlined in [#445](https://github.com/wordpress-mobile/AztecEditor-Android/issues/445) and are described below. 

#### Recording History
Previously, the history component would record every textChangedEvent, even the ones not initiated by the user but by different TextWatchers. This caused the historical stack to include changes not visible until a user clicks UNDO. To fix this issue, I have added a timer that will only record text changes when there haven't been any new TextChangedEvents for 500 ms. 

Added a new variable to `AztecText` called `consumeHistoryEvent`, which is used to prevent history from recording events like the backspace generated in the `setupZeroIndexBackspaceDetection` method. 

#### Copy/Pasting
Most of the issues with copy/pasting were due to editor changes by other TextWatchers that were being recorded in the history stack. The rest:
- Copy/Pasting Images: Added a new method to History called `refreshLastHistoryItem`. This method replaces the last history item with fresh data. This is specifically important for Images because the image html has the `uploading` attribute until the upload is complete. History was just storing the initial html so if you UNDO/REDO, the editor will be left in a bad state with the user unable to switch to source view. 

#### Block Elements
The `ParagraphCollapseRemover` was breaking REDO history by removing spans of type _IParagraphFlagged_ during the _beforeTextChanged_ event. For example, I would add the following html:
```kotlin
<h1>Heading 1</h1>
<h2>Heading 2</h2>
<h3>Heading 3</h3>
```
Switch to editor mode, select all, and then hit delete. Click UNDO and the html was be restored as:
```
<h1>Heading 1</h1>
Heading 2
Heading 3
```
The H2 and H3 heading spans were being removed by the `ParagraphCollapseRemover` _before_ history was recorded. `ParagraphCollapseRemover` must do all of it's changes in the beforeTextChanged event because that's where the positions are still in tact. So to fix this, I moved the registering of the `AztecText` and `ParagraphCollapseRemover` text watchers. Now they are the very last textWatchers to register and AztecText **_must_** register **_before_** `ParagraphCollapseRemover`:
```
// History related logging has to happen before the changes in [ParagraphCollapseRemover]
addTextChangedListener(this)
ParagraphCollapseRemover.install(this)
```
### Tests
A bunch of tests were added, at least one test for each type of element. All the tests are Instrumented Tests. I added some `Actions` and `Matchers` and several new methods to `EditorPage` to support my tests. One of those new methods is `threadSleep`. This is used during History testing to wait for history to be recorded now that there is a timer in use. I added this method here for convenience so when I'm writing tests, instead of writing:
```
editorPage
	.insertText(word1)
Thread.sleep(throttleTime())
editorPage
	.insertText(word2)	
Thread.sleep(throttleTime())
editorPage
	.insertText(word3)
Thread.sleep(throttleTime())	
editorPage
	.insertText(word4)
Thread.sleep(throttleTime())	
```
I can write it like this:
```
editorPage
    .insertText(word1)
    .threadSleep(throttleTime())
    .insertText(word2)
    .threadSleep(throttleTime())
    .insertText(word3)
    .threadSleep(throttleTime())
    .insertText(word4)
    .threadSleep(throttleTime())
```
### Misc
- Cleaned up some lint errors for classes I had modified. 
- Optimized styles.xml by replacing "" with `@null` for textOn and textOff. While stepping through the code I ran into a bunch of work that was happening to process the empty text. 
- Cleaned up ktlint formatting errors.
### Review
@0nko Please let me know if there are any scenarios that can be added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ annotations/
 # Enforce plugins
 !/.idea/externalDependencies.xml
 
+# OS generated
+/.DS_Store
+
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,5 @@ annotations/
 # OS generated
 /.DS_Store
 
+# Kotlin lint
+ktlint
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/Actions.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/Actions.kt
@@ -14,6 +14,8 @@ import android.view.View
 import android.widget.EditText
 import org.hamcrest.Matcher
 import org.hamcrest.Matchers.allOf
+import org.wordpress.aztec.AztecText
+import org.wordpress.aztec.watchers.EndOfBufferMarkerAdder
 
 object Actions {
     fun invokeClick(): ViewAction {
@@ -32,7 +34,6 @@ object Actions {
         }
     }
 
-    @Suppress("DEPRECATION")
     fun relativeClick(xPercent: Float = 0.5f, yPercent: Float = 0.5f): ViewAction {
         return GeneralClickAction(
                 Tap.SINGLE,
@@ -45,7 +46,7 @@ object Actions {
 
                     floatArrayOf(x, y)
                 },
-                Press.FINGER)
+                Press.FINGER, 0, 0)
     }
 
     fun selectAll(): ViewAction {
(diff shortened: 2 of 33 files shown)


maxme:
🍾

AmandaRiu:
To summarize the final fixes for this PR:

- **History with Images:** This is due to the "uploading" flag in history. If it exists, then the toolbar will not allow you to switch to HTML view. Per @0nko he spoke with Mario Zorz and confirmed that the media upload is canceled when an image is deleted and there is a hook in the WordPress app that detects the redo action and restart the image upload if necessary. _So no changes needed here._ 

- **Heading History:** Broken in the main branch and not due to code changes in this PR. _No changes needed._

- **Copy/Pasting of styled text:** Added a new test to try and replicate the issue. Was never able to. Per @0nko this only seems to happen on emulators. _No changes needed._

PR #554, opened 2017-12-04
Fixed URL span attributes

khaykov:
URL span didn't include a `href` attribute when additional attributes, like `target`, was passed to the constructor.


0nko:
:shipit:

PR #557, opened 2017-12-05
Check the beginning of changed text for images

0nko:
Resolves the problem when adding text behind an image doesn't put the text on a new line.


maxme:
:shipit: confirmed this fix the issue in the wpandroid app.

Issue #434, opened 2017-07-25
HTML inconsistency with <div>


0nko:
Parsing the following HTML gives an incorrect output HTML:

```
<h3><strong>Shipped:</strong></h3><br>
<div class="entry-content">
	<br>
	<ul>
		<li style="list-style: none"><br></li>
		<li>
			<strong>Something:</strong> Text Here (<a href="https://wordpress.com">#</a>).
		</li>
		<li style="list-style: none"><br></li>
		<li>
			<strong>Title:</strong> List Item <a href="https://wordpress.com">link</a>.
		</li>
		<li style="list-style: none"><br></li>
		<li>
			<strong>Some bold text:</strong> plain here, <a href="https://wordpress.com">and a link!</a>.
		</li>
		<li style="list-style: none"><br></li>
	</ul><br>
	<h3><strong>Another list:</strong></h3><br>
	<ul>
		<li style="list-style: none"><br>
		<br></li>
		<li>
			<strong>Another Item</strong>Another <a href="https://wordpress.com">link</a>
		</li>
		<li style="list-style: none"><br></li>
	</ul><br>
</div><br>
#yolo
```

### Expected

The same as input.

### Observed

```
<h3><b>Shipped:
<div class="entry-content"></div>
</b></h3>
<ul>
	<li style="list-style: none"></li>
	<li>			<b>Something:</b> Text Here (<a href="https://wordpress.com">#</a>).</li>
	<li style="list-style: none"></li>
	<li>			<b>Title:</b> List Item <a href="https://wordpress.com">link</a>.</li>
	<li style="list-style: none"></li>
	<li>			<b>Some bold text:</b> plain here, <a href="https://wordpress.com">and a link!</a>.</li>
	<li style="list-style: none"></li>
</ul>
<h3><b>Another list:</b></h3>
<ul>
	<li style="list-style: none"></li>
	<li>			<b>Another Item</b>Another <a href="https://wordpress.com">link</a></li>
	<li style="list-style: none"></li>
</ul>
#yolo
```

### Reproduced
1. Paste the original code in the HTML mode
2. Switch to visual mode
3. Switch back to HTML
4. Notice the output is not the same as input

#### Tested
Emulator on Android 7.1.1 with 1.0-beta.6

maxme:
We should have a unit test to reproduce that one - cc @rachelmcr

rachelmcr:
[...]

PR #568, opened 2017-12-06
Add unit tests for div HTML parsing

rachelmcr:
This PR adds unit tests for the two `div` parsing issues described in https://github.com/wordpress-mobile/AztecEditor-Android/issues/434#issuecomment-349706228.

1. `div` tags are removed (as above) when a `br` tag precedes it.
Input: `<br><div><br></div>`
Output: `<br><br>`

2. `div` tags are moved inside a list item when there is a list inside the div.
Input: `<div><ul><li>Unordered</li></ul></div>`
Output: `<ul><li><div>Unordered</div></li></ul>`

### Review
@maxme

### DIFF
--- a/aztec/src/test/kotlin/org/wordpress/aztec/AztecParserTest.kt
+++ b/aztec/src/test/kotlin/org/wordpress/aztec/AztecParserTest.kt
@@ -8,6 +8,7 @@ import android.text.SpannableString
 import android.text.SpannableStringBuilder
 import org.junit.Assert
 import org.junit.Before
+import org.junit.Ignore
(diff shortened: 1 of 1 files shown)


maxme:
Tests looks good, but I'm not sure what to do with them since they'll break our travis checks. I wonder if we could put them in a special bucket that is not executed by travis. When we fix #434 we can move them back.

rachelmcr:
I added `@Ignore` to the new tests so they'll be ignored for now; we can remove that when we want to start running them.

maxme:
👍 and TIL! :shipit:

PR #569, opened 2017-12-06
Exposes some static methods to java

mzorz:
This PR adds `@JvmStatic` to several methods that we need to use from outside to cope with Calypso mode formatting.

### Review
@0nko


0nko:
:shipit:

PR #571, opened 2017-12-11
Log the text that caused RuntimeException in AztecText.toPlainHtml

daniloercoli:
This PR just adds the text that caused the issue to AppLog, and keep letting the app crash.
We need to remove the log line once we've enough data to replicate it. I've added a FIXME there to remember us to remove it.


mzorz:
:shipit:

PR #575, opened 2017-12-18
Add VideoPress support

### LINKED ISSUE #469: Add support for [wpvideo] shortcode
`[wpvideo]` is another video shortcode that must be implemented. It does not contain a full path to the video as `[video]` does, only a reference to videopress.com file.

More information: https://en.support.wordpress.com/videopress/

daniloercoli:
This PR adds support for VideoPress short codes. Fixes #469

Since VideoPress videos don't have high usage in the app (it's actually a rarities) i've decided to not introduce a new shortcode class, instead playing around the already available VideoShortcodePlugin code. 
This PR just adds 2 hidden attributes (removed when switching to HTML or posting) and use those attributes to ask for video's thumb, update the video placeholder when the App returns back with the actual poster image, and use the original src when a video is tapped.

### Test
1. Open the demo app
2. See videopress videos at the bottom
They should get updated once the app is started.
3. switch to HTML mode and check that there aren't unexpected attributes added to them 

I've already a first working demo branch of wp-android. Need to fix some issue with uploading a new video.

### Review
@0nko

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -839,7 +839,10 @@ open class MainActivity : AppCompatActivity(),
         See: https://developer.wordpress.com/docs/api/1.1/get/videos/%24guid/
         The response has all info in it. We're skipping it here, and set the poster image directly
         */
-        aztec.visualEditor.updateVideoPressThumb("https://videos.files.wordpress.com/OcobLTqC/img_5786_hd.original.jpg", videoID)
+        aztec.visualEditor.updateVideoPressThumb(
+                "https://videos.files.wordpress.com/OcobLTqC/img_5786_hd.original.jpg",
+                "https://videos.files.wordpress.com/OcobLTqC/img_5786.m4v",
+                videoID)
     }
 
     override fun onAudioTapped(attrs: AztecAttributes) {
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -63,6 +63,7 @@ open class MainActivity : AppCompatActivity(),
         AztecText.OnVideoTappedListener,
         AztecText.OnAudioTappedListener,
         AztecText.OnMediaDeletedListener,
+        AztecText.OnVideoPressInfoRequestedListener,
         IAztecToolbarClickListener,
         IHistoryListener,
         OnRequestPermissionsResultCallback,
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -819,6 +831,20 @@ open class MainActivity : AppCompatActivity(),
...
 
+    override fun onVideoPressInfoRequested(videoID: String) {
+        AppLog.d(AppLog.T.EDITOR, "VideoPress Info Requested for video ID " + videoID)
+
+        /*
+        Here should go the Network request that retrieves additional info about the video.
+        See: https://developer.wordpress.com/docs/api/1.1/get/videos/%24guid/
+        The response has all info in it. We're skipping it here, and set the poster image directly
+        */
+        aztec.visualEditor.updateVideoPressThumb(
```
0nko: I noticed the thumbnail is not displayed in the demo app.

aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
```diff
@@ -21,6 +21,7 @@ open class Aztec private constructor(val visualEditor: AztecText, val toolbar: A
     private var onVideoTappedListener: AztecText.OnVideoTappedListener? = null
     private var onAudioTappedListener: AztecText.OnAudioTappedListener? = null
     private var onMediaDeletedListener: AztecText.OnMediaDeletedListener? = null
+    private var onVideoPressInfoRequestedListener: AztecText.OnVideoPressInfoRequestedListener? = null
```
0nko: Since we'd like to keep Aztec generic and VideoPress is WP-specific, we could move this to the shortcode plugin (using extension functions or inheritance).

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -134,6 +134,7 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
     private var onVideoTappedListener: OnVideoTappedListener? = null
     private var onAudioTappedListener: OnAudioTappedListener? = null
     private var onMediaDeletedListener: OnMediaDeletedListener? = null
+    private var onVideoPressInfoRequestedListener: OnVideoPressInfoRequestedListener? = null
```
0nko: VidePress stuff in `AztecText` could also be moved to a plugin.

aztec/src/main/kotlin/org/wordpress/aztec/Constants.kt
```diff
@@ -13,4 +13,6 @@ object Constants {
     val NEWLINE_STRING = "" + NEWLINE
     val END_OF_BUFFER_MARKER = ZWJ_CHAR
     val END_OF_BUFFER_MARKER_STRING = "" + ZWJ_CHAR
+    val ATTRIBUTE_VIDEOPRESS_HIDDEN_ID = "videopress_hidden_id"
+    val ATTRIBUTE_VIDEOPRESS_HIDDEN_SRC = "videopress_hidden_src"
```
0nko: Also WP-specific.

daniloercoli:
Thanks @0nko for the review. I've made the video listener generic, not tied to VidePress only;  that way it could be reused later. Also made an extensions that adds VP feature to AztecText.

This is ready for another pass.

0nko:
Perfect! :shipit:

PR #576, opened 2017-12-19
Issue/572 caption newline

### LINKED ISSUE #572: <br> in front of caption text pushes it outside of caption tag
Following HTML will render caption text outside of caption tag.

`[caption align="alignright"]<br>Caption[/caption]`

We don't support multiline captions yet, so ideally this tag should be stripped.

0nko:
Fixes #572.

Removes the linebreaks at the beginning and end of a caption and replaces them with a spec in the middle.


mzorz:
tested, LGTM :shipit:

PR #581, opened 2017-12-21
Issue/559 comment cursor

### LINKED ISSUE #559: Cursor not positioned to the end after adding a special comment
### Expected
When writing, if the user doesn't explicitely move the cursor around, cursor should be kept at the end and, if the user continues typing, text should always be appended.

### Observed
If you try to insert a separator or more special comment on the same line some characters have already been entered, the cursor ends being at the left of the line instead of being places on the right (end) of text.

### Reproduced
1. start a draft post
2. write some text and, on then same line, insert a line separator / `more` separator
3. continue writing: you’ll see the line is pushed below, instead of the cursor being positioned _after_  the inserted line to continue appending text.

#### Tested
[Google Pixel] on [8.0] with [alpha-97]

0nko:
Fixes #559. 

Fixes the cursor position after insertion for special comments and the horizontal rule.


daniloercoli:
Should not be the cursor positioned on a new line after inserting the more or page item?
https://cloudup.com/c-PWIDZ5Foa

I see the cursor blinking at line end, and the moving to the new line as soon as I start typing again.

0nko:
That's expected. The HR line is surrounded with a newline (just like images) when you try to add text on the same line.  Since you added the HR at the end of the editor a newline is not initially added.

daniloercoli:
Looks good! :shipit:

Issue #452, opened 2017-08-07
Crash report: RuntimeException in AztecText.toPlainHtml


tonyr59h:
There aren't a _ton_ of users effected so I've marked it medium priority.

```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=2621; index=2621
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:912)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:917)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:917)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:917)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:917)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:878)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:878)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:878)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:917)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:878)
       at android.text.SpannableStringBuilder.getSpansRec(SpannableStringBuilder.java:917)
       at android.text.SpannableStringBuilder.getSpans(SpannableStringBuilder.java:826)
       at android.text.SpannableStringBuilder.<init>(SpannableStringBuilder.java:76)
       at android.text.SpannableStringBuilder.<init>(SpannableStringBuilder.java:49)
       at org.wordpress.aztec.AztecText.toPlainHtml(AztecText.kt:800)
       at org.wordpress.aztec.AztecText.toHtml(AztecText.kt:786)
       at org.wordpress.android.editor.AztecEditorFragment.getContent(AztecEditorFragment.java:570)
       at org.wordpress.android.ui.posts.EditPostActivity.updatePostObject(EditPostActivity.java:900)
       at org.wordpress.android.ui.posts.EditPostActivity.access$1000(EditPostActivity.java:149)
       at org.wordpress.android.ui.posts.EditPostActivity$5.run(EditPostActivity.java:916)
       at java.lang.Thread.run(Thread.java:818)
```

khaykov:
[...]

PR #583, opened 2017-12-22
Add the text that caused RuntimeException in AztecText.toPlainHtml to crash reports

daniloercoli:
This PR just adds the text that caused the issue to crash reports, and keep letting the app crash.
We need to remove the log line once we've enough data to replicate it. I've added a FIXME there to remember us to remove it.

This is an addition to the previous PR here #571 where we had added the text to the log of the app only, but not the exceptions stack, so it was not reported in Crashlytics.
Original report of the issue #452


daniloercoli:
Stupid me. This is not necessary, since we're already logging it fine in wp-android.

PR #590, opened 2018-01-03
Issue/574 caption removal

0nko:
Removes the caption text as per discussion in wordpress-mobile/WordPress-Android#7044.


khaykov:
I found one issue when a caption is the last element in the editor (no text below it).
When you delete the image in this situation the last character of the caption remains, like this:

[https://gyazo.com/7d71f5512b009551eb605a02000bb2c0](https://gyazo.com/7d71f5512b009551eb605a02000bb2c0)

0nko:
Thanks for the test! That should fix it.

khaykov:
:shipit:

PR #591, opened 2018-01-04
Unknown element tap fix

0nko:
Fixes a problem when the unknown element editor dialog wouldn't show up after previously opening and saving the unknown element.

To reproduce:
1. Tap an unknown element
2. Tap Save on the dialog
3. Try tapping on the unknown element again
4. A dialog should pop up again


daniloercoli:
:shipit:

PR #593, opened 2018-01-05
Fix Image/Video Placeholders dimensions

daniloercoli:
This PR makes sure that image/video loading and error placeholders are drawn on the screen by using the `max-pictures-width` dimension set in Aztec. In this way placeholders are going to match the real dimension of the pictures much closer than before. This is a fix for https://github.com/wordpress-mobile/WordPress-Android/issues/6942, especially for failed image requests where the placeholder icon was so small.

cc @mzorz 

Note: I think a real fix for this is to patch `AztecDynamicImageSpan` and friends, and make sure that those will respect the `setBounds` values set with a simple `loadingDrawable.setBounds(0,0, maxImageWidthForVisualEditor, maxImageWidthForVisualEditor)`. That at the moment doesn't work fine. This patch works fine since it does convert it to a bitmap with the correct size, that are then drawn on the screen.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -21,11 +21,17 @@ import android.annotation.SuppressLint
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.Canvas
+import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
+import android.graphics.drawable.VectorDrawable
 import android.os.Build
 import android.os.Bundle
 import android.os.Parcel
 import android.os.Parcelable
+import android.support.annotation.DrawableRes
+import android.support.graphics.drawable.VectorDrawableCompat
 import android.support.v4.content.ContextCompat
 import android.support.v7.app.AlertDialog
 import android.support.v7.widget.AppCompatEditText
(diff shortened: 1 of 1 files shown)


mzorz:
I tested the feature by referencing this branch in the editor gradle file in branch `issue/6945-thumbnails-same-sizes` of WPAndroid, then 
1. starting a draft and adding pictures/videos, 
2. then forcing error (by switching airplane mode on) and seeing the placeholders. 
3. The pictures/videos are shown with the thumbnails and the error overlays are shown on them, 
4. but then if I rotate at that point, the original thumbnails are lost and only the placeholders are shown, and they do not respect the original image size. 

Screenshots here:

Right after point 4:

image


After adding another video and rotating:
image

Also, rotating the screen when you have a video makes the placeholder occupy a lot of space (exceeds the screen height, as we calculate things based on width)

image

Tested on Google Pixel, Android 8.0

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -868,17 +896,17 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
...
         spans.forEach {
             val callbacks = object : Html.VideoThumbnailGetter.Callbacks {
                 override fun onThumbnailFailed() {
-                    replaceImage(ContextCompat.getDrawable(context, drawableFailed))
+                    AztecText.getPlaceholderDrawableFromResID(context, drawableFailed, maxImagesWidth)
                 }
 
                 override fun onThumbnailLoaded(drawable: Drawable?) {
-                    replaceImage(drawable)
+                    //replaceImage(drawable)
```
mzorz: should we uncomment this line here?

daniloercoli: Opps! 🤕

mzorz:
Tested on the other branch (https://github.com/wordpress-mobile/WordPress-Android/pull/7057), LGTM :shipit:

Issue #541, opened 2017-11-23
Hide comments in visual mode


0nko:
Add a configuration option to hide the HTML comments in visual mode (in preparation for Gutenberg).

maxme:
There might be another change to consider to support basic Gutenberg blocks: images are wrapped in a `figure` tag like:

```
<!-- wp:core/image {"id":316} -->
<figure class="wp-block-image"><img src="https://bia.is/hola-que-tal.jpg" alt="" /></figure>
<!-- /wp:core/image -->
```

Aztec could simply ignore visually these tags as a first step.

daniloercoli:
Fixed in #592

PR #594, opened 2018-01-05
Hide Gutenberg blocks

0nko:
This PR implements the special Gutenberg code block HTML from the visual editor, as suggested in #541.

The regular comment-hiding was unsuitable for this task so now there is a new type of plugin interface - [`IHtmlContentHandler`](https://github.com/wordpress-mobile/AztecEditor-Android/blob/13e180462267b79b13a4ce176e1f404930ad859b/aztec/src/main/kotlin/org/wordpress/aztec/plugins/html2visual/IHtmlContentHandler.kt). A plugin that implements it is supplied with the content of a declared element and the element is skipped by the parser.


daniloercoli:
I've tested this PR with a small GB post, and was wondering if we should show the Unknown block item `?`, instead of hiding GB blocks altogether.
If you edit a GB post it looks empty in the visual mode, or, if you edit a post that is a mix of GB blocks and "conventional" text, it looks half empty and the result is a weird experience. 

Sorry if this has already been discussed somewhere and I've forgot about it, or missed the convo.

0nko:
Good question. @diegoreymendez, do have this in Aztec iOS?

diegoreymendez:
We're replacing the Gutenberg comment blocks with zero-width spaces.  At the time we implemented this it was agreed to be a simple compromise solution.  Especially since Aztec is an HTML-standard editor so any efforts to make it "handle" Gutenberg code would be misplaced and counterproductive IMO.

Anything more complicated has the potential of becoming a big effort we'd rather avoid in Aztec (since Gutenberg mobile is where we want to be putting that effort).

@maxme is the closest we have to a product owner though, so maybe he has other thoughts on this.

diegoreymendez:
Just to be clear... we're replacing the opening and closing comment blocks, but the content is parsed as regular HTML.

0nko:
I've changed the implementation so that only the GB comments are hidden, the content is rendered normally. I've added `figure`, `figcaption` and `section` to a list of hidden elements because GB blocks place HTML inside them.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -109,6 +110,9 @@ open class MainActivity : AppCompatActivity(),
                         "    <span class=\"second last\"></span>" +
                         "</div>" +
                         "<br>"
+        private val GUTENBERG_CODE_BLOCK = "<!-- wp:core/image {\"id\":316} -->\n" +
+                "<figure class=\"wp-block-image\"><img src=\"https://bia.is/hola-que-tal.jpg\" alt=\"\" /></figure>\n" +
```
daniloercoli: May worth to point to a picture that doesn't return 404?

daniloercoli:
@mzorz I'd like to get another review on this PR. Would you mind to take a quick look?

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
0nko: Oops. Changed it.

mzorz:
> Would you mind to take a quick look?

sure thing!

PR #596, opened 2018-01-08
Video - Make sure to show the correct bitmap once loaded from the network

daniloercoli:
This was an issue introduced in #593 where I left commented the line of code that actually replaces the placeholder with the correct bitmap.

cc @mzorz


mzorz:
LGTM :shipit:

PR #597, opened 2018-01-08
Fix Video/Image error placeholder size when Aztec is used in apps

daniloercoli:
This PR fixes an issue where the error placeholder was always created at default 48DP dimensions, even though a different dimension was set fine by 3rd part app.

I bet it's something going wrong the way the value was referenced inside the `forEach` loop. It was reading the value of `maxImagesWidth` set during `init`. 

## To Test
- Update `wp-android` and reference this hash
```
    compile ('com.github.wordpress-mobile.WordPress-Aztec-Android:aztec:d9f1926f63d48c83da2405725410977a97064fb7')
    compile ('com.github.wordpress-mobile.WordPress-Aztec-Android:wordpress-shortcodes:d9f1926f63d48c83da2405725410977a97064fb7')
    compile ('com.github.wordpress-mobile.WordPress-Aztec-Android:wordpress-comments:d9f1926f63d48c83da2405725410977a97064fb7')
```
- On the web, create a new post with a 404 picture on it. EX `src="https://eritreocazzulati.files.wordpress.com/2018/01/img_20180107_141620.std.jpg`
- Start wp-android and edit the post
- The error placeholder should appear in the correct dimensions and not too small 

### Review
@mzorz

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -864,15 +864,15 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
     }
 
     private fun loadImages() {
-        val maxDimension = maxImagesWidth
         val spans = this.text.getSpans(0, text.length, AztecImageSpan::class.java)
-        val loadingDrawable = AztecText.getPlaceholderDrawableFromResID(context, drawableLoading, maxDimension)
+        val loadingDrawable = AztecText.getPlaceholderDrawableFromResID(context, drawableLoading, maxImagesWidth)
 
+        val maxDimension = maxImagesWidth
         spans.forEach {
             val callbacks = object : Html.ImageGetter.Callbacks {
                 override fun onImageFailed() {
                     replaceImage(
-                            AztecText.getPlaceholderDrawableFromResID(context, drawableFailed, maxDimension)
+                            AztecText.getPlaceholderDrawableFromResID(context, drawableFailed, maxImagesWidth)
                     )
                 }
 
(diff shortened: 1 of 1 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -864,14 +864,15 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
     }
 
     private fun loadImages() {
+        val maxDimension = maxImagesWidth
```
mzorz: I think we can remove `val maxDimension` and use `maxImagesWidth` directly in this method

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -890,19 +891,20 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
                     }
                 }
             }
-            imageGetter?.loadImage(it.getSource(), callbacks, this@AztecText.maxImagesWidth, this@AztecText.minImagesWidth)
+            imageGetter?.loadImage(it.getSource(), callbacks, maxDimension, maxDimension)
```
mzorz: replacing this call here with `imageGetter?.loadImage(it.getSource(), callbacks, maxImagesWidth, maxImagesWidth)` should work alright

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -920,7 +922,7 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
                     }
                 }
             }
-            videoThumbnailGetter?.loadVideoThumbnail(it.getSource(), callbacks, this@AztecText.maxImagesWidth, this@AztecText.minImagesWidth)
+            videoThumbnailGetter?.loadVideoThumbnail(it.getSource(), callbacks, maxDimension, maxDimension)
```
mzorz: same here, we can get rid of `val maxDimension` and use `maxImagesWidth` instead, for both max and min parameters in this call

daniloercoli:
@mzorz This is now ready for another pass. I've added better testing steps as well.

mzorz:
Thank you for the details @daniloercoli ! 
Tested, works alright
LGTM :shipit:

PR #598, opened 2018-01-11
Issue/259 calypso underline

### LINKED ISSUE #259: Underline tag used in calypso doesn't import style in Aztec
In Calypso the underline tag is implemented using the following:

`<span style="text-decoration: underline;">underlined text</span>`

Because we use the `<u>` tag this is simply ignored.

#### Tested

0nko:
Fixes #259.

The CSS-style underline support is implemented using the span preprocessing/postprocessing plugins. Basically any span with `text-decoration: underline` is converted to `AztecUnderlineSpan` and vice versa. I tried adding parallel support both for `<u>` and spans but since there is only a single underline styling button and it resulted in a mess.


aztec/src/main/kotlin/org/wordpress/aztec/source/InlineCssStyleFormatter.kt
```diff
@@ -36,21 +33,66 @@ class InlineCssStyleFormatter {
...
+            }
+        }
+
+        private fun getPattern(styleAttr: String): Pattern {
+            return Pattern.compile(
+                    "(?:;|\\A)$styleAttr:(.+?)(?:;|$)", Pattern.CASE_INSENSITIVE or Pattern.MULTILINE)
+        }
+
+        private fun processColor(style: String, text: Editable, start: Int, end: Int) {
+            val m = getPattern(CSS_COLOR_ATTRIBUTE).matcher(style)
```
daniloercoli: What do you thin if we move `getPattern(CSS_COLOR_ATTRIBUTE)`, and the other call `getPattern(CSS_TEXT_DECORATION_ATTRIBUTE)`, to a static variable in this class?
I see those methods called many times when I pass a text that does contain `style` attribute.

aztec/src/main/kotlin/org/wordpress/aztec/source/InlineCssStyleFormatter.kt
```diff
@@ -36,21 +33,66 @@ class InlineCssStyleFormatter {
...
+
+                if (newStyle.isBlank()) {
+                    attributes.removeAttribute(STYLE_ATTRIBUTE)
+                } else {
+                    attributes.setValue(STYLE_ATTRIBUTE, m.replaceAll(""))
+                }
+            }
+        }
+
+        fun addStyle(attributes: AztecAttributes, styleAttribute: String) {
```
daniloercoli: This is not used anywhere. Do you think we should keep it for future usage?

aztec/src/main/kotlin/org/wordpress/aztec/source/InlineCssStyleFormatter.kt
0nko: What about making it more general - I could I add a method for retrieving style attributes `getStyleAttribute(attributeName)` so you don't have to deal with regex matching? I'd also change `containsUnderlineDecorationStyle()` to a more general public `containsStyle(styleAttr): Boolean`.

aztec/src/main/kotlin/org/wordpress/aztec/source/InlineCssStyleFormatter.kt
0nko: This is sort of a utility class for CSS manipulation. I think it could be useful in the future if we support more styles.

aztec/src/main/kotlin/org/wordpress/aztec/source/InlineCssStyleFormatter.kt
daniloercoli: That seems a nice idea!

0nko:
This took a bit more work than I expected. When I was testing some edge cases I realized the underline `<span>` element nesting wasn't incorrect due to broken custom hidden span nesting logic. I've used the `IAztecNestable` interface instead and the parsing is now a lot simpler. I added some unit tests to make sure everything is working as expected.

daniloercoli:
I've tried this PR with a simple demo text that has both one `u` tag, and one span with style underline in it.
- It seems that the `u` tag is converted to `span` tag with styling when switching to HTML mode. 
- It seems that single space character are converted to &nbsp; entity in HTML mode.

To repro the issue use the following text
```
        private val EXAMPLE = "Test <span style=\"text-decoration: underline;\">underline here </span>\n" +
                "\n" +
                "<u>underline with u tag</u>"
```

daniloercoli:
Tested it further by adding additional content in visual mode (just another underlined text) and got this:

screenshot_20180118-122234631988776

0nko:
> - It seems that the u tag is converted to span tag with styling when switching to HTML mode.

Yes, this is working as intended. It's assumed that when you use the plugin you want to use the CSS style. I considered supporting both at the same time but it would become messy when manipulating the underline with the toolbar button in the visual mode, since
> - It seems that single space characters are converted to nbsp; entities in HTML mode.

I wasn't able to reproduce this using the sample text you provided. I tried it on API 25 and 27 and both with Calypso mode enabled and disabled.

> Tested it further by adding additional content in visual mode (just another underlined text) and got this

Fixed.

daniloercoli:
> Yes, this is working as intended. It's assumed that when you use the plugin you want to use the CSS style. I considered supporting both at the same time but it would become messy when manipulating the underline with the toolbar button in the visual mode, since

I mentioned this because Calypso behave differently, and keeps the `u` tags.

daniloercoli:
I see some tests failing now :(
Most (if not all of them) are about Formatting. Is it because `u` tags are converted to span with style?

daniloercoli:
We can merge this PR once lint issues are fixed.

Just to double check: The CssUnderlinePlugin seems to be off when running tests. I noticed that was done on purpose, not sure if it was a temporary commit or not.

0nko:
> Just to double check: The CssUnderlinePlugin seems to be off when running tests. I noticed that was done on purpose, not sure if it was a temporary commit or not.

Yeah, the plugin breaks the existing formatting tests. I'll add a couple of new ones where a plugin will be explicitly added to test the new functionality.

daniloercoli:
:shipit:

PR #601, opened 2018-01-15
Add the ability to log the content in case of crashes

### LINKED ISSUE #558: Add the ability to Log the whole content of the editor in case of errors/warnings
We've added our Logging framework to Aztec in #550. We should add the ability of logging the whole content of the editor in case of errors/warnings. Otherwise it will be hard for us to replicate the issue in case of crashes or other problems with the editor.

**Note:** We should add a new flag and a new interface to the editor, that will be used by the main app, in order to make the content "loggable" or not. The main app, when instantiates Aztec, could enable the logging of the content or not. This is because we should not log the content on pvt blogs, or other sensitive content. Maybe private posts too.

daniloercoli:
Fixes #558 by adding a global exception handler that logs all the content of the editor in case of crashes.
Also adds an interface that can be used by 3rd party apps, to enable/disable log on a per case basis. The global exception handler, before logging the content, asks if it should be logged or not.

### Test
- Force an exception in MainActivity, and check Logcat. It should contains the content of the editor.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -364,6 +364,7 @@ open class MainActivity : AppCompatActivity(),
                 .setOnAudioTappedListener(this)
                 .setOnMediaDeletedListener(this)
                 .setOnVideoInfoRequestedListener(this)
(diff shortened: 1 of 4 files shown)


mzorz:
Tested, LGTM :shipit:

PR #604, opened 2018-01-15
Add the ability for apps to set an external logger

daniloercoli:
This PR adds the ability for apps that use Aztec to set an external logger on the editor.

The external logger should be used to track important errors and info only. In our case we should use it in wp-android to keep tracks of some exceptions and messages in Crashlytics.

This PR already adds the logger to track InlineFormatter span issues.


mzorz:
Tested (artificially simulating the `start > end` condition is met) and works alright 
LGTM :shipit:

Issue #335, opened 2017-04-30
Style lost when deleting a newline before a block span


0nko:
Edit: Updated steps to reproduce in the comment here: https://github.com/wordpress-mobile/AztecEditor-Android/issues/335#issuecomment-298574701

(Leaving the following steps for reference)
~~1. Start with a newline followed by a block span (heading, for example)
2. Put the cursor at the start of the block span
3. Remove the newline with a backspace
4. Notice the original block style was removed~~

0nko:
The style is being removed by the the [ParagraphCollapseRemover](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/watchers/ParagraphCollapseRemover.kt#L74). The logic seems to assume that by deleting a newline we want to merge 2 code block spans, which is not always the case. @hypest, could you confirm this please?

hypest:
The behaviour observed with the described steps is the intended one. Deleting the newline before some block element "attracts" the text into the block element before. For example, in the demo app, deleting the newline at the end of "Heading 1" (either by deleting it, or backspacing from "Heading 2's line) causes the "Heading 2" text to be included in "Heading 1"'s block style. That's how text is extracted out of the start of a block element as well.

The same behaviour, for example, is observed while deleting the final newline from a list, causing text that lies after the list to be attracted _into_ the list.

What should be the desired behaviour instead? Why? Some specific examples might also help. Thanks!

hypest:
> The style is being removed by the the ParagraphCollapseRemover. The logic seems to assume that by deleting a newline we want to merge 2 block spans, which is not always the case. @hypest, could you confirm this please?

Here's what happening: `PARAGRAPH_SPAN`s try to re-anchor when their leading anchor (the preceding newline) is deleted from the text. The android framework tries to move the span's start to the next (moving towards the end of the text) available anchor. If the span doesn't include a newline other than its ending newline than the span will actually collapse and will be found at the end of the (whole) text. That's where `ParagraphCollapseRemover` comes into play, detecting that situation and "garbage collecting" the collapsing span.

In essence, it's not the ParagraphCollapseRemover that removes the span; it merely cleans up the text from collapsing (block) spans. The collapse itself happens because the span's leading anchor disappears. That's a "fortunate" event in our case since the way editing works, we do want the collapsing span to be removed and its text become assimilated to the preceding block.

[...]

PR #606, opened 2018-01-16
Issue/567 lost style

### LINKED ISSUE #567: Style lost when adding a new line before formatted text
### Expected

I expect to be able to place the cursor at the start of formatted text and press enter to move the formatted text to a new line, without affecting the formatting.

### Observed

Block elements:

When I place the cursor at the beginning of a block element (heading, list, quote) and press enter, the formatting is removed and no new line is added. Instead, the cursor moves forward one character.

Other formatting:

When I place the cursor at the beginning of the formatted text (e.g. bold, italic) and press enter, a new line is added but it also removes the formatting.

### Reproduced

Block elements:

1. In the demo app, enable a block formatting style (e.g. the "Heading 1" style).
2. Enter text (e.g. the letter "a").
3. Place the cursor at the beginning of the line.
4. Press enter.

Result: The style is removed and the cursor moves one character to the right (after the "a").

Non-block elements:

1. In the demo app, enable a regular formatting style (e.g. the bold style).
2. Enter text (e.g. the letter "a").
3. Place the cursor at the beginning of the line.
4. Press enter.

Result: A new line is added but the style is removed from the text.

#### Tested
Nexus 9 on Android 7.1.1 with develop

0nko:
Fixes #567 and #335.

Just FYI, the case when a backspace is pressed at the beginning of a heading must be handled in the `beforeTextChanged()` listener because the spans must be manipulated before the text gets removed for it to work.

Also, I didn't apply the fix above to a blockquote because I thought it makes more sense to delete it rather that move it up. But that can be changed... What do you think @rachelmcr?


rachelmcr:
>Also, I didn't apply the fix above to a blockquote because I thought it makes more sense to delete it rather that move it up.

Sounds good to me. That's how the Calypso editor behaves, which is what I use as a measure of expected behavior.

mzorz:
I hate to say this but the fix doesn't work on API 26 (Android 8). 😕 
Tested both with and without content as per comments in the #567 issue, and it doesn't work either way.
Will look into it more closer and hope to find it's a candidate for the buffered detection mechanism.
- Tested it on 4.2.2 and works alright ✅ 
- Tested it on 5.1.0 and works alright (note that the simple inline formatting like bold or italic was not lost on this platform, checked this on `develop`) ✅ 
- Tested it on 7.0.0, works alright except for a minor glitch in line heights (which I guess we can live with, but keeping the steps to reproduce here for the record):
```
1. go to last line and hit enter to make a new line
2. tap B to toggle bold
3. enter text “a”
4. place the cursor to the left of the character (beginning of line)
5. tap enter
6. observe the format is not lost, which is great
7. now place the cursor to the right and tap enter, to be on the last line
8. enable H1 heading
9. enter letter “a”
10. place the cursor to the left
11. tap ENTER
12. observe the formatting is kept, which is good

13. now go to the empty line between the last Aztec demo app content and the content you’ve been entering in steps above (the one generated at step 5)
14. tap on the H1 heading to enable it
15. observe the last line below is moved down
16. enter some letter
17. now observe the line below adjusts its position

```
- Tested it on 6.0.0, works alright, same glitch observed as in 7.0.0 ✅

mzorz:
I've opened a new issue https://github.com/wordpress-mobile/AztecEditor-Android/issues/610 so we can handle the specific case of API26/27 separately from this PR 👍

aztec/src/main/kotlin/org/wordpress/aztec/watchers/BlockElementWatcher.kt
```diff
@@ -19,7 +22,35 @@ open class BlockElementWatcher(aztecText: AztecText) : TextWatcher {
...
+
+                    if (spans.isNotEmpty()) {
+                        // save the text state before the funky business, then skip the history
+                        val aztecText = aztecTextRef.get()
+                        aztecText?.let {
+                            aztecText.history.beforeTextChanged(aztecText.toFormattedHtml())
+                            aztecText.consumeHistoryEvent = false
+
+                            spans.forEach {
+                                spannable.setSpan(AztecHeadingSpan(it.span.nestingLevel, it.span.TAG, it.span.attributes, (it.span as AztecHeadingSpan).headerStyle), deleteEnd - 1, deleteEnd, it.flags)
```
mzorz: I think we can remove the `as AztecHeadingSpan` cast and just use `it.span.headerStyle`

mzorz: also this line is too long? not sure if we're sticking to 130 chars max per line, but on my screen I had to scroll. It'd be more readable if we broke it like this:

```
                            spans.forEach {
                                spannable.setSpan(AztecHeadingSpan(it.span.nestingLevel, it.span.TAG, it.span.attributes,
                                        it.span.headerStyle), deleteEnd - 1, deleteEnd, it.flags)
                            }

```

mzorz:
This is working for me @0nko 👍 
Just left some comments to be addressed and then we'll be good to go

aztec/src/main/kotlin/org/wordpress/aztec/watchers/BlockElementWatcher.kt
```diff
@@ -19,7 +22,35 @@ open class BlockElementWatcher(aztecText: AztecText) : TextWatcher {
 
     private val aztecTextRef: WeakReference<AztecText?> = WeakReference(aztecText)
 
-    override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) { }
+    override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) {
+        if (count > 0) {
+
+            val deleteEnd = start + count
+            if (text[deleteEnd - 1] == Constants.NEWLINE && (deleteEnd - 1 == 0 || text[deleteEnd - 2] == Constants.NEWLINE)) {
```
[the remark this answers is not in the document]
0nko: Not possible, `deleteEnd - 1 == 0` check takes care of it, because `deleteEnd` is always greater than 0.

aztec/src/main/kotlin/org/wordpress/aztec/watchers/BlockElementWatcher.kt
0nko: Weird, the lint checker complained about not being able to infer the type without the cast before. Maybe the new Kotlin release fixed it?

aztec/src/main/kotlin/org/wordpress/aztec/watchers/BlockElementWatcher.kt
```diff
@@ -19,7 +22,35 @@ open class BlockElementWatcher(aztecText: AztecText) : TextWatcher {
...
+
+                // if a newline is deleted right above a heading, we want to preserve it and move it up
+                if (text[deleteEnd - 1] == Constants.NEWLINE && (deleteEnd - 1 == 0 || text[deleteEnd - 2] == Constants.NEWLINE)) {
```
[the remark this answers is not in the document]
mzorz: addressed in a684ecf

mzorz:
:shipit:

PR #607, opened 2018-01-16
Make sure all OS-originated .DS_Store files are ignored

mzorz:
This PR just makes sure that OS-generated files `.DS_Store` in any directory (not just project root) are ignored in git.


0nko:
:shipit:

PR #608, opened 2018-01-17
Fix for buffered detection mechanism with nested TextWatchers

### LINKED ISSUE #605: Space insert before formatted text not working when not in first line
This seems like a regression of #530, addressed by #555. 

### Expected
Space insert should never break formatting

### Observed
Space insertion breaks formatting when it's done on the second or N+1 lines

### Reproduced
1. get on an empty AztecText (i.e. demo app without content)
2. start typing "he"
3. toggle Bold
4. type "llo"
5. place the cursor right before the bold-formatted text, like "he|**llo**" and tap SPACE
6. observe the bold text remains in bold, while the non-formatted text remains unstyled, like "he **llo**" which is expected.
7. place the cursor at the end of the line and hit ENTER
8. on this new line, repeat steps 2 to 5.
9. observe that the second line loses formatting (everything is unstyled, like "he llo").

#### Tested
[Google Pixel] on [8.0, API 26] with [latest develop demo]

mzorz:
### Fix #605 

The buffered detection mechanism relies on a pattern (sequence of events) being consistent in order to be able to detect such a pattern, but due to the fact that we have chained recursive `TextWatchers` the events come in a different order (as nested events come in between expected events), thus rendering the pattern undetectable to the mechanism.

This PR implements a slight addition to the detection mechanism so it can tell whether the `TextWatcher` events correspond to the main nesting level or not, and only those who correspond to the first nesting level can make it to the `ObservationQueue`.

### Test
1. get on an empty AztecText (i.e. demo app without content)
2. start typing "he"
3. toggle Bold
4. type "llo"
5. place the cursor right before the bold-formatted text, like "he|**llo**" and tap SPACE
6. observe the bold text remains in bold, while the non-formatted text remains unstyled, like "he **llo**" which is expected.
7. place the cursor at the end of the line and hit ENTER
8. on this new line, repeat steps 2 to 5.
9. observe that the second line replicates the same behavior as in point 6 above.
10. do this as many times as you want, and verify the pattern is always detected.

### Review
@0nko

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -135,6 +135,8 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
 
         val DEFAULT_IMAGE_WIDTH = 800
 
+        var watchersNestingLevel: Int = 0
+
         private fun getPlaceholderDrawableFromResID(context: Context, @DrawableRes drawableId: Int, maxImageWidthForVisualEditor: Int): BitmapDrawable {
             val drawable = ContextCompat.getDrawable(context, drawableId)
             var bitmap: Bitmap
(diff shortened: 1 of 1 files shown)


0nko:
Tested the steps and it works.

:shipit:

PR #616, opened 2018-01-23
Only care for collecting TextWatcher events for the correct platforms

mzorz:
Good catch @daniloercoli - we were collecting events in the `ObservationQueue` on every platform, even when there were no active buckets for a specific platform per se.

This PR takes care of checking there are active buckets for the ObservationQueue before attempting to add a copy of the event to it.

### Test
On API < 26 (i.e. Android 7.x or anything less than 8)
1. put  a breakpoint in line 869 in `beforeTextChanged` in AztecText.kt
2. type something
3. notice it should never stop execution there

On API == 26/27
1. put  a breakpoint in line 869 in `beforeTextChanged` in AztecText.kt
2. type something
3. notice it should  stop execution there.

### Review
@daniloercoli


daniloercoli:
This PR seems OK to me, and can be merged.

Profiling it seems that the history is now the major problem in regard to performances.
screen shot 2018-01-23 at 22 35 01

PR #617, opened 2018-01-24
Run instrumentation tests in Firebase Test Lab from travis-ci

maxme:
Run instrumentation tests in Firebase Test Lab from travis-ci


0nko:
Excellent! The build takes 25 minutes now  but we can finally sleep at night knowing we didn't break the UI tests in develop 😁.

PR #619, opened 2018-01-25
Issue/610 api26 style lost prepend new line

mzorz:
### Fix #610
A continuation of https://github.com/wordpress-mobile/AztecEditor-Android/pull/606, This implements the right sequences as observed and described in API26+ in #610, so we can re-work the original intent from user when these situations occur:
- prepending a new line to an inline-styled (for example, bold, italic, etc) piece of text should now not make the style be lost (implemented in `API26PrependNewLineOnStyledTextEvent`)
- prepending a new line to a Block-styled (for example, a Header) piece of text should now not make the style be lost (implemented in `API26PrependNewLineOnStyledSpecialTextEvent`)

Also we fixed a bad behavior that would make the `API26InWordSpaceInsertionEvent` sequence be detected when a space is inserted within a Block, but Blocks behave well in this case, so now there are 3 values that can be returned by `isUserOperationObservedInSequence`, as described here:
```
    enum class ObservedOperationResultType {
        SEQUENCE_FOUND,
        SEQUENCE_NOT_FOUND,
        SEQUENCE_FOUND_CLEAR_QUEUE
    }
```
This way, the third enum value `SEQUENCE_FOUND_CLEAR_QUEUE`  indicates the ObservationQueue that even when the sequence has been found, it can clear itself as no action will be taken.

### Test
1. tap bold
2. enter "a"
3. place the cursor to the beginning (left of "a")
4. tap ENTER
5. observe the style is not lost anymore.

Also repeat with H1 instead of Bold:
1. tap H1
2. enter "a"
3. place the cursor to the beginning (left of "a")
4. tap ENTER
5. observe the style is not lost anymore.

Note: if you're on a stock emulator, please use the soft keyboard for this

### Review
@0nko @daniloercoli

PR #620, opened 2018-01-25
Issue/618 onMediaDeleted called unnecessarily in sequence

### LINKED ISSUE #618: API26/27: onImageDeleted callback called when space inserted in a word
### Expected
`onImageDeleted` callback should only be called when the user deletes images.

### Observed
Inserting a space in a word provokes all media in a Post to go through the `onMediaDeleted` callback

This is because there are 4 actual TextWatchers events for every middle-of-word space insertion in API26+ (see https://github.com/wordpress-mobile/AztecEditor-Android/issues/555), and in these events both the text to the left and the right of the insertion index are deleted, thus triggering the `onMediaDeleted` callback if images exist within such blocks. 

### Reproduced
1. on the demo app, look in the middle of content and start a new line with a word, like "hello"
2. place the cursor somewhere in the middle of the word you just entered
3. tap SPACE
4. observe the Toasts saying "Media Deleted!" are shown for each of the images found in the content.

#### Tested
[Google Pixel] on [Android 8.0] with [latest `develop`]

mzorz:
### Fix #618 

This PR implements a small (500ms) wait on the call to `onMediaDeleted` callback to make sure that, if an `AztecMediaSpan` has been deleted, it has been due to a user-triggered action rather than an event generated by the platform as discovered in API26+.

The check makes sense by identifying whether the `ObservationQueue` is being populated (with events that are very close together in time). If it's not the case, we can have a good amount of confidence that this is a user-triggered event, and so we end up triggering the actual `onMediaDeleted` events collected in the meantime.

Note that the 100ms time distance in between events is the one as already being used in `API26InWordSpaceInsertionEvent`. This value is being taken over to `ObservationQueue` on another PR. As both PRs are merged, we can get rid of this value declaration [here](https://github.com/wordpress-mobile/AztecEditor-Android/compare/issue/618-onimagedeleted-called-in-sequence?expand=1#diff-7ca71cd1f009ab8438e0901205001b14R857) and make use of the constant defined in `ObservationQueue` in `https://github.com/wordpress-mobile/AztecEditor-Android/pull/619/files#diff-5c63c1be1a17a9099fee1cf7d4476ea4R82`

### Test
On API26/27, using the demo app:

TEST A:
1. anywhere in the text, insert a line and type some letters, for example "hel".
2. now toggle Bold
3. now type "lo" (so the "lo" part is styled in bold)
4. place the cursor right before "lo", it should be seen like this at this position: "hel|**lo**""
5. now press SPACE - this will trigger the event sequence the `ObservationQueue` is watching for on API26+
6. Observe no "Media Deleted!" Toasts are shown.

TEST B:
1. now place the cursor right after any of the images in the demo app content
2. press BACKSPACE
3. observe the "Media Deleted!" Toast is shown accordingly.

TEST C:
On any other supported API, using the demo app, repeat TEST A.

### Review
@daniloercoli @0nko

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
@@ -8,6 +8,9 @@ import java.lang.ref.WeakReference
 
 class DeleteMediaElementWatcher(aztecText: AztecText) : TextWatcher {
     private val aztecTextRef: WeakReference<AztecText?> = WeakReference(aztecText)
+    private var deleted = false
+    private var queueHasBeenPopulatedInThisTimeframe = false
+    private var deletedSpans = ArrayList<AztecMediaSpan>()
 
     override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) {
         if (aztecTextRef.get()?.isTextChangedListenerDisabled() ?: true) {
@@ -15,10 +18,25 @@ class DeleteMediaElementWatcher(aztecText: AztecText) : TextWatcher {
         }
 
         if (count > 0) {
+            deleted = true
             aztecTextRef.get()?.text?.getSpans(start, start + count, AztecMediaSpan::class.java)
                     ?.forEach {
-                        it.onMediaDeleted()
+                        deletedSpans.add(it)
                     }
+
+            // only call the onMediaDeleted callback if we are sure the ObservationQueue has not been filled with
+            // platform-only events in a short time. These platform-originated events shall not be confused with
+            // real user deletions.
+            aztecTextRef.get()?.postDelayed( object : Runnable {
+                override fun run() {
+                    if (!queueHasBeenPopulatedInThisTimeframe) {
+                        deletedSpans.forEach { it.onMediaDeleted() }
+                    }
+                    // reset flag
+                    deletedSpans.clear()
+                    queueHasBeenPopulatedInThisTimeframe = false
+                }
+            }, 500)
         }
     }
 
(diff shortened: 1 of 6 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcher.kt
```diff
@@ -8,17 +8,38 @@ import java.lang.ref.WeakReference
...
             aztecTextRef.get()?.text?.getSpans(start, start + count, AztecMediaSpan::class.java)
                     ?.forEach {
-                        it.onMediaDeleted()
+                        deletedSpans.add(it)
+                    }
+
+            // only call the onMediaDeleted callback if we are sure the ObservationQueue has not been filled with
+            // platform-only events in a short time. These platform-originated events shall not be confused with
+            // real user deletions.
+            aztecTextRef.get()?.postDelayed( object : Runnable {
```
daniloercoli: This PR works fine, tested on API26+ and lower APIs, and I think we can merge it.
I have one consideration though. May I suggest to split `DeleteMediaElementWatcher.kt` in 2 classes, one for API26+ that does use `postDelayed` and the other for API25 and lower?

If you think it doesn't worth the effort I'm ok with this, but the Runnable is being executed a lot of times on API25 devices, even when no media are deleted, with `deletedSpans` empty, and this is another point that could slow down Aztec even more.

daniloercoli:
@0nko - I've made some small refactoring to this PR, in order to improve performances on older devices (API 25 and lower). So you may need to review this one.

I've split DeleteMediaElementWatcher in two classes. One takes care of API26+devices, and the other class API25 and lower. This way all the new delayed checks for API26+ are only done in one class, and older devices should not be affected by these other PostDelay calls.

mzorz:
> I've split DeleteMediaElementWatcher in two classes. One takes care of API26+devices, and the other class API25 and lower. This way all the new delayed checks for API26+ are only done in one class, and older devices should not be affected by these other PostDelay calls

Excellent thanks for taking care of this @daniloercoli ! Agreed with the changes :+1:

0nko:
Working as expected. :shipit:

PR #623, opened 2018-01-30
Log Aztec details when HTML is not available

daniloercoli:
Adds the ability to log details of the editor when HTML representation of content is not available. This could happen in case of errors in `toHTML` method of `AztecText`, that is used to get the HTML code of the content.

### Review
@0nko

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/util/Azteclog.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/util/Azteclog.kt
@@ -1,9 +1,42 @@
 package org.wordpress.aztec.util
 
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+import org.wordpress.android.util.AppLog
+import org.wordpress.aztec.AztecText
+import org.wordpress.aztec.spans.IAztecAttributedSpan
+
 class AztecLog {
     interface ExternalLogger {
         fun log(message : String)
         fun logException(tr : Throwable)
         fun logException(tr : Throwable, message : String)
+    }
+
+    companion object {
+        fun logEditorContentDetails(aztecText: AztecText) {
+            try {
+                val logContentJSON = JSONObject()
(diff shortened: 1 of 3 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/util/Azteclog.kt
```diff
@@ -1,9 +1,42 @@
...
     }
+
+    companion object {
+        fun logEditorContentDetails(aztecText: AztecText) {
+            try {
+                val logContentJSON = JSONObject()
+                logContentJSON.put("content", aztecText.text.toString())
+                logContentJSON.put("length", aztecText.text.length)
+                val spansJSON = JSONArray()
+                val spans = aztecText.text.getSpans(0, aztecText.text.length, IAztecAttributedSpan::class.java)
```
0nko: There could be spans of interest that don't implement `IAztecAttributedSpan`: `UnknownHtmlSpan`, `UknownClickableSpan`, `CommentSpan`, `AztecMediaClickableSpan`, `AztecCursorSpan`, `GutenbergCommentSpan` and `WordPressCommentSpan`.

0nko: I think the unknown spans and comment spans could be useful, since they could contain HTML that affects parsing.

daniloercoli: Right, I've changed the code so all types of spans are logged in case of crash. Better to log everything.

0nko: I don't see the change. Maybe you forgot to push your commit?

daniloercoli: 🤦‍♂️

0nko:
:shipit:

Issue #465, opened 2017-08-11
Text alignment support


0nko:
Support left, center and right alignment of paragraphs.

Related iOS issue: https://github.com/wordpress-mobile/AztecEditor-iOS/issues/649

ChrissiePollock:
This was requested again in internal issue #217678-h

ChrissiePollock:
requested again in #238383-h

ChrissiePollock:
and again #238453-h

SiobhyB:
This was requested again in 256661-h.

daniloercoli:
Addressed in #624.

PR #624, opened 2018-01-31
Text alignment support

0nko:
Addresses #465. 

All inline and block elements can now by aligned to the left, right and center. If alignment is applied to plain text it is surrounded by `<p>` tags. Multi-selection is also supported.

The alignment formatting is parsed to HTML as a CSS style.

The toolbar button functionality is overloaded for the `[caption]` shortcode, where it's used for caption-specific alignment parsing (e.g. `align="alignright"`).

Don't be put off by the large number of changes as a lot of them are just resource files and unit tests.

Note that the cursor position is behaving a little weird around empty hidden spans, because special zero-width characters are used for the spans.

screenshot_1517417175


daniloercoli:
When I start the demo app the caption is right aligned even though there is no `align="alignright"` specified.
device-2018-02-01-151805

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
```diff
@@ -368,7 +369,19 @@ class AztecParser(val plugins: List<IAztecPlugin> = ArrayList()) {
     }
 
     private fun withinNestable(out: StringBuilder, text: Spanned, start: Int, end: Int,
-                               nestable: IAztecSpan, parents: ArrayList<IAztecNestable>?, nestingLevel: Int) {
+                               nestable: IAztecParagraphStyle, parents: ArrayList<IAztecNestable>?, nestingLevel: Int) {
+
+        CssStyleFormatter.removeStyleAttribute(nestable.attributes, CssStyleFormatter.CSS_TEXT_ALIGN_ATTRIBUTE)
```
daniloercoli: It's not clear to me why we're removing the text-align style attribute, and then re-applying it in case `nestable.shouldParseAlignmentToHtml()` is true.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
```diff
@@ -209,7 +238,7 @@ class BlockFormatter(editor: AztecText, val listStyle: ListStyle, val quoteStyle
             val endModifier = if (hasLingeringEmptyListItem) 1 else 0
 
             val spans = editableText.getSpans(start, end + endModifier, spanType)
-            spans.forEach { span ->
+            spans.forEach innerForEach@ { span ->
```
daniloercoli: Niptick - Is `innerForEach@` needed here?

0nko:
About the caption alignment - [it's here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L129) :).

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
0nko: Thanks, it's a leftover from a change that required it but eventually got removed.

aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
0nko: Hmm, I'm trying to think of a case when this could be a problem. It's not an issue for captions because the plugin removes the style attribute. But I suppose we can move it within the if statement.

daniloercoli:
Impressive.  :shipit:

PR #628, opened 2018-02-05
Issue/527 quote issues

### LINKED ISSUE #527: Can't remove blockquote style from a list correctly
### Expected

I expect that when I create a list inside a blockquote, toggling off the blockquote style will remove it completely so only the list style is left.

For example, starting with this:

```
<blockquote>
<ul>
<li>first item</li>
<li>second item</li>
<li>third item</li>
</ul>
</blockquote>
```

I expect that with my cursor at any point in the list, toggling off the blockquote style will give me this:

```
<ul>
<li>first item</li>
<li>second item</li>
<li>third item</li>
</ul>
```

### Observed

There are two scenarios:

1. If you have the cursor on the first line of the quoted list, and you toggle off the blockquote style, it appears to only remove the blockquote style from the first list item. However, switching to HTML view shows that it was removed from the entire list.
2. If you have the cursor on any other line of the quoted list, and you toggle off the blockquote style, it appears to remove the blockquote style from only that list item. However, switching to HTML view shows that the full list is still inside the blockquote and a duplicate list without the blockquote style appears underneath it, starting from the list item where the cursor was located.

### Reproduced

First scenario:

1. Toggle on the blockquote style.
2. Toggle on a list style.
3. Add several list items.
4. Move the cursor to the first list item and tap the blockquote button to toggle it off. The blockquote style still appears to be applied to the other list items.
5. Switch to HTML view.

Resulting HTML:

```
<ul>
<li>first item</li>
<li>second item</li>
<li>third item</li>
</ul>
```

Second scenario:

1. Toggle on the blockquote style.
2. Toggle on a list style.
3. Add several list items.
4. Move the cursor to the second list item and tap the blockquote button to toggle it off. The blockquote style still appears to be applied to the other list items, but not the second list item.
5. Switch to HTML view.
6. Switch back to Visual view.

Resulting HTML:

```
<blockquote>
<ul>
<li>first item</li>
<li>second item</li>
<li>third item</li>
</ul>
</blockquote>
<ul>
<li>second item</li>
<li>third item</li>
</ul>
```

#### Tested
Pixel emulator on API 26 with develop branch

0nko:
Fixes #527. Toggling the quote on any line removes the entire blockquote. 

As I worked on this issue I've discovered a series of problems, namely incorrect block nesting when applying one block over another, applying block elements over list items, applying block elements across different levels of nesting (plain text + list item, for example), broken list formatting when applied to the last line, and more. I've refactored the block styling logic so that addresses all of the above.


daniloercoli:
Weird, I see `testRemoveQuoteFormatting` failing, even though Travis is green.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
```diff
@@ -166,6 +166,16 @@ class BlockFormatter(editor: AztecText, val listStyle: ListStyle, val quoteStyle
         removeBlockStyle(textFormat, selectionStart, selectionEnd, makeBlock(textFormat, 0).map { it -> it.javaClass })
     }
 
+    fun <T : IAztecBlockSpan> removeEntireBlock(type: Class<T>) {
+        val bounds = getBoundsOfText(editableText, selectionStart, selectionEnd)
```
daniloercoli: Would you mind adding some comments for future reference?
Also, I'm not not sure I completely understood it. Can't we just get `BlockQuote` spana that fall in `selectionStart, selectionEnd` and remove them?

0nko: Nice catch, this is redundant.

daniloercoli: Is the call to `val bounds = getBoundsOfText(editableText, selectionStart, selectionEnd)` required? 
Can't we use `selectionStart,selectionEnd` when getting the spans like the snippet below, and avoid that call?

```
    fun <T : IAztecBlockSpan> removeEntireBlock(type: Class<T>) {
        editableText.getSpans(selectionStart,selectionEnd, type).forEach {
            editableText.removeSpan(it)
        }
    }
```


I also see testRemoveQuoteFormatting failing. Maybe we need to refresh it?

0nko:
@daniloercoli, this PR is ready for another pass. About the failing UI tests - it seems like something has changed with the emulator and styled text is not copied as HTML anymore. The tests are also failing in `develop` so it's not caused by this.

daniloercoli:
@0nko I see Lint errors on this PR. Would you mind to fix it? In the meanwhile I start reviewing it.

daniloercoli:
I've tested this PR a bit and found out some weird behaviours when mixing list and block-quote.

First scenario:
- Start with an empty editor.
- Toggle on the blockquote style.
- Toggle on a list style.
- Add a couple of items list items.
- Switch to HTML view.

You will see that the HTML code isn't the right one:
 
screen shot 2018-02-26 at 15 54 07
screen shot 2018-02-26 at 15 54 16

0nko:
Ready for another pass. This thing was another newline can of worms but I think I got all the edge cases nailed 😓. 

The problem was that the list did not expect multi-line blocks inside list items, so they'd show the line indicator. This wasn't obvious previously because list items that contain other blocks didn't show (which was wrong) any line indicators.

I've refactored the way that lines are counted. I've tweaked & tested it with all kinds of different blocks inside lists, such as headings, quotes and also other lists (which should only display the inner line indicators).

daniloercoli:
It seems to me that the background applied to a blockquote item doesn't work fine on RTL. 
**This is on develop**
screenshot_20180228-124126

**This PR**
screenshot_20180228-124337

0nko:
I found a few more problems with block-closing using a newline, when there are multiple nested blocks. The RTL quote background is also fixed now. Thanks for the thorough review, @daniloercoli!

daniloercoli:
Ship it! 🥇 :shipit: 🚀

PR #630, opened 2018-02-05
Fix order of multiple media

### LINKED ISSUE #627: Multiple images are added in incorrect order
Start with a blank demo app, and add > 3 pictures in a row.
The 3rd picture is added to the editor between the 1st and the 2nd.

The problem lies in `insertMedia` in `LineBlockFormatter`.
Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/7199

daniloercoli:
This PR fixes #627 by making sure the cursor is correctly placed in the editor when media is added to it.

Before this PR there was an issue where starting from an empty editor, and adding 2 media to it, resulting in a cursor placed between the 1st and the 2nd picture. Then if you added a 3rd picture to the editor, it was being added in between the two before. As reported in `wp-android`.

**Before this PR**
-*Insert first media*
`newSelection` = 1 and text length=1 (correct, only one item in the editor, no new line at the end).

-*Insert second media*
`newSelection` = 2 1 and text length=3 (incorrect, the cursor is being placed in between the pictures). newSelection should be 3.

After this PR `newSelection` is being calculated correctly.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
@@ -174,7 +174,7 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
 
         // We need to be sure the cursor is placed correctly after media insertion
         // Note that media has '\n' around them when needed
-        val isLastItem = if (selectionEnd == EndOfBufferMarkerAdder.safeLength(editor)) { true } else false
+        val isLastItem = selectionEnd == EndOfBufferMarkerAdder.safeLength(editor)
         editableText.replace(selectionStart, selectionEnd, ssb)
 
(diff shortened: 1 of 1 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
```diff
@@ -172,10 +172,17 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
                 Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
         )
 
+        // We need to be sure the cursor is placed correctly after media insertion
+        // Note that media has '\n' around them when needed
+        val isLastItem = if (selectionEnd == EndOfBufferMarkerAdder.safeLength(editor)) { true } else false
```
0nko: Nitpick - no need for the curly braces.

0nko:
:shipit:

Issue #565, opened 2017-12-05
Missing padding after block elements in visual mode


rachelmcr:
### Expected

I expect to see padding after a block element (e.g. list or quote) in Calypso mode, as in Calypso web:

<img width="739" alt="calypsovisual-blocks" src="https://user-images.githubusercontent.com/8658164/33618200-2b1bc120-d9e2-11e7-81cd-1f767c1de5c5.png">

### Observed

There isn't any visual padding below block elements:

![wpandroidaztecvisual-blocks](https://user-images.githubusercontent.com/8658164/33618287-648d4b72-d9e2-11e7-9c06-f15735492ee4.png)

### Reproduced
1. In Calypso mode, create a new post with a blockquote.
2. End the blockquote and enter a new line of non-quoted text below the quote.
3. Compare the results with Calypso web (note that there is no space between the quote and the new line of text).

#### Tested
Nexus 9 on Android 7.1.1 with WPAndroid alpha-87

daniloercoli:
I had a look at this issue and seems that the only way to fix it is to add 2 NewLine characters instead of 1 in `AztecParser.addVisualNewlinesToBlockElements`. (and also change `AztecParser.syncVisualNewlinesOfBlockElements`.)
The downside of this solution is that the 2nd "new-line" is visible on the screen, and you can put the cursor on the empty line between 2 blocks.

I wish there was a way to just simply add some margin to each Span. Instead of doing text manipulation with "hidden" characters.

daniloercoli:
@0nko found a library that we can check to take inspiration to fix this issue.
See https://github.com/hanks-zyh/LineHeightEditText

PR #631, opened 2018-02-06
Try to fix - Missing padding between blocks

daniloercoli:
This PR tries to fix the missing margins between blocks (#565) by making edge lines slightly taller, and adapt the background and the leading margin when necessary, during the drawing phase.

Note these are only visual artifacts to simulate top and bottom margins. Unfortunately the cursor heigh is set by the system by using the line height. The drawback of this, is that edge lines also get a slightly taller cursor, that is noticeable on quote and pre-format (not in lists) since there is a BG color applied to them.

This PR is not ready to be merged, the code needs to be refactored avoiding duplicate code, and will do this if we think this is an acceptable solution to fix the issue.

cc @0nko 

The other possible fix to the issue would be adding a new type of visual only span, that just adds spaces between 2 blocks, but it's not easy to do that when inserting new content, or when parsing from HTML. It would complicate the TextWatchers chain even more.


daniloercoli:
I kinda forgot about this PR of mine! :D

> listens to text changes and sets a custom line height and custom cursor drawable 

I had tried with a custom cursor with no luck, but without listening to events. 
I agree with you, we should probably do something similar here. That library looks promising.

Issue #629, opened 2018-02-05
Crash in 1.1: IOOB in HeadingHandler.handleNewlineInBody


maxme:
Happen in v1.1 (wpandroid 9.2)

```
Fatal Exception: java.lang.IndexOutOfBoundsException: setSpan (159 ... 90) has end before start
       at android.text.SpannableStringBuilder.checkRange(SpannableStringBuilder.java:1302)
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:680)
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:672)
       at org.wordpress.aztec.handlers.BlockHandler$Companion.set(BlockHandler.kt:121)
       at org.wordpress.aztec.handlers.HeadingHandler$Companion.cloneHeading(HeadingHandler.kt:58)
       at org.wordpress.aztec.handlers.HeadingHandler.handleNewlineInBody(HeadingHandler.kt:39)
       at org.wordpress.aztec.handlers.BlockHandler.handleTextChanged(BlockHandler.kt:59)
       at org.wordpress.aztec.watchers.BlockElementWatcher.onTextChanged(BlockElementWatcher.kt:72)
       at android.widget.TextView.sendOnTextChanged(TextView.java:9359)
       at android.widget.TextView.setText(TextView.java:5395)
       at android.widget.TextView.setText(TextView.java:5248)
       at android.widget.EditText.setText(EditText.java:113)
       at android.widget.TextView.setText(TextView.java:5205)
       at org.wordpress.aztec.AztecText.executeEvent(AztecText.kt:1557)
       at org.wordpress.aztec.watchers.event.sequence.ObservationQueue.processQueue(ObservationQueue.kt:69)
       at org.wordpress.aztec.watchers.event.sequence.ObservationQueue.add(ObservationQueue.kt:35)
       at org.wordpress.aztec.AztecText.afterTextChanged(AztecText.kt:901)
       at android.widget.TextView.sendAfterTextChanged(TextView.java:9375)
       at android.widget.TextView$ChangeWatcher.afterTextChanged(TextView.java:11917)
       at android.text.SpannableStringBuilder.sendAfterTextChanged(SpannableStringBuilder.java:1262)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:574)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:504)
[...]

PR #632, opened 2018-02-08
"Fix" IOOB handling newline in Headings

daniloercoli:
I was not able to reproduce the issue reported in #629. In this PR I've just added log of the content, and avoided the crash.

I think we can change this behavior, and let the app crash, if we think it will be more useful to replicate the issue, since there now in place an exceptions logger, that bumps the content of the  editor into the log.

While "fixing" this I also noticed there was a check over the position of the text, that I think it should not be there, since the method `handleNewlineInBody` is only called when a new line is added "in body".

cc @0nko


0nko:
You're right that the check is misplaced. But simply removing it doesn't work, though. Try it with this HTML:`<h1>Heading</h1>text`. When you add a newline at the end of the heading you'll notice that the heading style continues on a new line. This specific condition should probably be handled in `BlockHandler.getNewlinePositionType()`.

daniloercoli:
Oh right, brought it back. And also tests were failing...

Do you think we should catch the exception (as done in this PR) or keep letting the app crashing?

0nko:
:shipit:

PR #634, opened 2018-02-11
Feature/rtl quote support

khaykov:
This PR adds a correct drawing of quote for RTL languages.

Drawing direction is determined based on the language of each individual line.

[https://gyazo.com/bceeb0352bc4f8d36e360a9300d447d4](https://gyazo.com/bceeb0352bc4f8d36e360a9300d447d4)

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -134,6 +134,7 @@ open class MainActivity : AppCompatActivity(),
         private val AUDIO = "[audio src=\"https://upload.wikimedia.org/wikipedia/commons/9/94/H-Moll.ogg\"]"
         private val VIDEOPRESS = "[wpvideo OcobLTqC]"
         private val VIDEOPRESS_2 = "[wpvideo OcobLTqC w=640 h=400 autoplay=true html5only=true3]"
+        private val QUOTE_RTL = "<blockquote>לְצַטֵט<br>same quote but LTR</blockquote>"
 
         private val EXAMPLE =
                 IMG +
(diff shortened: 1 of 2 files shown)


malinajirka:
Thanks for the PR @khaykov ;)! It looks good, however I've found 2 issues.

1. No text in the quote - LtR locale
default_ltr
No text in the quote - RtL locale
deafult_rtl
When there isn't any text in the quote, it automatically fallbacks to LtR quote. Maybe we could pick `TextDirectionHeuristicsCompat.FIRSTSTRONG_LTR` vs `TextDirectionHeuristicsCompat.FIRSTSTRONG_RTL` depending on the default locale? I haven't tested it, it's just an idea. Btw the leading margin rectangle is missing on the second image.

2. On api level 18 the leading margin rectangle doesn't show up in RtL mode. leading_marign_sdk_18

khaykov:
Good catch! I totally forgot to test empty quotes :) Defaulting to the locale is also a great idea - lists behave like this by default, so I change the quote to do it too.

malinajirka:
Hmm, it still doesn't work as expected in some cases - I wasn't able to find any pattern, it seems random:D. When I checked the "text" string using the debugger it had length = 1 but appeared to be empty.
I've tried this code and it seems to work as expected. WDYT @khaykov ?
```
val textDirectionHeuristic: TextDirectionHeuristicCompat = 
if (TextUtilsCompat.getLayoutDirectionFromLocale(Locale.getDefault()) == ViewCompat.LAYOUT_DIRECTION_RTL) {
   TextDirectionHeuristicsCompat.FIRSTSTRONG_RTL
  } else {
   TextDirectionHeuristicsCompat.FIRSTSTRONG_LTR
 }
return textDirectionHeuristic.isRtl(text, start, end - start)```

khaykov:
Hmm. Yes, it's weird. I noticed that it doesn't work for me every time too. I switched the `isRtlQuote` to your method, and looks like it works ok :) Thanks!

malinajirka:
I'm glad it works;). Thanks!

PR #635, opened 2018-02-12
Feature/rtl toolbar

khaykov:
This PR fixes toolbar in RTL direction.

[https://gyazo.com/a388afe138a6960ebc43b21e1a945cba](https://gyazo.com/a388afe138a6960ebc43b21e1a945cba)

To test:

1. Change locale to RTL language or toggle "Force RTL layout direction" in Dev options.
2. Check that basic and advanced toolbars work in RTL mode.

Note: Toolbar is using `HorizontalScrollView` which did not get  RTL support until API 19. On API 17 and 18 LTR layout will be forced.


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -702,43 +709,54 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
...
+                        //keep hard focus on media button to avoid it
+                        toolbarScrolView.requestChildFocus(buttonMediaCollapsed, buttonMediaCollapsed)
+                        layoutExpanded.startAnimation(layoutExpandedTranslateInEnd)
                     }
                 }
         )
     }
 
+    fun scrollToBeginingOfToolbar() {
+        if (TextUtilsCompat.getLayoutDirectionFromLocale(Locale.getDefault()) == ViewCompat.LAYOUT_DIRECTION_LTR
```
malinajirka: We might consider adding a comment explaining that HorizontalScrollView on older android versions doesn't support RtL layouts.

aztec/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<manifest package="org.wordpress.aztec">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="org.wordpress.aztec">
+
+    <application
```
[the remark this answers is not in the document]
khaykov: Thanks for explaining - I wasn't too sure about this myself :)  I moved the attribute to the app, which makes way more sense.

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
khaykov: Good call! I added a comment about RTL support to this method and to the toolbar itself.

khaykov:
Ready for another round 👍

malinajirka:
Thanks! ;)

Issue #561, opened 2017-12-05
Large text blocks make the app crash


mzorz:
### Expected
AztecText should handle a big piece of text

### Observed
Pasting a 1mb block of text twice at the end of the demo app, (sample content [used here](http://www.sample-videos.com/text/Sample-text-file-1000kb.txt)) then 
scrolling up and down, adding some text on the keyboard you’ll see that it lags a lot, but it works and keeps formatted text as you enter it, scroll up/down and it will be kept. It even works well when rotating the device as well.

At some point, if you go all the way down and tap on the last position (right of the last character) it will crash with this stacktrace:

```
12-05 16:25:08.962 6516-6516/org.wordpress.aztec D/AndroidRuntime: Shutting down VM
12-05 16:25:08.964 6516-6516/org.wordpress.aztec E/AndroidRuntime: FATAL EXCEPTION: main
                                                                   Process: org.wordpress.aztec, PID: 6516
                                                                   java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 1160676 bytes
                                                                       at android.view.autofill.AutofillManager.updateSessionLocked(AutofillManager.java:884)
                                                                       at android.view.autofill.AutofillManager.notifyViewEntered(AutofillManager.java:467)
                                                                       at android.view.autofill.AutofillManager.notifyViewEntered(AutofillManager.java:443)
                                                                       at android.view.View.notifyEnterOrExitForAutoFillIfNeeded(View.java:6942)
                                                                       at android.view.View.performClick(View.java:6264)
                                                                       at android.view.View$PerformClick.run(View.java:24697)
                                                                       at android.os.Handler.handleCallback(Handler.java:789)
                                                                       at android.os.Handler.dispatchMessage(Handler.java:98)
                                                                       at android.os.Looper.loop(Looper.java:164)
                                                                       at android.app.ActivityThread.main(ActivityThread.java:6541)
                                                                       at java.lang.reflect.Method.invoke(Native Method)
                                                                       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
                                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
                                                                    Caused by: android.os.TransactionTooLargeException: data parcel size 1160676 bytes
                                                                       at android.os.BinderProxy.transactNative(Native Method)
                                                                       at android.os.BinderProxy.transact(Binder.java:748)
[...]

PR #641, opened 2018-03-09
Save big instance data to files instead of the state bundle

hypest:
### Fix
Instead of saving big instance data to the state/bundle, preserve them in files. On state restore, read the data from the files.

This addresses the cases where the system needs to save/restore the AztecText's state, possibly including the case in #561 

### Test
1. In the `MainActivity.kt` change the `EXAMPLE` variable to match [this gist](https://gist.github.com/hypest/def4601866be90b2386aa55b2e95bf37)
2. Run the demo app. The big text will populate the editor.
3. Add a couple of characters in the beginning of text
4. Tap "Home" to background the app
5. Resume the app by using the "Recent apps" button
6. Notice the editor having the text, including the small change introduced in step 3

Feel free to try the steps above without the the fix to observe the demo app crashing with `TransactionTooLargeException`.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -602,16 +602,13 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
     }
 
     private fun writeTempInstance(filename: String, obj: Any?) {
-        with(File.createTempFile(filename, ".inst", context.getCacheDir())) {
-            deleteOnExit() // just make sure the file gets deleted if we don't do it ourselves
-            with(File(context.getCacheDir(), "$filename.inst")) {
-                with(FileOutputStream(this)) {
-                    with(ObjectOutputStream(this)) {
-                        writeObject(obj)
-                        close()
-                    }
+        with(File(context.getCacheDir(), "$filename.inst")) {
+            with(FileOutputStream(this)) {
+                with(ObjectOutputStream(this)) {
+                    writeObject(obj)
                     close()
                 }
+                close()
             }
         }
     }
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -635,7 +635,7 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
         val filename = bundle.getString(cacheFilenameKey(varName))
 
         if (TextUtils.isEmpty(filename)) {
-            return defaultValue;
+            return defaultValue
         }
 
         val file = File(filename)
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -108,6 +108,11 @@ import org.wordpress.aztec.watchers.event.text.BeforeTextChangedEventData
 import org.wordpress.aztec.watchers.event.text.OnTextChangedEventData
 import org.wordpress.aztec.watchers.event.text.TextWatcherEvent
 import org.xml.sax.Attributes
+import java.io.File
+import java.io.FileInputStream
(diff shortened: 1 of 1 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -596,6 +601,44 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
         return savedState
     }
 
+    private fun writeTempInstance(filename: String, obj: Any?) {
+        with(File.createTempFile(filename, ".inst", context.getCacheDir())) {
```
daniloercoli: Probably better catching `IOException` and `SecurityException` here.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -596,6 +601,44 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
...
+            deleteOnExit() // just make sure the file gets deleted if we don't do it ourselves
+            with(File(context.getCacheDir(), "$filename.inst")) {
+                if (this.exists()) {
```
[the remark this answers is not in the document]
hypest: The check-and-createNewFile code is actually wrong and unneeded. Left over from an intermediate state of the code :facepalm:. Removed it with f013904. While at it, removed more unneeded code with 68032e2.

Regarding your question, the file creation happens in the `FileOutputStream` constructor below anyway.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
hypest: Good call @daniloercoli ! I've added guards with 8af62f6. (Note, it's on a different code line since the `createTempFile` code was removed with 68032e2.

hypest:
Thanks for the review @0nko !

@daniloercoli , addressed your comment about the exception handling so, ready for another pass, thanks!

hypest:
Following a discussion over Slack, we think that using a unique filename for the temp files is more robust and better to be included in this PR, rather than later. I've added d7d98f0 which utilizes the system provided `createTempFile()` API and keeps the unique filename around by preserving it in the state bundle.

@0nko , @daniloercoli would be good if you could make another pass, thanks!

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -596,6 +603,64 @@ class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknownHtmlT
...
+                    ObjectOutputStream(output).use { objectOutput ->
+                        objectOutput.writeObject(obj)
+
+                        // keep the filename in the bundle to use it to read the object back
+                        bundle.putString(cacheFilenameKey(varName), this.path)
+                    }
+                }
+            }
+        } catch (e: IOException) {
+            AppLog.w(AppLog.T.EDITOR, "Error trying to write cache for $varName. Exception: ${e.message}")
```
daniloercoli: What do you think on logging errors about accessing the file in Crashlytics so we can see how many users have problem with this new method of writing state?
`externalLogger?.logException(e, "Error trying to write cache for $varName.")`

daniloercoli: When Aztec is integrated in `wp-android` the external logger is Crashlytics. The exception is wrapped in a special Aztec Exception and easily searchable in the web interface (I hope).

0nko:
Nice job! :shipit:

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
hypest: Makes sense @daniloercoli , will work on this on a separate PR since this one got merged already. 👍

koke:
It seems that this crash still happens sometimes in 9.7, but the bulk of the crashes is gone 👏 

screen shot 2018-04-10 at 10 32 44

maxme:
@koke I think it comes from another part of the app (it's hard to track the source off that one).

PR #644, opened 2018-03-13
Log cache writes errors to external logger too

hypest:
Follow up PR after #641 . This one just adds a call to the external logger to log cache file exceptions.


daniloercoli:
Looks good! :shipit:

PR #646, opened 2018-03-29
Add support for Touch Exploration

malinajirka:
Adds support for Touch Exploration to the AztecText and SourceViewEditText.

### Test
1. Enable TalkBack
2. Open the AztecApp
3. Hover over lines - each line should be read by talkback.
4. The cursor should be moved to the last position of your finger when you move up you finger.
5. Make sure that the TalkBack context menu contains "Cursor control/Editing" menu item(naming depends on the SDK version).

Issues
1. When we want to move the cursor to the new position, we need to remove the cursor first, otherwise the TalkBack reads the text between old and new cursor position.
The issues is that on SDK 27, TalkBack announces "selection cleared" when the ".removeSelection(...)" method is invoked.

2. It reads raw text ignoring spans. I wanted to add support for spans, but I run out of time. Some spans (eg. Comment, More, PageBreak) are in different modules, so I'm not even sure it's doable.

Note: The app sometimes crashed when I changed TalkBack navigation granularity (characters, words, lines, default etc..).  So I added if check into the getAppliedStyles(..) method, but tbh I'm not sure if it doesn't break something else. I've tested it and it seemed ok.... I just wanted to point it out.
Steps to reproduce (before the fix)
1. Enable TalkBack
2. Keep swiping right, after the editor view is focused (focused with talkback)
3. Swipe down to change TalkBack navigation granularity
4. Boom

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextAccessibilityDelegate.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextAccessibilityDelegate.kt
@@ -1,0 +1,100 @@
+package org.wordpress.aztec
+
+import android.os.Build
+import android.support.v4.view.accessibility.AccessibilityEventCompat
+import android.text.Selection
+import android.view.MotionEvent
+import android.view.View
+import android.view.accessibility.AccessibilityEvent
+import android.widget.EditText
+import org.apache.commons.lang3.StringUtils
+
+/**
+ * Delegate which adds support for ExploreByTouch to an EditText. TalkBack reads each line as the user hovers over them.
+ */
+class AztecTextAccessibilityDelegate(private val aztecText: EditText) {
+    private val ACCESSIBILITY_INVALID_LINE_ID = -1
+
+    private val mediaItemContentDescription = aztecText.getContext().getString(R.string.media_item_content_description)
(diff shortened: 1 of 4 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecTextAccessibilityDelegate.kt
```diff
@@ -0,0 +1,107 @@
...
+            }
+            if (lineOffset != ACCESSIBILITY_INVALID_LINE_ID) {
+                return true
+            }
+        }
+        return false
+    }
+
+    private fun updateContentDescription(parentForAccessibility: View, lineOffset: Int) {
+        // we can't use announceForAccessibility(..) as the announcement doesn't get interrupted as we move to another element
```
khaykov: Not sure if this will change anything, but `accessibilityManager` has `interrupt()` method :)

malinajirka:
>Not sure if this will change anything, but accessibilityManager has interrupt() method :)

Thanks @khaykov, I wasn't aware of the interrupt() method. 

I've replaced the View.setContentDescription() call with AccessibilityManager.interrupt() followed by View.announceForAccessibility(). I was a bit worried that the interrupt() method would interrupt announcements from other applications - new email notifications etc. But as it turned out announcements from other applications are getting interrupted whenever the AccessibilityFocus is changed. So we are ok, I guess.

PR #648, opened 2018-04-05
Feature/awesome accessibility toggle button

khaykov:
This PR fixes the issue with `ToggleButton` with menus (like Heading, List and media buttons) being checked even when nothing in the in the menu is selected. 

1. Enable TalkBack.
2. When no heading style is selected, focus on Headin button.
3. Make sure announcement says it's "not checked"
4. Open the Headin menu.
5. Without selecting anything (the Default option is OK) press back button and close the menu.
6. Focus on the Heading button and make sure it's announced as "not checked"

Same should apply to List and buttons inside Media toolbar.

In addition, I added correct checked/unchecked logic to Media button. When Media toolbar is invisible the button is unchecked when it's visible the button is checked.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
@@ -816,6 +816,7 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
                         buttonMediaCollapsed.visibility = View.GONE
                         buttonMediaExpanded.visibility = View.VISIBLE
                         buttonMediaExpanded.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)
+                        buttonMediaExpanded.isChecked = true
                     }
 
                     override fun onAnimationRepeat(animation: Animation) {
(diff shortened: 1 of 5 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -849,12 +851,22 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
         headingMenu = PopupMenu(context, view)
         headingMenu?.setOnMenuItemClickListener(this)
         headingMenu?.inflate(R.menu.heading)
+        headingMenu?.setOnDismissListener({
+            if (getSelectedHeadingMenuItem() == null || getSelectedHeadingMenuItem() == AztecTextFormat.FORMAT_PARAGRAPH) {
```
malinajirka: I believe we should add an 'else' branch with `isChecked = true`. WDYT?
Issue
1. Select H1 heading
2. Open heading pop-up menu
3. Without selecting anything press back button and close the pop-up menu.
4. Focus on the Heading button - talkback announces `not checked`

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -849,12 +851,22 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
...
+            }
+        })
     }
 
     private fun setListMenu(view: View) {
         listMenu = PopupMenu(context, view)
         listMenu?.setOnMenuItemClickListener(this)
         listMenu?.inflate(R.menu.list)
+        listMenu?.setOnDismissListener({
+            if (getSelectedListMenuItem() == null) {
```
malinajirka: I believe we should add an 'else' branch with `isChecked = true`. WDYT?

khaykov:
Thanks for checking the PR!
I totally missed that issue 🤦‍♂️  Fixed now!

malinajirka:
Looks good thanks!

PR #650, opened 2018-04-06
Prepare Aztec for integration in React Native - step 1

daniloercoli:
This PR is the beginning of required changes to Aztec to be integrated in a RN component.

Aztec was already in a good shape for integration in a RN component, in this PR we just refactored small things and added the ability to disable the media toolbar.

In details:
- Added the ability to completely disable the media toolbar. The `+` button is not shown on the screen.
- Introduced an interface that Toolbars need to implement.
- Made a couple of methods public, since we needed to extends AztecText in order to properly calculate the height in RN.

### Review
@mzorz @0nko

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
@@ -36,7 +36,7 @@ import java.util.Locale
  * Contains both Styling and Media toolbars.
  * Supports RTL layout direction on API 19+
  */
-class AztecToolbar : FrameLayout, OnMenuItemClickListener {
+class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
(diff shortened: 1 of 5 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -738,6 +740,10 @@ class AztecToolbar : FrameLayout, OnMenuItemClickListener {
     }
 
     private fun setupMediaToolbar() {
+        val mediaToolbarContainer : LinearLayout = findViewById(R.id.media_button_container)
```
mzorz: Can we make this `LinearLayout` a more generic `View` instead? Just in case we want to change it in the future, and as all we really care about for now is just to be able to make the view `VISIBLE` or `GONE` only

0nko:
The changes look 👍.

mzorz:
:shipit: then!

PR #656, opened 2018-04-23
[Aztec for RN] Hide the HTML button if source editor is not set

daniloercoli:
This PR hides the `HTML` button in `Toolbar` if the source editor was not set.

This is in required GB mobile side, since we don't need (yet?) the HTML editor bundled in Aztec.

Note: This PR also makes SourceViewEditText `open` to be used later in AztecRN if needed.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
@@ -58,6 +58,7 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
     private lateinit var layoutExpandedTranslateInEnd: Animation
     private lateinit var layoutExpandedTranslateOutStart: Animation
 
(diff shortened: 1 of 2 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -58,6 +58,7 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
     private lateinit var layoutExpandedTranslateInEnd: Animation
     private lateinit var layoutExpandedTranslateOutStart: Animation
 
+    private lateinit var HTMLButton: RippleToggleButton
```
0nko: This is just a nitpick, but we've been using lowercase for acronyms in variable names. So for consistency I'd name this `htmlButton`. Feel free to ignore though :).

daniloercoli: No nitpick at all, renamed it!

PR #661, opened 2018-04-27
Changes to <hr> tag handling

### LINKED ISSUE #660: [Aztec + GB] `class` attribute gets lost when putting and extracting content in Aztec
Note: this has been originally detected when handling Gutenberg content but is actually an HTML-handling problem.

### Expected
`class` attributes (and any other attribute) should be kept

### Observed
When entering content in Aztec through `fromHtml()` then taking it out again with `toHtml()` the `class` attributes are gone

### Reproduced
1. Set an instance of `Aztec` with the following content (you can try setting this in the example app):
```
<!-- wp:separator -->
<hr class="wp-block-separator" />
 <!-- /wp:separator -->
```
2. run the following code:
```
            var pureHtml = aztec.sourceEditor?.getPureHtml()
            aztec.visualEditor.fromHtml(pureHtml!!)
            var htmlobtained = aztec.visualEditor.toHtml(false)
```
3. Observe `htmlobtained` results in the following:
```
<!-- wp:separator -->
<hr/>
 <!-- /wp:separator -->
```
(see the `<hr/>` tag does not contain the `class` attribute anymore)

#### Tested
[Google Pixel] on [8.0] with [1.3.2]

mzorz:
### Fix
This PR fixes #660, in which `<hr class="something">` tags would miss the class attribute when re-constructed back from Aztec to HTML.

1. Added a new test named `rulerAttributes` in `Attributestest` set in d082e80. 
2. now attributes for `<hr>` are retained, addressed in 1f51c88
3. **IMPORTANT** reflecting changes to cope with the `<hr/>`  style, which is supposed to be both valid HTML and XHTML as it's an enclosing tag, addressed in 572d200.

For point 3 above, I've based my decision in this https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr and  https://stackoverflow.com/questions/6428184/can-a-hr-element-have-a-closing-tag.  Given I consider myself no expert in HTML, please chime in with any reflections you may have on this.

### Test
You can either run the included new test and verify it passes, or:

1. Set an instance of `Aztec` with the following content (you can try setting this in the example app):
```
<!-- wp:separator --><hr class="wp-block-separator" /><!-- /wp:separator -->
```
2. run the following code:
```
            var pureHtml = aztec.sourceEditor?.getPureHtml()
            aztec.visualEditor.fromHtml(pureHtml!!)
            var htmlobtained = aztec.visualEditor.toHtml(false)
```
3. Observe `htmlobtained` now results in the following, matching the original content:
```
<!-- wp:separator --><hr class="wp-block-separator" /><!-- /wp:separator -->
```

### Review
@daniloercoli  @0nko

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/HistoryMixedTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/HistoryMixedTests.kt
@@ -24,7 +24,7 @@ class HistoryMixedTests : BaseHistoryTest() {
             "\t<li></li>\n" +
             "</ol>\n" +
             "\n" +
-            "<hr>\n" +
+            "<hr/>\n" +
             "\n" +
             "<ul>\n" +
             "\t<li>Unordered</li>\n" +
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/HistoryMixedTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/HistoryMixedTests.kt
@@ -24,7 +24,7 @@ class HistoryMixedTests : BaseHistoryTest() {
             "\t<li></li>\n" +
             "</ol>\n" +
             "\n" +
-            "<hr/>\n" +
+            "<hr />\n" +
             "\n" +
             "<ul>\n" +
             "\t<li>Unordered</li>\n" +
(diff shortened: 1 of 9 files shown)


app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/HistoryMixedTests.kt
```diff
@@ -24,7 +24,7 @@ class HistoryMixedTests : BaseHistoryTest() {
             "\t<li></li>\n" +
             "</ol>\n" +
             "\n" +
-            "<hr>\n" +
+            "<hr/>\n" +
```
0nko: Technically there should be a space between "hr" and "/": `<hr />`. We never bothered with this since Calypso would always change it back to `<hr>`. The same thing would happen with `<br />`.

daniloercoli:
Once @0nko comments are addressed we can merge this.

app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/HistoryMixedTests.kt
mzorz: Thanks! addressed in d8c45fea

0nko:
:shipit:

PR #663, opened 2018-05-02
Issue/662 keep focus on copy

### LINKED ISSUE #662: Should not clear focus on text copy
### Expected
When copying text, the focus should stay with the editor.

### Observed
See https://github.com/wordpress-mobile/WordPress-Android/issues/7695

### Reproduced
See https://github.com/wordpress-mobile/WordPress-Android/issues/7695

#### Tested
Nexus 5X on Android v7.1.2 with Aztec v1.3.2, wpandroid vAlpha-108

hypest:
### Fix #662 

This PR revises what happens right after copying some text to the clipboard. Instead of clearing the focus, it tries to emulate what the parent component (EditText) does and just sets the cursor position to the end of the selection. That has the desired effect of dismissing the action menu.

### Test
1. Select some text
2. Tap on "Copy" on the popup action menu
3. Notice the cursor moving to the end of the selected text

### Review
@daniloercoli / @0nko


maxme:
Looks good, :shipit: - I'll publish a v1.3.3

PR #664, opened 2018-05-03
Issue/659 gutenberg fix preformatted closing delimiter

### LINKED ISSUE #659: [Aztec + GB] Properly handle new lines at end of blocks
Noticed that GB posts can be easily broken if you edit a post by adding new lines at the end of a block.
The web interface notice the error, and propose a fix, but we should make sure to remember the "GB closing comment" position when exiting a block. (2 new lines).

Steps to repro:
- Open a GB block with a `pre-formatted` block in it.
- Tap at the end of the block, and tap `enter` twice to exit the preformatted block.
- Add new content.
- Ended up with the following content:
```
<!-- wp:preformatted -->
<pre class="wp-block-preformatted">Preformatted block content</pre>
<br>
<br>
Hello There,
<br>
Look at the sky, it's falling!
<br>
<!-- /wp:preformatted --> <!-- wp:paragraph -->
<p></p>
<!-- /wp:paragraph -->
```

mzorz:
### Fix

This PR fixes #659 as it tries to re-accomodate content in between closing HTML `</pre>` and GB `<!-- /wp:preformatted -->` or HTML `</pre>` and GB `<!-- /wp:code -->` tags, shifting it below the Gutenberg block end delimiter to keep the `</pre>` tag and GB block end delimiter together.

See it in action (test steps below):

gutenenter

In order to cope with this:
- the mechanism is coded in the `HiddenGutenbergPlugin`, serving itself of the Aztec plugin architecture to encapsulate the Gutenberg logic outside of Aztec itself.
- I chose to only trigger the mechanism every time an _ending_ Gutenberg comment span is processed, so to minimize the amount of times this is checked for. Also, this is good as we know for sure the content passed to `handleSpanStart()` should have the newly added content right before the span (shortest path possible).
- while it would normally be possible to use `RegEx.replace()` once a matching expression is found, this method only returns a new `String` object, which can be memory consuming, so in this implementation we're operating on the passed `StringBuffer` itself, to avoid having to keep up with copies of content unnecessarily

### Test
1. Initialize Aztec with the following html content, and set Calypso mode off
```
<!-- wp:preformatted -->﻿<pre class="wp-block-preformatted">bla bla bla</pre><!-- /wp:preformatted -->
<!-- wp:code --><pre class="wp-block-code"><code> javascript code here </code></pre><!-- /wp:code -->
```
2. in the visual mode, place the cursor right aftr the end of the first block, right after `bla bla bla`.
3. tap enter once, and then tap enter again to exit the block
4. switch to HTML and observe the `<br>` was added _after_ the GB block end delimiter `<!-- /wp:preformatted -->` 
5. switch back to visual editor
6. position the cursor at the end of the code block, right after `javascript code here`.
7. Tap `enter` twice to exit the code block, and enter some content, optionally hitting `enter` as well
8. switch to HTML mode and observe the content has been added  _after_ the GB block end delimiter `<!-- /wp:code -->` 

### Review
@daniloercoli

### DIFF
--- a/wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/HiddenGutenbergPlugin.kt
+++ b/wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/HiddenGutenbergPlugin.kt
@@ -7,6 +7,8 @@ import org.wordpress.aztec.Constants
 import org.wordpress.aztec.plugins.html2visual.IHtmlCommentHandler
 import org.wordpress.aztec.plugins.visual2html.IInlineSpanHandler
 import org.wordpress.aztec.plugins.wpcomments.spans.GutenbergCommentSpan
+import java.util.regex.Matcher
+import java.util.regex.Pattern
 
 class HiddenGutenbergPlugin : IHtmlCommentHandler, IInlineSpanHandler {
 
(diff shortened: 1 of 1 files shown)


daniloercoli:
Tried this PR a bit and found out a couple of problems.
1. There is a white line at the beginning of the visual editor when I set the content to
`<!-- wp:preformatted -->﻿<pre class="wp-block-preformatted">bla bla bla</pre><!-- /wp:preformatted -->`
The cursor is also positioned on that line when I start the demo app.
2. There is a problem when I tried to erase content below (and outside) the block. If you keep tapping `del` you will end up deleting the GB closing comment `<!-- /wp:preformatted -->`
erasing-closing-tag
To replicate this issue make sure to delete all the `br` tags (actually all the content) below the pre-formatted tag.

If it's OK for you Mario we can chat a bit tomorrow, hoping there is a simpler way to preserve those opening and closing hidden tags.  cc @0nko since we may need some guidance on that.

mzorz:
Sure @daniloercoli if there's a different way that covers more of the problems we observed then I'm all for it - this PR only aims at keeping content between the ending tags out of the block end delimiter only, as described. All other cases are IMO exactly that: different cases.
@0nko feel free to chime in please :bow:

mzorz:
Closing this in favor of #666

PR #665, opened 2018-05-04
Keep track of GB HTML comments and try to preserve them

daniloercoli:
This is a HTML raw approach on preserving GB comments, attached to the correct HTML tag.

cc @mzorz @0nko


0nko:
This seems to work well. I tried an alternative approach with block spans: #666.

Issue #136, opened 2016-12-09
Improve span2html parsing logic for inline elements


khaykov:
Currently, this html `<b>bold <i>italic</i> bold</b>` after being parsed to span and back to html will become `<b>bold </b><b><i>italic</i></b><b> bold</b>`. This is not a bug, this is how Google originally implemented the parsing inside `Html.java`. It's based on `nextSpanTransition` method, which is simple and works just fine for Google, because the look of styled text does not change, and the resulting html is not really important for most people using the Html class. We care about html, so we need to change this at some point.

PR #666, opened 2018-05-05
Gutenberg block plugin fix

### LINKED ISSUE #659: [Aztec + GB] Properly handle new lines at end of blocks

0nko:
Fixes #659.

I was curious how difficult it would be to parse GB comments as a block. I had to add a new plugin type `IBlockSpanHandler` because we only had one for inline spans. 

There were a couple of issues along the way, like broken nesting because of extra `<html>` and `<body>` being inserted by the TagSoup parser. But it seems like this could be the right approach and the result looks promising.

gb_block

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/GutenbergCompatTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/GutenbergCompatTests.kt
@@ -1,0 +1,261 @@
+package org.wordpress.aztec.demo.tests
+
+import android.app.Activity
+import android.app.Instrumentation
+import android.content.Intent
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.net.Uri
+import android.support.test.espresso.intent.Intents
+import android.support.test.espresso.intent.matcher.IntentMatchers
+import android.support.test.espresso.intent.rule.IntentsTestRule
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.wordpress.aztec.AztecText
+import org.wordpress.aztec.demo.BaseTest
+import org.wordpress.aztec.demo.MainActivity
+import org.wordpress.aztec.demo.R
+import org.wordpress.aztec.demo.pages.EditorPage
+import org.wordpress.aztec.source.SourceViewEditText
+import java.io.File
+import java.io.FileOutputStream
+
+class GutenbergCompatTests : BaseTest() {
+
+    @Rule
+    @JvmField
+    val mActivityIntentsTestRule = IntentsTestRule<MainActivity>(MainActivity::class.java)
+
+    @Before
+    fun init() {
+        val aztecText = mActivityIntentsTestRule.activity.findViewById<AztecText>(R.id.aztec)
+        val sourceText = mActivityIntentsTestRule.activity.findViewById<SourceViewEditText>(R.id.source)
+
+        aztecText.setCalypsoMode(false)
+        sourceText.setCalypsoMode(false)
+    }
+
+    @Test
+    fun testRetainGutenbergPostContent() {
+        val html = "<!-- wp:paragraph --><p>Blue is not a color</p><!-- /wp:paragraph -->" +
+                "<!-- wp:list --><ul><li>item 1</li><li>item2</li></ul><!-- /wp:list -->" +
+                "<!-- wp:heading --><h2>H2</h2><!-- /wp:heading -->"
+
+        EditorPage().toggleHtml()
+                .insertHTML(html)
+                .toggleHtml()
+                .toggleHtml()
+                .verifyHTML(html)
+    }
+
+    @Test
+    fun testRetainGutenbergCommentsOnAddingText() {
+        val htmlStart = "<!-- wp:paragraph --><p>Blue is not</p><!-- /wp:paragraph -->"
+        val appendText = " a beautiful color"
+        val htmlVerify = "<!-- wp:paragraph --><p>Blue is not$appendText</p><!-- /wp:paragraph -->"
+
+        val editorPage = EditorPage()
+
+        // insert starter text
+        editorPage
+                .toggleHtml()
+                .insertHTML(htmlStart)
+                .toggleHtml()
+
+        // append text
+        editorPage
+                .setCursorPositionAtEnd()
+                .insertText(appendText)
+
+        // verify html
+        editorPage
+                .toggleHtml()
+                .verifyHTML(htmlVerify)
+    }
+
+    @Test
+    fun testAddPhotoAtBeginningOfGutenberParagraph() {
+        val htmlStart = "<!-- wp:paragraph --><p>"
+        val htmlEnd = "Blue is not a color</p><!-- /wp:paragraph -->"
+        val imagePlaceholder = "<img src=.+>"
+
+        val contentBeforeImageInsertion = htmlStart + htmlEnd
+        val contentAfterImageInsertion = htmlStart + imagePlaceholder + htmlEnd
+        val regex = Regex(contentAfterImageInsertion, RegexOption.DOT_MATCHES_ALL)
+        val editorPage = EditorPage()
+
+        // insert starter text
+        editorPage
+                .toggleHtml()
+                .insertHTML(contentBeforeImageInsertion)
+                .toggleHtml()
+
+        // insert image
+        editorPage
+                .moveCursorLeftAsManyTimes(contentBeforeImageInsertion.length)
+        createImageIntentFilter()
+        editorPage
+                .insertMedia()
+
+        // Must wait for simulated upload
+        Thread.sleep(10000)
+
+        editorPage
+                .toggleHtml()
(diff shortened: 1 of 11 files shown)


daniloercoli:
I tried this PR with a simple GB paragraph, and then added a picture to it from Aztec, ending up on having 2 pictures added to the editor.

Steps to repro:
- Open the demo app with the following HTML content:
```
        private val EXAMPLE = "<!-- wp:paragraph -->" +
                "<p>This is a paragraph</p>" +
                "<!-- /wp:paragraph -->"
```
- Put the cursor at the end of first line
- Add a picture to the editor

It added 2 pictures instead of one. 
The first picture is inside the `<!-- /wp:paragraph -->` GB comment.
screen shot 2018-05-07 at 18 22 24

mzorz:
Been able to repro the same issue as described in @daniloercoli 's [comment](https://github.com/wordpress-mobile/AztecEditor-Android/pull/666#issuecomment-387121262) here on different OS versions. Furthermore and to narrow down, this issue seems to happen _only when you add an image at the end of the **last** paragraph_. It doesn't fail like that when there’s more content and you try to insert at the end of the *first* paragraph. For example, using this code:
```
        var GBTEST =
                "<!-- wp:paragraph -->" +
                        "<p>This is a paragraph</p>" +
                        "<!-- /wp:paragraph -->" +
        "<!-- wp:paragraph -->" +
        "<p>This is a paragraph2</p>" +
        "<!-- /wp:paragraph -->"// +
```
if you place the cursor at the end of the first sentence and insert an image, checking the HTML editor shows this:

screen shot 2018-05-07 at 13 57 27

Then, inserting an image at the end of the _last_ (second) paragraph makes it duplicate the image when converting to HTML:

screen shot 2018-05-07 at 13 59 33

Just inserting an image at the end of the _last_ paragraph makes it duplicate.

Another thing observed is that placing the cursor there at the end is not remembered well when switching to/from HTML mode:
1. open the app with these 2 paragraphs as shown in this comment 
2. place the cursor at the end of the first paragraph
3. switch to HTML mode and observe the cursor is there at the end of the first paragraph
4. switch back to visual mode, and observe the cursor place is preserved
5. now place the cursor at the end of the second (last) sentence
6. switch to HTML mode: observe the placement of cursor is not preserved

gutencursor

mzorz:
### Explanation of the double image issue
[see description here above](https://github.com/wordpress-mobile/AztecEditor-Android/pull/666#issuecomment-387134347)
According to my tests this was happening as a span was being not set correctly because start and end spans for the GB comment and the included content need to start/end at the same nesting level. While the nesting level concept doesn't really exist in spans, it does matter that both the `<!-- wp:paragraph>` comment and the actual paragraph `<p>` are represented by `IAztecBlockSpan` types (namely both classes  `GutenbergCommentSpan` and `ParagraphSpan` are subclasses of `IAztecBlockSpan`), and we have no way of knowing at any given position where 2 spans are set, _which_ span is set _first_, as both are at the same position in the Spanned, but rather we have this information (nestingLevel) elsewhere. We were looking for  [`firstOrNull`](https://github.com/wordpress-mobile/AztecEditor-Android/compare/issue/659-gb-blocks...issue/659-gb-blocks-mzfix?expand=1#diff-3cbce8dda3659982ca9cdc966ce2563aL146) and thus the start/end spans were not coming in order when inserting an image.

### Fix
I managed to patch that by [looking into the nesting level and only considering a match when the same `nestingLevel` is reached as well](https://github.com/wordpress-mobile/AztecEditor-Android/compare/issue/659-gb-blocks...issue/659-gb-blocks-mzfix?expand=1), but even with that patch (and unrelated to it), there are other multiple problems that the span-based approach doesn't handle well (or better put, is not designed to handle), involving the inability to pair and tie 2 opening [HTML|comment] and 2 closing [HTML|comment] tags together, which is ultimately what is causing errors when converting back to HTML and opening the content in Gutenberg. Here are a couple of examples of things that are wrong and will be quite difficult to keep up with:

### Example 1
Putting the cursor at the beginning of the example text (if you switch to HTML you'll see the cursor is at the beginning of the text but after both the GB delimiter and the `<p>` opening tag) then inserting an image produces the following output HTML:

screen shot 2018-05-08 at 12 52 43

(the image is inserted _after_ the starting GB delimiter, but _before_  the opening `<p>` tag, thus making this paragraph block have 2 root tags which doesn't work well as of today.

### Example 2
Inserting an image in the last paragraph, then switching to HTML, then deleting all traces of that image, leaving the HTML code "as it was originally". Then back to visual mode and insert the image again, produces the following output:

screen shot 2018-05-08 at 13 05 17

As you can see, the GB delimiter is now gone.

### Thoughts
I think we should go with the "pre/post process and tie tags" approach Danilo made (https://github.com/wordpress-mobile/AztecEditor-Android/pull/665), here's why:
handling Gutenberg comments as blocks seems like a good idea but in order to prevent breaking the blocks as per how Gutenberg handles them today we also need to tie them to the opening/closing tags, for example there can't be anything in between the closing  `</p></we:paragraph>` tags, which is a specific case my first attempt was addressing successfully (https://github.com/wordpress-mobile/AztecEditor-Android/pull/664). By using spans only, we can't handle the possibility that the user places the cursor wherever they want and end up inserting stuff where they shouldn't.

Therefore, I think hiding the GB delimiters in the known last opening / closing HTML tag and then bloating them again once converting back from visual to html is IMO way safer.

hypest:
I noticed that the double-image issue mentioned by @daniloercoli (and @mzorz ) is actually an "issue" present on develop too.

What happens is that when adding the image at text buffer end, [the code](https://github.com/wordpress-mobile/AztecEditor-Android/blob/1f51c88fa9e6ff8aa3867854e81a06dc287f1383/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt#L139) appends the `` *outside* the `<p>`, but inserts it normally inside the `<p>` if the insertion position is not at the end of text/buffer.

I think that we need tackle this root cause first, by making sure the ` is always inserted *inside* the `<p></p>` since the buffer end position should be considered part of the `<p></p>` span.

daniloercoli:
> I noticed that the double-image issue mentioned by @daniloercoli (and @mzorz ) is actually an "issue" present on develop too.

That's weird, because today I wrote a test for this and there wasn't a problem running it on develop.

hypest:
Opened a [PR with a couple of fixes](https://github.com/wordpress-mobile/AztecEditor-Android/pull/668). Not sure if more issues are still existing though.

mzorz:
I've been performing tests integrating with WPAndroid  - we at least don’t seem to be breaking Gutenberg posts now, which is great! 🎉 

Interesting findings here:

1. there are Classic blocks appearing there where in Aztec I inserted new lines in the middle of an existing block, that was converted to `<p></p>` This is harmless IMO, and also something that seems to be new in Gutenberg web (didn’t observe this happening in previous versions of Gutenberg)

2. opening a Post in Aztec and then just leaving Aztec will produce slight variations - none of them break GB per se, but produce a new version of the Post that WPAndroid then uploads.

While this is not the best UX, I think it's "compatible enough" for now. A note on this, once the post goes through Aztec once, it doesn't seem to update itself unpurposedly if you open and edit on Gutenberg web afterwards (and open back in WPAndroid again). So I think this is OK to bear with.

Some of the differences found:
screen shot 2018-05-10 at 11 15 51

screen shot 2018-05-10 at 11 17 07

That is, a space is trimmed (first case), and a special character is added in the second case. This latter case might be worth looking into and further follow up, but I don't think (as per my tests) it's going to happen on every post, I'd say it's probably not the highest priority.

Attaching the files with content before /  after processing here so you can diff locally and see.

[post-afterAztec.html.txt](https://github.com/wordpress-mobile/AztecEditor-Android/files/1991977/post-afterAztec.html.txt)
[post-original.html.txt](https://github.com/wordpress-mobile/AztecEditor-Android/files/1991978/post-original.html.txt)

(renamed to `.txt` so I could attach them here)

Looking forward to merge if you agree @daniloercoli @hypest

daniloercoli:
Merging this one then :shipit: 

Mario, when you have the time would you mind to open new tickets for the issues you're reported above?

mzorz:
> 2. opening a Post in Aztec and then just leaving Aztec will produce slight variations - none of them break GB per se, but produce a new version of the Post that WPAndroid then uploads.

Was going to track this one but @daniloercoli found the mother issue already tracked in #136 🙇

PR #667, opened 2018-05-07
Google chromebook

kwonye:
### Fix

On Chromebooks ctrl + alt + T is the shortcut for the terminal. This removes the more keyboard shortcut from being a duplicate shortcut with that.

### Test
1. Open the WP app and go into the editor
2. Press ctrl + alt + T and the terminal should open
3. Verify the more tag isn't added to the post

### Review
@maxme

### DIFF
--- a/wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/toolbar/MoreToolbarButton.kt
+++ b/wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/toolbar/MoreToolbarButton.kt
@@ -7,6 +7,7 @@ import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import android.widget.ToggleButton
+import org.wordpress.android.util.DeviceUtils
 import org.wordpress.aztec.AztecText
 import org.wordpress.aztec.Constants
 import org.wordpress.aztec.plugins.IToolbarButton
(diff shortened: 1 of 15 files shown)


kwonye:
@maxme tests were not run by me so they're failing, sorry. Let me fix them and push another commit before you review.

maxme:
The lib doesn't build because the support libs were updated (via wputils update). And the new version of `ContextCompat.getDrawable` returns a nullable but our code expect a non null drawable.

See [here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/1f51c88fa9e6ff8aa3867854e81a06dc287f1383/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt#L57) and [there](https://github.com/wordpress-mobile/AztecEditor-Android/blob/1f51c88fa9e6ff8aa3867854e81a06dc287f1383/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt#L107). There are 2 options: you can check if the returned value is null recover (or load default) or change the field signature and let the lib crash if a null drawable is not expected.

wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/toolbar/MoreToolbarButton.kt
```diff
@@ -45,6 +46,10 @@ class MoreToolbarButton(val visualEditor: AztecText) : IToolbarButton {
     }
 
     override fun matchesKeyShortcut(keyCode: Int, event: KeyEvent): Boolean {
+        if (DeviceUtils.getInstance().isChromebook(context)) {
```
maxme: Should we add an alternative shortcuts for Chrome OS? This is usually what other apps do when there is a conflict with a specific OS or another app.

maxme: > Should we add an alternative shortcuts for Chrome OS? This is usually what other apps do when there is a conflict with a specific OS or another app.

maxme:
Lint issues fixed, travis-ci ✅, :shipit:

PR #670, opened 2018-05-09
overloaded AztecParser constructor for optional params

mzorz:
This PR just adds the Java overloads to `AztecParser` constructor so it can be called from Java without the optional parameters

cc @hypest @daniloercoli


daniloercoli:
:shipit:

PR #674, opened 2018-05-14
Add the ability to check if changes have been made to the content of the editor

daniloercoli:
This PR adds a utility method to `AztecText` that should be used to check wether the users have made changes to the content of the posts.

For example this html `<b>bold <i>italic</i> bold</b>` after being parsed to span and back to html will become `<b>bold </b><b><i>italic</i></b><b> bold</b>`. Reported in #136 - this is how Google originally implemented the parsing inside Html.java.
We've also seen similar problem with `img` tags.

This PR keeps a copy of the initial content set to the editor and use it later to check if changes have been made to the post content.
Tests are also available.

[wp-android details]
This PR tries to fix a bad UX where the post is uploaded to the server even if the user has not made any changes to it.

Not sure if I missed something, so better to check with @0nko

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/pages/EditorPage.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/pages/EditorPage.kt
@@ -14,9 +14,12 @@ import android.support.test.espresso.matcher.ViewMatchers.withId
 import android.support.test.espresso.matcher.ViewMatchers.withText
 import android.view.KeyEvent
 import android.view.View
+import org.hamcrest.Description
 import org.hamcrest.Matcher
 import org.hamcrest.Matchers.allOf
 import org.hamcrest.Matchers.hasToString
+import org.hamcrest.TypeSafeMatcher
+import org.wordpress.aztec.AztecText
 import org.wordpress.aztec.demo.Actions
 import org.wordpress.aztec.demo.BasePage
 import org.wordpress.aztec.demo.Matchers
@@ -364,6 +367,25 @@ class EditorPage : BasePage() {
         htmlEditor.check(matches(withText(expected)))
         label("Verified expected HTML editor contents without stripping")
 
+        return this
+    }
+
+    fun hasChanges(shouldHaveChanges : Boolean): EditorPage {
+        val hasNoChangesMatcher = object : TypeSafeMatcher<View>() {
+            override fun describeTo(description: Description) {
+                description.appendText("Aztec has no changes")
+            }
+
+            public override fun matchesSafely(view: View): Boolean {
+                if (view is AztecText) {
+                    return view.hasChanges() == shouldHaveChanges
+                }
+
+                return false
+            }
+        }
+
+        editor.check(matches(hasNoChangesMatcher))
         return this
     }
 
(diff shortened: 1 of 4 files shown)


mzorz:
We could also check whether there's a zero length change history and avoid further checking if that's the case, wdyt @daniloercoli?

daniloercoli:
> We could also check whether there's a zero length change history and avoid further checking if that's the case, wdyt @daniloercoli?

I wanted to keep it as simple as possible and tried to avoid checks on history, and other "edge" cases, since the performance drawback on checking if there are changes are minimal compared to all the other things the editor is doing in BG. Also the checks for `hasChanges` are executed on demand, when the host app call the method.

mzorz:
Another thing I'm thinking of is, instead of having `initialEditorContent` in memory all the time which could potentially lead to memory issues with big Posts, why don't we calculate a SHA256 digest on the content, keep that as `initialEditorContentSHA` and only compare that value to a new digest of `toPlainHtml()` in `hasChanges()`?

mzorz:
> instead of having initialEditorContent in memory all the time which could potentially lead to memory issues with big Posts, why don't we calculate a SHA256 digest on the content

@daniloercoli @0nko would you take a look into this?
https://github.com/wordpress-mobile/AztecEditor-Android/compare/feature/add-has-changes-check...feature/add-has-changes-check-sha256?expand=1

This other branch checks for the hash and compares this way, so we don't need to keep a copy of the initial content all the time. I'd revert this last commit as well https://github.com/wordpress-mobile/AztecEditor-Android/pull/674/commits/1cdf83ea34c160458404d9d2e2036e65f9d13c9f  and leave it up to the caller to just call `hasChanged()` and let them decide what is best for them, instead of us keeping a copy in memory (and disk) to circumvent this.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1053,6 +1066,39 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
+                initialEditorContentParsedSHA256 = calculateSHA256(initialHTMLParsed)
+            }
+        } catch (e: Throwable) {
+            // Do nothing here. `toPlainHtml` can throw exceptions, also calculateSHA256 -> NoSuchAlgorithmException
+        }
+    }
+
+    @Throws(NoSuchAlgorithmException::class)
+    private fun calculateSHA256(s: String): ByteArray {
+        val digest = java.security.MessageDigest.getInstance("SHA-256")
```
mzorz: let's make this just `MessageDigest.getInstance("SHA-256")` and add `import java.security.MessageDigest` at the beginning of the file

app/src/androidTest/kotlin/org/wordpress/aztec/demo/pages/EditorPage.kt
```diff
@@ -367,6 +370,25 @@ class EditorPage : BasePage() {
         return this
     }
 
+    fun hasChanges(shouldHaveChanges : AztecText.EditorHasChanges): EditorPage {
+        val hasNoChangesMatcher = object : TypeSafeMatcher<View>() {
+            override fun describeTo(description: Description) {
+                description.appendText("User has been made changes to the post: $shouldHaveChanges")
```
mzorz: super nit, it should be  `User has made changes to the post...` (remove the word "been")

PR #678, opened 2018-05-18
Fix for inline GB comments

### LINKED ISSUE #675: [Aztec + GB] self-closing GB blocks are removed when opened in Aztec
### Expected
Don't lose any GB block when opening a Gutenberg post in Aztec

### Observed
When a Post containing self-closing comments (GB blocks) such as 
```<!-- wp:latest-posts /-->```
these are removed entirely

### Reproduced
1. write a draft post on Gutenberg web and include a `latest posts` or `categories` block in it.
2. save the draft
3. open it on Aztec
4. switch to HTML, observe the comment is not there anymore

#### Tested
[Google Pixel] on [Android 8.0] with [1.3.3]

daniloercoli:
This PR fixes #675 by making sure inline Gutenberg comments are not eat by the GB block handler, and shows a `?` placeholder on Visual Editor.

We can ask designer to prepare a new drawable with some text in it, like `GB block`, and show it instead. (See how More and Page WordPress comments are handled visually in Aztec).

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/GutenbergCompatTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/GutenbergCompatTests.kt
@@ -51,6 +51,21 @@ class GutenbergCompatTests : BaseTest() {
     }
 
     @Test
+    fun testRetainGutenbergPostContentAndInlineGutenbergComment() {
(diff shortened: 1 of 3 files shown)


mzorz:
Good job, this is exactly the approach I had in mind. @iamthomasbishop may we ask what image resource do you think would be best to represent a generic Gutenberg block ?

PR #679, opened 2018-05-18
Fix tests and improve CSS styles merger

daniloercoli:
Some *CSS* related tests are failing in `develop` due to a problem with the CSS styles merger that does not take in consideration duplicates. 

The problems with tests were introduced in #674, where at the beginning of the initialization process we call `calculateInitialHTMLSHA` to keep an hash of the initial content of the editor.
The routine `calculateInitialHTMLSHA` internally calls `toPlainHtml` , and it seems that 2 consecutive calls to `toPlainHtml` does make the CSS merger in trouble since it doesn't not take in consideration duplicates. I think there is an underlying problem in the CSS plugin, but for now I fixed it by improving the CSS merger class making sure to address duplicate styles on merging.


mzorz:
:shipit:

PR #682, opened 2018-05-23
Fix - Add missing hasChanges method to HTML editor

daniloercoli:
We had add a way to detect if changes were made to the Visual Editor in https://github.com/wordpress-mobile/AztecEditor-Android/pull/674 but forgot to add the same checks to the HTML Editor. Resulting in hosting apps that are unaware if users made changes in the HTML editor or not.

To test this PR, open the demo app, makes some changes to the content in the HTL Editor and verify by calling `hasChanges` on it.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecInitialContentHolder.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecInitialContentHolder.kt
@@ -13,10 +13,12 @@ class AztecInitialContentHolder() : Parcelable {
     }
 
     constructor(parcel : Parcel) : this() {
+        initialEditorContentParsedSHA256 = ByteArray(parcel.readInt())
         parcel.readByteArray(initialEditorContentParsedSHA256)
     }
 
     override fun writeToParcel(dest: Parcel?, flags: Int) {
+        dest?.writeInt(initialEditorContentParsedSHA256.size)
         dest?.writeByteArray(initialEditorContentParsedSHA256)
     }
 
@@ -24,14 +26,16 @@ class AztecInitialContentHolder() : Parcelable {
         return 0
     }
 
-    @SuppressWarnings("unused")
-    val CREATOR: Parcelable.Creator<AztecInitialContentHolder> = object : Parcelable.Creator<AztecInitialContentHolder> {
-        override fun createFromParcel(`in`: Parcel): AztecInitialContentHolder {
-            return AztecInitialContentHolder(`in`)
-        }
+    companion object {
+        @JvmField
+        val CREATOR: Parcelable.Creator<AztecInitialContentHolder> = object : Parcelable.Creator<AztecInitialContentHolder> {
+            override fun createFromParcel(`in`: Parcel): AztecInitialContentHolder {
+                return AztecInitialContentHolder(`in`)
+            }
 
-        override fun newArray(size: Int): Array<AztecInitialContentHolder?> {
-            return arrayOfNulls(size)
+            override fun newArray(size: Int): Array<AztecInitialContentHolder?> {
+                return arrayOfNulls(size)
+            }
         }
     }
 
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecInitialContentHolder.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecInitialContentHolder.kt
@@ -12,6 +12,10 @@ class AztecInitialContentHolder() : Parcelable {
         CHANGES, NO_CHANGES, UNKNOWN
     }
 
+    interface EditorHasChangesInterface {
+        fun hasChanges(): AztecInitialContentHolder.EditorHasChanges
+    }
+
     constructor(parcel : Parcel) : this() {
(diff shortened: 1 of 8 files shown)


mzorz:
Nice catch on the need to observe the same `hasChanges()` behavior in the HTML editor too @daniloercoli 🙇  
Been looking into this quite a bit, the way the `testHasNoChangesOnHTMLEditor` test is structured made me think - the name suggests we're trying to verify there are no changes in HTML, but given we are positioning cursor at the end and then inserting more code, I think to myself  "we're actually changing it, why are we trying to test there are no changes then?", so I looked closer and found out something interesting.

There's a problem with the fact that `AztecText` and `SourceViewEditText` each keep their own separate "memory" hash to track if something has changed, so toggling between one and the other makes such memory only be initialized when editing has maybe already started in the other editor and vice versa.

I've tested this on WPAndroid with the other corresponding PR, and that's what's happening:
Situation A:
a) don't enter anything
b) switch to html mode
c) enter some text
d) switch back to visual mode
e) hit publish: nothing new to publish aka empty post body (but there should, as there's the text entered in step c)

Situation B:
a) enter some text in visual mode
b) switch to HTML
c) add some more
d) switch to visual
e) tap PUBLISH: nothing new in visual, so the Post that gets sent to the server has the content entered in point a) only.

There's also code duplication in both `AztecText` and `SourceViewEditText` for all the hash handling mechanism - I think we should make a new "initializer" component as a separate class, that can be instantiated in each editor, and also make sure to initialize both editors at the same time with the same content when the default editor mode comes up (i.e. if the VisualEditor is up first, also initialize the HTML editor instance of  this "initializer component" so it holds the equivalent initial content hash), what do you think?

aztec/src/main/kotlin/org/wordpress/aztec/source/SourceViewEditText.kt
```diff
@@ -184,6 +191,12 @@ open class SourceViewEditText : android.support.v7.widget.AppCompatEditText, Tex
     }
 
     fun displayStyledAndFormattedHtml(source: String) {
+        initialContentHolder?.let {
+            if (it.needToSetInitialValue()) {
+                it.setInitialContent(source)
+            }
+        }
+
         val styledHtml = styleHtml(Format.addSourceEditorFormatting(source, isInCalypsoMode))
```
mzorz: I found something interesting here. I think we should first run `Format.addSourceEditorFormatting` and save the output as the `initialContent`, to make sure we account for Calypso mode, something like this (haven't tested):

```
fun displayStyledAndFormattedHtml(source: String) {

        val formattedSource = Format.addSourceEditorFormatting(source, isInCalypsoMode)
        initialContentHolder?.let {
            if (it.needToSetInitialValue()) {
                it.setInitialContent(formattedSource)
            }
        }

        val styledHtml = styleHtml(formattedSource)
[...]
```

I'm saying this because I found an interesting situation testing on WPAndroid with the corresponding PR, as follows.

Write a post with this content on the web:
```
<p>Dzssgshdzx did</p>
<p>Cudjddgn</p>
```

Now on WPAndroid:
1. open the post
2. don't edit anything
3. switch to HTML
4. tap BACK
5. -> observe the `hasChanges` method on the source editor returns `true`

This is due to how we handle Calypso mode, we need to calculate the initial hash taking calypso mode into account.

Please note that, in WPAndroid the actual Post does not get updated to the server, because there's still some code in WPAndroid that prevents this from happening, but formally we should probably make sure that Calypso mode conversions are not confounded by Aztec itself as a "change".

Given the code in WPAndroid saves this from happening, I think we can probably live with this as is, but I thought it's worth mentioning, what do you think?

aztec/src/main/kotlin/org/wordpress/aztec/AztecInitialContentHolder.kt
```diff
@@ -0,0 +1,80 @@
...
+        dest?.writeByteArray(initialEditorContentParsedSHA256)
+    }
+
+    override fun describeContents(): Int {
+        return 0
+    }
+
+    companion object {
+        @JvmField
+        val CREATOR: Parcelable.Creator<AztecInitialContentHolder> = object : Parcelable.Creator<AztecInitialContentHolder> {
```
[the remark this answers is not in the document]
daniloercoli: Tried to suppress with with annotation, but no luck. There are other instances of it in Aztec without any annotation, so I left it without adding it.

aztec/src/main/kotlin/org/wordpress/aztec/source/SourceViewEditText.kt
daniloercoli: Thanks Mario for the testing effort and for the proposed solution. I tried with it with no luck. since the transformation to Calypso HTML, and Jsoup parsing is changing the content. 
Making  hard to get it back and compare it.

mzorz:
Just added the `[not ready for review]` label to keep this PR open until it makes sense to revisit cc @hypest @daniloercoli

daniloercoli:
Closing this since the problem was fixed in https://github.com/wordpress-mobile/AztecEditor-Android/pull/704

PR #684, opened 2018-05-28
Use TextWatcher to detect if changes are made to the editor

daniloercoli:
As per chat convo, this PR tries to check if changes were made to the editor by using TextWatcher.

Ref: https://github.com/wordpress-mobile/AztecEditor-Android/pull/682

I'm still investigating this option, just opening this PR to double check it's a viable solution.

Note 1: Already fixed an edge case with Pasting text in Aztec, but I guess there are other edge cases to consider.
Note 2: History is not considered at all. So, if you add text, and then Redo your changes, the editor is marked as changed. I hope we're not considering to fix this :)

cc @mzorz @hypest


daniloercoli:
Commenting here to have better tracking of problems we could have by just checking the History size to estimate the `hasChanges` variable.
- I started with a clean Aztec
- Inserted some text
- Did a couple of undo/redo
- Ended up on having a 0-length `undo` list, but NOT an empty editor,

screen shot 2018-05-28 at 15 23 46

daniloercoli:
We've fixed the problem by using hashes. See the 2 following PRs: #682 and #704.

I'd like to see us reconsidering the hash approach, that seems kind of hacky, in favour of a more cleaner TextWatcher implementation, like this PR does. 
Closing this for now.

PR #690, opened 2018-05-31
Instruct Aztec to produce valid HTML output

daniloercoli:
As per chat convo we've noticed that Aztec was producing in output valid XML document rather than valid HTML document.
As an example of the problem try the following valid HTML in the demo app
```
<!-- wp:audio {"id":40} -->
<figure class="wp-block-audio"><audio controls src="your-content.mp3"></audio></figure>
<!-- /wp:audio -->
```
Then swittch to HTML and you will see:

```
<!-- wp:audio {"id":40} -->
<figure class="wp-block-audio"><audio controls src="your-content.mp3" /></figure>
<!-- /wp:audio -->
```

The `Formatter` produces a slightly different version of the input document, where the `audio` tag is replaced by a self-closing audio tag. That's outside the HTML specs. 
(If you need the HTML ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio)

- In this PR we've set Jsoup to produce HTML when formatting HTML for the HTML Editor.
- We're also passing the output HTML to Jsoup to make sure the HTML produced by the editor is valid

**Note:** GB Tests need to be updated after these changes.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -53,6 +53,7 @@ import android.view.View
 import android.view.WindowManager
 import android.view.inputmethod.BaseInputConnection
 import android.widget.EditText
+import org.jsoup.Jsoup
 import org.wordpress.android.util.AppLog
 import org.wordpress.android.util.ImageUtils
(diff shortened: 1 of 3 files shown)


daniloercoli:
Closing this for now, since https://github.com/wordpress-mobile/AztecEditor-Android/issues/691

PR #692, opened 2018-05-31
adds a Regex to make sure audio tag gets both start and end tags

### LINKED ISSUE #691: Aztec should produce valid HTML document
### Expected
* Valid HTML document displayed in the HTML editor.
* valid HTML document returned by both the Visual Editor (`ztecText.kt` method `toHTML`) and and HTML editor (`SourceViewEditText.kt` method `getPureHtml`) 

### Observed
If you start the demo app with valid HTML document like the following

```
<!-- wp:audio {"id":40} -->
<figure class="wp-block-audio"><audio controls src="your-content.mp3"></audio></figure>
<!-- /wp:audio -->
```
Then switch to HTML and you will see:

```
<!-- wp:audio {"id":40} -->
<figure class="wp-block-audio"><audio controls src="your-content.mp3" /></figure>
<!-- /wp:audio -->
```
The `audio` tag is replaced by a self-closing audio tag. That's outside the HTML specs. 
 ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio

### Observed
There are basically 2 problems there, the function that formats the input to be displayed on the screen doesn't set Jsoup to produce HTML.
The 2 methods (one in the Visual editor and one in the HTML editor) that return the HTML are not invoking Jsoup, so there is the possibility, especially for the visual editor to get back unexpected HTML code. 

### Note
Since this would be a rather big change, we've agreed on fixing the problem with `audio` tag with a regexp (a PR is coming for it), and postpone the full fix to another PR with better testing.

For reference the first PR that should fixe it, once and for all, is available here: https://github.com/wordpress-mobile/AztecEditor-Android/pull/690

### LINKED ISSUE #694: Audio tag should have start and end tags
### Expected
The `<audio></audio>` tag should always contain an ending tag.

### Observed
If you start the demo app with valid HTML document like the following

```
<!-- wp:audio {"id":40} -->
<figure class="wp-block-audio"><audio controls src="your-content.mp3"></audio></figure>
<!-- /wp:audio -->
```
Then switch to HTML and you will see:

```
<!-- wp:audio {"id":40} -->
<figure class="wp-block-audio"><audio controls src="your-content.mp3" /></figure>
<!-- /wp:audio -->
```
The `audio` tag is replaced by a self-closing audio tag. That's outside the HTML specs. 
 ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio

#### Tested
[Google Pixel] on [8.0] with [1.3.4]

mzorz:
### Fix
Fixes #694  by introducing a regex in the `onHtmlProcessed` method of `AudioShortcodePlugin` to make sure we have a start and ending tags for `audio`.

As per discussed eslewhere, this should be the less disruptive change, and that's why I decided to encapsulate it in the plugin _and_ only for Gutenberg blocks handling, even when the problem is a broader HTML validation thing.

### Test
Test case has been provided in  https://github.com/wordpress-mobile/AztecEditor-Android/compare/issue/691-fix-audio-tag-through-plugin?expand=1#diff-36410cec871dab8b74e812d4f1efbbee 
You can also manually test by doing this:
1. Feed the demo app with
```
<!-- wp:audio {\"id\":435} --><figure class=\"wp-block-audio\"><audio controls=\"controls\" src=\"https://domain.com/content.mp3\"></audio><figcaption>a caption</figcaption></figure><!-- /wp:audio -->
```
2. switch to Visual
3. witch back to HTML mode, you should see the `</audio>` ending tag is still there.

### DIFF
--- a/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/AudioShortcodePlugin.kt
+++ b/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/AudioShortcodePlugin.kt
@@ -14,7 +14,13 @@ class AudioShortcodePlugin : IHtmlPreprocessor, IHtmlPostprocessor {
     }
 
     override fun onHtmlProcessed(source: String): String {
-        if (GutenbergUtils.contentContainsGutenbergBlocks(source)) return source
+        if (GutenbergUtils.contentContainsGutenbergBlocks(source)) {
+            // From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio
+            // > Tag omission	None, both the starting and ending tag are mandatory.
+            return StringBuilder(source)
+                    .replace(Regex("(<$TAG[^>]*?)(\\s*/>)"), "\$1></$TAG>")
+        }
+
         return StringBuilder(source)
                 .replace(Regex("<$TAG([^>]*(?<! )) */>"), "[$TAG$1]")
                 .replace(Regex("<$TAG([^>]*(?<! )) *></$TAG>"), "[$TAG$1]")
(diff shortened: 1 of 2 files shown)


daniloercoli:
Nice! :shipit:

PR #689, opened 2018-05-30
Fix audio tag handling for Gutenberg audio blocks

mzorz:
### Fix 
Fixes #685 as per the discussion in https://github.com/wordpress-mobile/WordPress-Android/pull/7826#issuecomment-393172951

The idea is that the plugin itself should be aware of Gutenberg blocks, and be able to differentiate and only apply itself to `<audio>` tags that are not within a Gutenberg block.

### Test
1. Feed Aztec with this HTML code, which is the Gutenberg code for an audio block:

```
<!-- wp:audio {"id":435} -->
<figure class="wp-block-audio"><audio controls src="http://www.noiseaddicts.com/samples_1w72b820/4159.mp3"></audio> <figcaption>
  a caption
 </figcaption></figure>
<!-- /wp:audio -->
```
or, create your own on Gutenberg.

2. Switch to visual mode
3. Now switch to HTML-mode
4. Observe the structure for the <audio> tag is preserved as <audio> (note the `controls="controls"` is expected and not harmful):
```
<!-- wp:audio {"id":435} --><figure class="wp-block-audio">
 <audio controls="controls" src="http://www.noiseaddicts.com/samples_1w72b820/4159.mp3"></audio>
 <figcaption>
  a caption
 </figcaption></figure><!-- /wp:audio -->
```

Also, test the same without the enclosing Gutenberg end/start tags:
```<figure class="wp-block-audio"><audio controls src="https://selfhostedmario.mystagingwebsite.com/wp-content/uploads/2018/05/ArgentinaAnthem.mp3"></audio><figcaption>a caption</figcaption></figure>```
In this case, the plugin should rightfully change the `<audio>` tags for `[audio]` shortcode, as expected.

**Note**: For easier review, please note the actual code change is in cfc512f, and tests have been added in  7178316, while instrumented tests have been added on 99d3efc.

PR #696, opened 2018-05-31
Fix video tag shortcode & start/end tags through plugin

### LINKED ISSUE #695: Video tag should contain both start and end tags
Similar to #694, but opening a new issue so each case is described on a block-type by block-type basis and is easier to track.

### Expected
The `<video></video>` tag should always contain an ending tag.

### Observed
If you start the demo app with valid HTML document like the following

```
<!-- wp:video {"id":25} -->
<figure class="wp-block-video"><video controls src="https://selfhostedmario.mystagingwebsite.com/wp-content/uploads/2017/05/VID-20170508-WA0011-1.mp4"></video></figure>
<!-- /wp:video -->
```
Then switch to HTML and you will see:

```
<!-- wp:video {"id":25} -->
<figure class="wp-block-video"><video controls src="https://selfhostedmario.mystagingwebsite.com/wp-content/uploads/2017/05/VID-20170508-WA0011-1.mp4" /></figure>
<!-- /wp:video -->
```
The `video` tag is replaced by a self-closing audio tag. That's outside the HTML specs. 
 ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

#### Tested
[Google Pixel] on [8.0] with [1.3.4]

mzorz:
### Fix
- Fixes #695 in a similar way to #692  by introducing a regex in the `onHtmlProcessed` method of `VideoShortcodePlugin` to make sure we have a start and ending tags for `video`.

- This PR also treats the `[video]` shortcode only on non-Gutenberg posts, in a similar way as #689 did for `[audio]` shortcode

- Also this PR has a fix in b7c2d7c5 as probably due to a copy/paste error the `VideoShortcodePlugin`  wasn't considering the case for start/end tags  to be replaced by the `[video]` shortcode. (bug [introduced here](https://github.com/wordpress-mobile/AztecEditor-Android/commit/22ba6fa37cb24d08f74ffbd14a32558a95a45630#diff-2e1ffe3f608cb361bfceb254d69e65b3) ).

### Test
Test cases for both shortcode handling and start/end tags preservation are provided in  https://github.com/wordpress-mobile/AztecEditor-Android/compare/issue/695-fix-video-tag-through-plugin?expand=1#diff-93c2d435bf9314a5dd124edac5ea7703
You can also manually test by doing this:
**TEST A**: keep the start/end tags on Gutenberg Video blocks
1. Feed the demo app with
```
<!-- wp:video {"id":25} --><figure class="wp-block-video"><video controls src="https://selfhostedmario.mystagingwebsite.com/wp-content/uploads/2017/05/VID-20170508-WA0011-1.mp4"></video></figure><!-- /wp:video -->
```
2. switch to Visual
3. witch back to HTML mode, you should see the `</video>` ending tag is still there.

**TEST B**: convert to shortcode
1. Feed the demo app with
```
<figure class="wp-block-video"><video controls src="https://selfhostedmario.mystagingwebsite.com/wp-content/uploads/2017/05/VID-20170508-WA0011-1.mp4"></video></figure>
```
2. switch to Visual
3. witch back to HTML mode, you should see the `<video>` tag converted to `[video]` shortcode format.

### DIFF
--- a/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/VideoShortcodePlugin.kt
+++ b/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/VideoShortcodePlugin.kt
@@ -28,7 +28,7 @@ class VideoShortcodePlugin : IHtmlPreprocessor, IHtmlPostprocessor {
 
         return StringBuilder(source)
                 .replace(Regex("<$TAG([^>]*(?<! )) */>"), { it -> fromHTMLToShortcode(it) })
-                .replace(Regex("<$TAG([^>]*(?<! )) */>"), { it -> fromHTMLToShortcode(it) })
+                .replace(Regex("<$TAG([^>]*(?<! )) *></$TAG>"), { it -> fromHTMLToShortcode(it) })
     }
 
     /**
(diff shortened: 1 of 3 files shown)


daniloercoli:
Looks good! 🐑 -it

PR #701, opened 2018-06-04
Don't steal focus on initial load of media

### LINKED ISSUE #700: Focus and cursor is lost when switching to visual mode
### Expected
Cursor stays at the input position when switching from html mode to visual mode.

### Observed
Cursor is not visible and focus is not actually in the editor at all.

### Reproduced
1. Run the demo app
2. Switch to html mode
3. Place the cursor at the end of the `Heading 6` text, right before the `</h6> tag
4. Switch back to visual mode and notice that the cursor is lost
5. Type anything on the virtual keyboard and notice that is written nowhere

#### Tested
Nexus 5X on Android v7.1.2 with Aztec v1.3.5

hypest:
### Fix #700 

As part of the work in https://github.com/wordpress-mobile/AztecEditor-Android/pull/349, a workaround was implemented to remove the focus from the editor's View. This PR specializes the workaround to exclude the case when initializing the editor, which happens everytime the user switches from html to visual.

### Test
1. Run the demo app
2. Switch to html mode
3. Place the cursor at the end of the `Heading 6` text, right before the `</h6> tag
4. Switch back to visual mode and notice the cursor blinking at the end of the `Heading 6` line

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -1025,7 +1025,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                 private fun replaceImage(drawable: Drawable?) {
                     it.drawable = drawable
                     post {
-                        refreshText()
+                        refreshText(false)
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


daniloercoli:
Should this call to [`refreshText()`](https://github.com/wordpress-mobile/AztecEditor-Android/blob/213f36e6cb3d9b6183ffc21ebe3cefd14b25d6de/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/extensions/VideoPressExtensions.kt#L27) be changed to `refreshText(false)` ?

hypest:
Good catch @daniloercoli , it does seem like we need to change that too. Will do. 🙇‍♂️

hypest:
Fixed with 9577c1f @daniloercoli , ready for another pass. Thanks!

daniloercoli:
Looks good!

PR #704, opened 2018-06-06
Detect change in html editor

### LINKED ISSUE #703: Detect content change in the source editor
Aztec can already detect user-initiated content change in the visual editor by employing a calculated hash-based mechanism. The source editor needs the same mechanism so a host app (like wpandroid) can query the html editor when active for whether it has changes.

hypest:
### Fix #703

This is a follow up PR to https://github.com/wordpress-mobile/AztecEditor-Android/pull/674, extending the idea of using a hash-based mechanism to detect when the html (source) editor has user-initiated changes in comparison to the original content.

This functionality is needed by wpandroid to properly update a post remotely when the editor is in html mode.

A key detail in this implementation is that now, the source editor's `displayStyledAndFormattedHtml()` method requires a boolean flag to be passed, effectively adopting the "content dirty" flag from the visual editor.

### Test
No test steps available.

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/Matchers.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/Matchers.kt
@@ -5,7 +5,9 @@ import android.widget.EditText
 import org.hamcrest.Description
 import org.hamcrest.Matcher
 import org.hamcrest.TypeSafeMatcher
+import org.wordpress.aztec.AztecText
 import org.wordpress.aztec.source.Format
+import org.wordpress.aztec.source.SourceViewEditText
 
 object Matchers {
     fun withRegex(expected: Regex): Matcher<View> {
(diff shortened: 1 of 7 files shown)


daniloercoli:
We should update current test cases for `hasChange` functionality, since they were already broken for the HTML editor, and it's a good time to include good tests in this PR.
I had proposed a fix for tests here https://github.com/wordpress-mobile/AztecEditor-Android/pull/682/files#diff-96db0ae2a3c3076d079b181b7512c885R275

daniloercoli:
Not quite sure I'm testing this PR in the right way, but starting the demo app, and switching to HTML editor results in `CHANGES` returned by the `hasChanges` method of `SourceViewEditText`.
Steps to repro:
- Start the demo app
- Switch to HTML
- Call the method `hasChanges` on the HTML editor.
- It returns `CHANGES` instead of NO_CHANGES.

hypest:
Hmm, I do see `getPureHtml(false)` producing different results at various stages in the process/steps @daniloercoli . Will continue to investigate.

hypest:
Made to fixes @daniloercoli :
1. There was a bug where `getPureHtml()` was actually mutating the editor's text just to insert and return the cursor position. Fixed with 16902ea.
2. The demo app was first initializing the source editor and then the visual editor. That way though, when switching back to html the source editor started detecting changes because the new text has different hash than the original one. I changed the demo app to prime the visual editor first, so source editor properly keeps the init hash from the text that includes the visual editor's mutations. Fixed with 3784dcd.

daniloercoli:
Cool, thanks @mzorz for your review. Going to merge this then, and let @hypest open another PR if he/we decide to implement utility methods you've mentioned in the other PR.

PR #707, opened 2018-06-07
Do not crash if Attributes->`setValue` and `removeAttribute` fail

### LINKED ISSUE #705: OOB Crash in AztecAttributes  setValue/removeAttribute
```
Fatal Exception: java.lang.ArrayIndexOutOfBoundsException: length=5; index=7
       at org.xml.sax.helpers.AttributesImpl.addAttribute(AttributesImpl.java:385)
       at org.wordpress.aztec.AztecAttributes.setValue(AztecAttributes.kt:10)
       at org.wordpress.aztec.source.CssStyleFormatter$Companion.addStyleAttribute(CssStyleFormatter.kt:123)
       at org.wordpress.aztec.AztecParser.withinNestable(AztecParser.kt:386)
       at org.wordpress.aztec.AztecParser.withinHtml(AztecParser.kt:359)
       at org.wordpress.aztec.AztecParser.withinHtml(AztecParser.kt:311)
       at org.wordpress.aztec.AztecParser.toHtml(AztecParser.kt:92)
       at org.wordpress.aztec.AztecText.toPlainHtml(AztecText.kt:1142)
       at org.wordpress.aztec.AztecText.toHtml(AztecText.kt:1105)
       at org.wordpress.aztec.AztecText.toHtml$default(AztecText.kt:1104)
       at org.wordpress.aztec.AztecText.toFormattedHtml(AztecText.kt:1146)
       at org.wordpress.aztec.History.beforeTextChanged(History.kt:42)
       at org.wordpress.aztec.AztecText$addHistoryLoggingWatcher$historyLoggingWatcher$1.beforeTextChanged(AztecText.kt:480)
       at android.widget.TextView.sendBeforeTextChanged(TextView.java:9144)
       at android.widget.TextView.access$1600(TextView.java:326)
       at android.widget.TextView$ChangeWatcher.beforeTextChanged(TextView.java:11954)
       at android.text.SpannableStringBuilder.sendBeforeTextChanged(SpannableStringBuilder.java:1027)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:523)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:494)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:34)
       at android.view.inputmethod.BaseInputConnection.replaceText(BaseInputConnection.java:691)
       at android.view.inputmethod.BaseInputConnection.setComposingText(BaseInputConnection.java:447)
       at com.android.internal.view.IInputConnectionWrapper.executeMessage(IInputConnectionWrapper.java:340)
       at com.android.internal.view.IInputConnectionWrapper$MyHandler.handleMessage(IInputConnectionWrapper.java:78)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:158)
       at android.app.ActivityThread.main(ActivityThread.java:7230)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
```

Internal Ref: 5a8423408cb3c2fa6302badc

daniloercoli:
Fix #705 by making sure the app doesn't crash on removing or set a value for un unexpected attribute.

Wasn't able to reproduce the issue, but it's pretty high on the list of crashes, so I just wrapped it with a try/catch block and log the problem.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecAttributes.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecAttributes.kt
@@ -23,13 +23,13 @@ class AztecAttributes(attributes: Attributes = AttributesImpl()) : AttributesImp
     }
 
     private fun logInternalState() {
-        AppLog.e(AppLog.T.EDITOR, "AttributesImpl has an internal length of $length")
-        // Since we're not sure the internal state of the Obj is correct we're wrapping toString in a try/catch
+        AppLog.e(AppLog.T.EDITOR, "Dumping internal state:")
+        AppLog.e(AppLog.T.EDITOR, "length = $length")
+        // Since the toString can throw OOB error we need to wrap it in a try/catch
         try {
-            AppLog.e(AppLog.T.EDITOR, "Dumping internal state:")
             AppLog.e(AppLog.T.EDITOR, toString())
-        } catch (t: Throwable) {
-            AppLog.e(AppLog.T.EDITOR, "Error dumping internal state!")
+        } catch (e: ArrayIndexOutOfBoundsException) {
+            AppLog.e(AppLog.T.EDITOR, "Error dumping internal state!", e)
         }
     }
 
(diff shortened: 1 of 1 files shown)


hypest:
👋 @daniloercoli , had a look at the PR and the only thing that's not clear to is how exactly do we decide this crash is low-importance enough that is OK to guard for it without rethrowing? Can you share what's on your mind regarding this? Thanks!

daniloercoli:
If we re-throw the error the app keep crashing for a problem in the CssStyleFormatter.
It also seems the problem is inside the platform, since it first returns that the attribute is available and then fails to set it.
I wasn't able to reproduce the problem with test documents, so decided to catch and log the problem, avoid the app crashing. We maybe want to log to external logger (in our case `wp-android` app log & crashlytics), for further investigation.

hypest:
Aha, I see, thanks for expanding @daniloercoli . My impression is that if we just "swallow" the exception we open ourselves up for potential content loss if user uses `undo`. The history will contain entries that are not complete (because of the exception and guard). I think that as it is, we can't recover from the error so, I'd probably suggest we leave the crash in until we have a more safe solution.

In the meantime, perhaps we can try to log (if possible) the Attributes object for inspection and debugging.

WDYT?

daniloercoli:
Ready for another pass!

PR #708, opened 2018-06-08
Proper handling of email, schemeless URLs

### LINKED ISSUE #460: Hyperlinking functionality inserting site address unnecessarily
### Expected

Adding a hyperlink without adding http:// would work.

### Observed

If a user adds a hyperlink without including http://, it adds the site URL of the site they are editing into the link.

Post with more information from the user: http://www.73ghia.com/?p=2000

### Reproduced

1. Edit a post
2. Go to add a link and add www.the370z.com
3. Save/insert and you will see the link becomes yoursiteurl.com/www.the370z.com

It only works if you add http:// at the beginning

#### Tested

SM-G930V on 7.0 with 7.9 (user says they are using beta).

hypest:
### Fix #460

This PR mirrors the [behavior from Calypso](https://github.com/Automattic/wp-calypso/blob/ab1c7779fb25d96220ec03dd9557d6fc89bbfd11/client/post-editor/editor-html-toolbar/add-link-dialog.jsx#L51-L60) to handle email links and links that need the scheme added.

### Test 1
1. Open the demo app and tap on the toolbar to insert a link
2. In the URL field type: `name@company.com`, copy the same to the Link text field and tap OK
3. Switch to html and notice that the `href` points to: `mailto:name@company.com` (prepended with `mailto:`)

### Test 2
1. Open the demo app and tap on the toolbar to insert a link
2. In the URL field type: `#internalanchor`, copy the same to the Link text field and tap OK
3. Switch to html and notice that the `href` points to: `#internalanchor` (no `http://` added)

### Test 3
1. Open the demo app and tap on the toolbar to insert a link
2. In the URL field type: `wordpress.org`, copy the same to the Link text field and tap OK
3. Switch to html and notice that the `href` points to: `http://wordpress.org` (there's an added `http://`)

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -162,7 +162,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         }
     }
 
-    private val REGEXP_EMAIL = Regex("^[A-Z0-9._%+-]+@[A-Z0-9.-]+.[A-Z]{2,20}$",
+    private val REGEXP_EMAIL = Regex("^[A-Z0-9._%+-]+@[A-Z0-9.-]+.[A-Z]{2,}$",
             setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.IGNORE_CASE))
     private val REGEXP_STANDALONE_URL = Regex("^(?:[a-z]+:|#|\\?|\\.|/)", RegexOption.DOT_MATCHES_ALL)
 
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -162,6 +162,10 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         }
     }
 
+    private val REGEXP_EMAIL = Regex("^[A-Z0-9._%+-]+@[A-Z0-9.-]+.[A-Z]{2,4}$",
+            setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.IGNORE_CASE))
+    private val REGEXP_STANDALONE_URL = Regex("^(?:[a-z]+:|#|\\?|\\.|/)", RegexOption.DOT_MATCHES_ALL)
+
     enum class EditorHasChanges {
         CHANGES, NO_CHANGES, UNKNOWN
     }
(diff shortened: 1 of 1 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -162,6 +162,10 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         }
     }
 
+    private val REGEXP_EMAIL = Regex("^[A-Z0-9._%+-]+@[A-Z0-9.-]+.[A-Z]{2,4}$",
```
koke: Top level domains can be longer than 4 characters now (e.g. `.design`)

hypest: Good point @koke . Any recommendation for how long to change it to? Some cursory googling has produced info for TLDs that can be evn 24chars long. Localized TLDs in particular start with `XN--` which is 4char long to begin with!

I think I will change it to 20chars.

daniloercoli: It seems that there is no set length limit for TLD. Checking on google I found think we can change it to 
`"^[A-Z0-9._%+-]+@[A-Z0-9.-]+.[A-Z]{2,}$"`

daniloercoli:
We should also escape the double quote character (`"` ) when inserted in links, otherwise we're breaking the resulting HTML link.
screen shot 2018-06-11 at 16 38 03

daniloercoli:
For reference pasting here the link to [email tests](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/forms/resources/ValidityState-typeMismatch-email.js?sq=package:chromium&type=cs) taken from chromium.

I've also found an interesting post on [stackoverflow](https://stackoverflow.com/a/46181) that could help with Regexps if we decide to investigate this further.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
hypest: Thanks @daniloercoli , lifted the upper limit with 3e9f744.

hypest:
> We should also escape the double quote character (" ) when inserted in links, otherwise we're breaking the resulting HTML link.

I added code (with 19c40b3) to encode the link text @daniloercoli . Let me know if that's what you had in mind.

By the way, as it currently is, the email regex won't match when there's a doublequote in the link text. How can I reproduce the case in the screenshot you shared above?

> For reference pasting here the link to email tests taken from chromium.
>
> I've also found an interesting post on stackoverflow that could help with Regexps if we decide to investigate this further.

Aha, thanks for looking into this! I feel though that an extensive support like that i outside the scope if this PR. Can we do it in a separate PR? If OK with you, can you open a ticket with the support you'd like to see implemented? Thanks!

daniloercoli:
> Aha, thanks for looking into this! I feel though that an extensive support like that i outside the scope if this PR. Can we do it in a separate PR? If OK with you, can you open a ticket with the support you'd like to see implemented? Thanks!

I agree and will open a new ticket.

daniloercoli:
> By the way, as it currently is, the email regex won't match when there's a doublequote in the link text. How can I reproduce the case in the screenshot you shared above?

If you try by inserting `dan"@text.com`  you will see there are 2 errors:
- The `http://` prefix is added to it
- The double quote is encoded fine. if you switch to HTML -> Visual -> HTML it's then lost (switched it back to plain character, I guess this is the problem with HTML ents we already have, so let's not consider it here for now).

daniloercoli:
:shipit:

hypest:
Thanks for the merge and for the new ticket @daniloercoli . We can move any regex discussion there I guess, cool?

PR #710, opened 2018-06-09
First class support for strong

### LINKED ISSUE #709: Don't transform <strong> to <b>
### Expected
When loading html that includes `<strong>` Aztec should not just render it as bold but serialize it as `<strong>` as well.

### Observed
Aztec transforms `<strong>` to <b>.

### Reproduced
1. Load `<strong>this is strong</strong>` into the demo app in html mode
2. Switch to visual mode
3. Switch to html mode and notice that the `strong`s have been transformed to `b`s

#### Tested
Nexus 5X on Android 7.1.2 with Aztec v1.3.5

hypest:
### Fix #709 

This PR:
1. Allows the `<strong>` element to be preserved when changing from html mode to visual and back
2. Makes the `<strong>` HTML element the "styling" element used when tapping on the "B" toolbar button

This implementation will probably help us also include similar support for the `em` element.

Implementation details:
* `ToolbarAction` now includes a set of formats it supports. What that means is that, for example, the corresponding toolbar button will get activated whenever any of those formats are detected. But, only the first format is considered "primary" and is the one applied when user taps the toolbar button to insert the format.
* The formats in the list above are considered "same". That is, the `isSameInlineSpanType` method will return `true` for them. That's useful for the cases we need to expand one into the other. For example, the [testExpandBoldAndSetToStrongFormatting](https://github.com/wordpress-mobile/AztecEditor-Android/blob/c4367d722f698e794ca4516a59203ed2544be5b2/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/ToolbarHighlightingTests.kt#L109) where `bold` is already in the text but gets transformed into `strong` when the style is applied to the whole text.

### Test
1. Load `<strong>this is strong</strong> and <b>this is bold</b>` into the demo app in html mode
2. Switch to visual mode
3. Switch back to html mode and notice that the `strong`s are still there, and same goes for `b`s

### Review
@mzorz or @daniloercoli 🙇‍♂️


daniloercoli:
This PR looks good. It seems thought that there are changes in it that regard the email fixes introduced by your other PR.
Will merge this once that code is remove from here.

hypest:
Thank you for having a look @daniloercoli and 🤦‍♂️, I pushed redundant commits, sorry. I rebased onto `develop` and force pushed now. Ready for another pass, thanks!

PR #713, opened 2018-06-12
Sync the editors on switch if change zeros out original change

### LINKED ISSUE #711: Switch to html misses text change if the change bring the content back to origin
### Expected
Removing changes and switching to html mode should mirror the new state of the content.

### Observed
The text in html mode is still the one that includes changes.

### Reproduced
1. Run the demo app and add some characters at the beginning of text
2. Switch to html mode
3. Switch back to visual
4. Remove/delete the characters added in step 1
5. Switch to html mode and observe the text still having the chars from step 1

This seems to be a regression introduced with https://github.com/wordpress-mobile/AztecEditor-Android/pull/704

#### Tested
Nexus 5X on Android 7.1.2 with app version efd43ed795f4c321590f1f7ff720eb1c4d4390a7

hypest:
### Fix #711 

This PR add another check for when switching between visual and html mode to have the editors sync their content between them if an editor has been modified since the last time a switch happened. That way, a change that effectively cancels out any change compared to the originally loaded content will be mirrored to the other editor as well.

### Test
1. Run the demo app and add some characters at the beginning of text
2. Switch to html mode
3. Switch back to visual
4. Remove/delete the characters added in step 1
5. Switch to html mode and observe the chars from Step 1 having been removed from the html mode too

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
@@ -32,6 +32,7 @@ import org.wordpress.aztec.R
 import org.wordpress.aztec.plugins.IMediaToolbarButton
 import org.wordpress.aztec.plugins.IToolbarButton
 import org.wordpress.aztec.source.SourceViewEditText
+import java.util.Arrays
 import java.util.ArrayList
 import java.util.Locale
 
@@ -41,6 +42,9 @@ import java.util.Locale
  * Supports RTL layout direction on API 19+
  */
 class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
+    val RETAINED_EDITOR_HTML_PARSED_SHA256_KEY = "RETAINED_EDITOR_HTML_PARSED_SHA256_KEY"
+    val RETAINED_SOURCE_HTML_PARSED_SHA256_KEY = "RETAINED_SOURCE_HTML_PARSED_SHA256_KEY"
+
     private var aztecToolbarListener: IAztecToolbarClickListener? = null
     private var editor: AztecText? = null
     private var headingMenu: PopupMenu? = null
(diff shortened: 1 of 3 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -561,22 +572,37 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
         }
     }
 
+    private fun syncEditorFromSource() {
```
daniloercoli: It seems to me that this method reads the content from the Visual editor, and in case of changes put the content into the HTML editor. 
If so, should we rename this method in `syncSourceEditorFromVisual`?

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -561,22 +572,37 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
...
+        val editorHtml = editor!!.toPlainHtml(true)
+        val sha256 = AztecText.calculateSHA256(editorHtml)
+        if (editor!!.hasChanges() != NO_CHANGES || !Arrays.equals(editorContentParsedSHA256LastSwitch, sha256)) {
+            sourceEditor!!.displayStyledAndFormattedHtml(editorHtml)
+        }
+        editorContentParsedSHA256LastSwitch = sha256
+    }
+
+    private fun syncSourceFromEditor() {
```
daniloercoli: Same comment of above, should we rename to `synchVisualEditorFromSourceEditor`?

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
hypest: Indeed, binary is hard! Will swap the names, good eyes!

hypest:
I've exchanged the method names with a973449 @daniloercoli , ready for another pass, thanks!

daniloercoli:
Cool! :shipit:

PR #715, opened 2018-06-13
Aztec may break css styles, and GB cover image blocks

### LINKED ISSUE #714: Aztec breaks CSS style on `background-image` with URL
- Load the following content into the demo app:
```
        private val EXAMPLE = "<div class=\"wp-block-cover-image has-background-dim alignwide\" style=\"background-image:url('https://cldup.com/Fz-ASbo2s3.jpg');\">" +
                "<p>Test</p>" +
                "</div>"
```

- Start the demo app
- Switch to HTML and see the style attribute:
`url(https: //cldup.com/Fz-ASbo2s3.jpg)` it has a space after `https`

This is due to the way we re-build the style, putting space after each `:` for readability, without checking if it's inside a url. (same happens on `;` on separating directives).

This also breaks GB blocks.

I think we can simply remove the space after `:` without going crazy checking if it needs or not a space after.

daniloercoli:
### Fix #714

Before this PR we were adding a space after a `:` in styles, for readability reason, without checking if the `:` char was used in URLs instead. In this PR i've just removed the code that adds the space, without doing regexp or other kind of hacks that check if the space was inside an URL or not.

Also updated CSS tests, and added a new test for GB cover block (that was the original reporter of the problem).

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/CssUnderlineFormattingTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/CssUnderlineFormattingTests.kt
@@ -27,7 +27,7 @@ class CssUnderlineFormattingTests : BaseTest() {
     fun testSimpleCssUnderlineFormatting() {
         val text1 = "some"
         val text2 = "text"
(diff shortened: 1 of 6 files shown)


mzorz:
Tested, works, checked code and commit history. LGTM :shipit:

PR #719, opened 2018-06-19
Issue/718 update on first mode switch

### LINKED ISSUE #718: Aztec updates source editor on first mode switch even if no user change
### Expected
First mode switch should not cause a sync of the editors contents if no user initiated change has been performed in the meantime.

### Observed
First time switching the editing mode causes the source editor to update its contents from the visual editor even if the user has made no changes to the visual editor's content.

This is problematic since the visual editor may have self-inflicted changes and those will migrate to the source editor and manifesting as changed content when the source editor is asked whether its contents are changed.

### Reproduced
1. Load the demo app
2. Attach the debugger and place a breakpoint at [this line](https://github.com/wordpress-mobile/AztecEditor-Android/blob/a9734495b7d7a456d9dca15536b111dfd46790ce/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt#L579), where the source editor is instructed to load new content
3. Back to the demo app, tap on the mode switch button on the toolbar and observe that the debugger stops at the breakpoint set in step 2.

#### Tested
Nexus 5X, Android v7.1.2, Aztec version v1.3.6

hypest:
### Fix #718 

This PR makes sure that the internal sha256 values the AztecToolbar is keeping to monitor whether an editor has changed content since the last time it got in the foreground are initialized right before the first time needed to be used.

The idea there goes like this: The first time switching to the source/html editor, we need to check if the _visual_ editor has changes compared to the original input content _or_ if it has changes since the last time the visual editor got in the foreground. But, since the visual editor was already in the foreground when starting, it makes sense to initialize the sha256 var with the current contents instead of checking against the empty var, as it was happening before this PR.

### Test
1. Load the demo app
2. Attach the debugger and place a breakpoint at [this line](https://github.com/wordpress-mobile/AztecEditor-Android/blob/b6893bc8a2beb15f0027f3499ff47bda98c29983/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt#L585), where the source editor is instructed to load new content
3. Back to the demo app, tap on the mode switch button on the toolbar and observe that the debugger doesn't stop at the breakpoint set in step 2

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -418,7 +418,9 @@ open class MainActivity : AppCompatActivity(),
         }
 
         if (savedInstanceState == null) {
-            aztec.visualEditor.fromHtml(EXAMPLE)
+            if (!isRunningTest) {
+                aztec.visualEditor.fromHtml(EXAMPLE)
+            }
             aztec.initSourceEditorHistory()
         }
 
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -418,7 +418,7 @@ open class MainActivity : AppCompatActivity(),
         }
 
         if (savedInstanceState == null) {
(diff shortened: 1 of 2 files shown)


daniloercoli:
Looks Good! :shipit:

PR #727, opened 2018-06-20
Treat div, figure, figcaption and section as block level html elements

### LINKED ISSUE #724: Prepending text in a GB image block caption breaks GB editing
### Expected
Adding text in front of the caption in a GB image block should not break the GB editing experience (`This block appears to have been modified externally.` placeholder).

### Observed
Adding text in front of the caption (`<figcaption>`) in a GB image block causes the `This block appears to have been modified externally.` placeholder to appear in GB, breaking the editing flow.

### Reproduced
1. Load the following html snippet in html mode:
```html
<!-- wp:image {"align":"center"} -->
<figure class="wp-block-image aligncenter">
  <img src="https://cldup.com/cXyG__fTLN.jpg" alt="Beautiful landscape" />
  <figcaption>If your theme supports it, you'll see the "wide" button on the image toolbar. Give it a try.
  </figcaption>
</figure>
<!-- /wp:image -->
```
2. Switch to visual mode and place the cursor in the very beginning of the caption (there might be a couple of spaces there so, make sure the cursor is at the start of the line)
3. Insert one or more characters
4. Switch to html mode and notice that the character(s) entered are inbetween the `` and the `<figcaption>` elements. That is not valid for GB and causes the "modified externally" placeholder to appear.

#### Tested
Nexus 5X on Android v7.1.2, Aztec v1.3.6

hypest:
### Fix #724 

This PR specializes the support for the `div`, `figure`, `figcaption` and `section` html elements to be compatible with block-level elements rather than inline-level elements that were implemented as so far. Besides, the HTML spec has them as block-level.

With this fix, placing the cursor at the beginning or the end of the respective span will make any text be included in the span. For example, adding text anywhere inside or the edges of the `figcaption` will have the text included in the caption as normal.

### Test
1. Load the following html snippet in html mode:
```html
<!-- wp:image {"align":"center"} -->
<figure class="wp-block-image aligncenter">
  <img src="https://cldup.com/cXyG__fTLN.jpg" alt="Beautiful landscape" />
  <figcaption>If your theme supports it, you'll see the "wide" button on the image toolbar. Give it a try.
  </figcaption>
</figure>
<!-- /wp:image -->
```
2. Switch to visual mode and place the cursor in the very beginning of the caption (there might be a couple of spaces there so, make sure the cursor is at the start of the line), or the very end
3. Insert one or more characters
4. Switch to html mode and notice that the character(s) entered lie inside the `<figcaption></figcaption>` element

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -113,7 +113,9 @@ open class MainActivity : AppCompatActivity(),
                         "</div>" +
                         "<br>"
         private val GUTENBERG_CODE_BLOCK = "<!-- wp:core/image {\"id\":316} -->\n" +
-                "<figure class=\"wp-block-image\"><img src=\"https://upload.wikimedia.org/wikipedia/commons/thumb/9/98/WordPress_blue_logo.svg/1200px-WordPress_blue_logo.svg.png\" alt=\"\" /></figure>\n" +
+                "<figure class=\"wp-block-image\"><img src=\"https://upload.wikimedia.org/wikipedia/commons/thumb/9/98/WordPress_blue_logo.svg/1200px-WordPress_blue_logo.svg.png\" alt=\"\" />\n" +
+                "  <figcaption>The WordPress logo!</figcaption>\n" +
+                "</figure>\n" +
                 "<!-- /wp:core/image -->"
         private val PREFORMAT =
                 "<pre>" +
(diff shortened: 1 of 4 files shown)


daniloercoli:
This PR looks good, and can be merged. :shipit:

Although I found a minor glitch where the cursor can be placed half-behind the picture just pressing `<-` array on a keyboard or tapping . See the video linked here: https://cloudup.com/cQMdUJygYbv

Will check if it was already reported, and create a new ticket in case it was not.

hypest:
Thanks for the review and merge @daniloercoli !

> a minor glitch

I think that's an issue with the `img` and the visual lines we artificially add for visualization purposes. Not directly related to the `figcaption` per se but yeah, in visual mode and to the casual user that space over there is supposed to be part of the caption.

daniloercoli:
Actually it was not even an issue :grin

The cursor was properly placed beside the picture, but it had the same color of the picture and it seemed placed behind it.
There is a small margin applied to pictures (only at the bottom) that results in that visual artifacts. The picture below explain this better than many words:
device-2018-06-21-130735

PR #743, opened 2018-09-05
Adds onTextContextMenuItemListener

mzorz:
Adds an `onTextContextMenuItemListener` that allows to signal a listener outside Aztec that `onCut`, `onCopy`, `onPaste` and `onPasteAsPlainText` have just happened

### Test
1. Open the demo app
2. Now switch to another app that allows copying content, for example open Chrome and navigate to a page that allows copying formatted content.
3. Select some content and choose `COPY` from the context menu
4. switch back to the Aztec demo app
5. long tap on the content area and choose `PASTE`
6. you should see the pasted content and a toast reading "on Paste!"

The same should happen for the remaining operations cut, copy, paste and paste as plain text.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -72,6 +72,7 @@ open class MainActivity : AppCompatActivity(),
         AztecText.OnAudioTappedListener,
         AztecText.OnMediaDeletedListener,
         AztecText.OnVideoInfoRequestedListener,
+        AztecText.OnTextContextMenuItemListener,
         IAztecToolbarClickListener,
         IHistoryListener,
         OnRequestPermissionsResultCallback,
@@ -394,6 +395,7 @@ open class MainActivity : AppCompatActivity(),
                 .setOnAudioTappedListener(this)
                 .setOnMediaDeletedListener(this)
                 .setOnVideoInfoRequestedListener(this)
+                .setOnTextContextMenuItemListener(this)
                 .addPlugin(WordPressCommentsPlugin(visualEditor))
                 .addPlugin(MoreToolbarButton(visualEditor))
                 .addPlugin(PageToolbarButton(visualEditor))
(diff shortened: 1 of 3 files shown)


hypest:
Since this PR is part of the work for https://github.com/wordpress-mobile/WordPress-Android/issues/8275, may I suggest putting it on hold until we also have the wpandroid side PR to test it with?

mzorz:
Closing this in favor of #745 

For completeness, will tell the full story here as this PR and #745 do not look alike whatsoever :)

### History
When considering https://github.com/wordpress-mobile/WordPress-Android/issues/8275 first I tried to follow along the lines of this other PR of mine (https://github.com/wordpress-mobile/WordPress-Android/pull/8204/) that fixed a similar issue by making sure the Post object gets updated when sharing content to WordPress app, in this sequence:

1. an intent to share is triggered and captured by EditPostActivity
2. given there's no new content still, AztecText is fed with the new content right away, and the initial SHA calculated corresponds to this content
3. hence, when tapping back or PUBLISH, `hasChanges` responded with `NO_CHANGES`, which is the right behavior we want from Aztec.
4. for this, we initialize/update the Post object manually with the content passed through the intent in `EditPostActvity` for this case

I thought we were under similar conditions here (instead of an intent to share, we get a paste action to initialize AztecText's content). Given the source of input for this event is the EditText itself, I thought let's use a listener interface and let's update the Post object on EditPostActivity just like we do for share intents, but only when the EditText signals us that there's content being pasted. So it goes like this:
1. user opens AztecText and its hash gets initialized with ""(empty string)
2. user performs the paste operation
3. we get signaled by this new interface that something has been pasted
4. once that has happened, we get current content from Aztec after it's been pasted and finally we update the Post object accordingly.

However, there's the logic about `hasChanges` that returns `NO_CHANGES`. I then started investigating and I could see there's a call to `fromHtml` in `paste()`. Each time `fromHtml` is called, a new initial hash gets calculated, as the code is assuming `fromHtml` makes sense when starting AztecText (or moving from HTML to the visual editing view, which is the same to all effects).

Hence, a new PR fixing the issue has been created in #745 .

PR #750, opened 2018-10-10
Fix empty span handling

### LINKED ISSUE #749: Crash when deleting text
#### To reproduce

1. Open the demo app (in the emulator)
2. Select all text
3. Press the delete key
4. Notice the app crashes

mzorz:
Fixes #749 

This PR makes sure to only run `handleTextChanged` in `BlockHandler` if there are spans of the kind to be processed within the passed indexes.

### Test
1. Open the demo app (in the emulator)
2. Select all text
3. Press the delete key
4. Notice the app doesn't crash anymore


daniloercoli:
Looks good! :shipit:

PR #754, opened 2018-11-06
Change image/video/audio onTapped interfaces for AztecAttributes immutability

mzorz:
This PR changes image/video/audio `onTapped` interfaces for `AztecAttributes`'s immutability in case the attributes passed to these listeners get modified during listener calls, needed after attributes immutability change in PR #751

### Test
Steps provided in WPAndroid PR.

### Review
@planarvoid @malinajirka

cc @loremattei

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecAudioSpan.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecAudioSpan.kt
@@ -20,6 +20,10 @@ class AztecAudioSpan(context: Context, drawable: Drawable?, override var nesting
     }
 
     override fun onClick() {
-        onAudioTappedListener?.onAudioTapped(attributes)
+        val newAttributes =
+                onAudioTappedListener?.onAudioTapped(attributes)
+        if (newAttributes != null) {
+            attributes = newAttributes
+        }
     }
 }
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -791,11 +791,12 @@ open class MainActivity : AppCompatActivity(),
         mediaUploadDialog!!.show()
     }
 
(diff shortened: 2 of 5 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecAudioSpan.kt
```diff
@@ -20,6 +20,10 @@ class AztecAudioSpan(context: Context, drawable: Drawable?, override var nesting
     }
 
     override fun onClick() {
-        onAudioTappedListener?.onAudioTapped(attributes)
+        val newAttributes =
```
hypest: Nit but, can we replace the variable plus null check conditional with:
```suggestion
        onAudioTappedListener?.onAudioTapped(attributes)?.let { attributes = it; }
```

Same for the other spans in this PR.

mzorz: oh thanks, addressed in bb9b116f

mzorz:
> Edit: Answering my own question after reading the other PR: Apparently yes, wpandroid tries to change some attributes in-place.

That's correct :) just for completeness I'll leave the description here: it is of particular help for uploads.  we know which Span was clicked because we're listening, but we don't have a way to search for them and match them individually. So,  after the upload finishes (and as it progresses) we need to be able to modify it to show the progress overlay, etc. and for that we need some way to identify the span (we do that with the help of setting an `id` attribute of the sorts). So that's why we need to be able to set/modify the attributes of an `AztecSpan` from within the touch listeners, to set the id (even if temporarily) and be able to search for the span again via `AttributePredicate` to be able to update their attributes or paint some overlay and so on.

We could have left it up to the host app to call `AztecText.updateElementAttributes()` once things were modified but it seemed more transparent to enforce it via the interface.

mzorz:
Marking with `[not ready for review]` until a decision is reached about https://github.com/wordpress-mobile/WordPress-Android/pull/8554

Issue #760, opened 2018-11-08
Repeating characters in the editor


0nko:
This looks similar to #580, which was caused by the use of an `InputFilter`. 

This bug was reported in WPAndroid `11.1` release, after [this change](https://github.com/wordpress-mobile/AztecEditor-Android/pull/746/files#diff-7ca71cd1f009ab8438e0901205001b14R456) from Aztec `1.3.10` was integrated.

![image](https://user-images.githubusercontent.com/1522856/48194573-b0067400-e34d-11e8-8be7-727eee0abd2f.png)

PR #761, opened 2018-11-08
Fix for repeated characters in the editor

daniloercoli:
This is an hotfix for #760 that removes the InputFilter used to detect the Enter.Key.

Enter.Key detection logic is only used in Gutenberg-mobile, and we will propose another solution, tested on custom keyboards that are giving us problems with InputFilters, in the next days.

cc @0nko

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -455,44 +455,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             source
         }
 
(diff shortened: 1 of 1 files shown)


sergpetrov:
Hi. How can I detect onEnterKey for current moment? How is correctly?

daniloercoli:
> Hi. How can I detect onEnterKey for current moment? How is correctly?

Hi @sergpetrov - We're using `TextWatchers` to detect Enter.Key pressed event. See the implementation here: https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/aztec/src/main/kotlin/org/wordpress/aztec/watchers/EnterPressedWatcher.kt

sergpetrov:
@daniloercoli thank you!

PR #763, opened 2018-11-12
Bring enter.key detection back to Aztec by using TextWatcher

daniloercoli:
`Enter.Key` detection code was added to Aztec a couple of weeks ago by using [InputFilters](https://github.com/wordpress-mobile/AztecEditor-Android/pull/746/files#diff-7ca71cd1f009ab8438e0901205001b14R456).
Unfortunately `InputFilters` can act badly on some combination of keyboard/language, resulting in repeating letters in editor.

This PR proposes an alternative solution to the problem by using `TextWatcher`, that should be safe to use. (Enter.Key detection logic is only used in Gutenberg-mobile at the moment, and placing a textwatcher that does check if the lister is set or not, before processing the text, should also be another +1 for this solution).

To test this PR, setup  a new OnAztecKeyListener in AztecText, and check that no new lines are added to the editor when Enter.Key is pressed on the keyboard of the device.
```
        onAztecKeyListener = object : OnAztecKeyListener {
            override fun onEnterKey() : Boolean {
                Log.d("Test", "testing")
                return true
            }
            override fun onBackspaceKey() : Boolean {
                return false
            }
        }
```

I've tested this with many devices, and languages, and didn't find the repeating letters issues, but wasn't able to replicate the issue even with the old `InputFilters` solution.

Pinging @khaykov , since he was able to replicate the original issue with InputFilters.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -47,7 +47,6 @@ import android.text.TextWatcher
 import android.text.style.SuggestionSpan
 import android.util.AttributeSet
 import android.util.DisplayMetrics
-import android.util.Log
 import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.MotionEvent
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -47,6 +47,7 @@ import android.text.TextWatcher
 import android.text.style.SuggestionSpan
 import android.util.AttributeSet
 import android.util.DisplayMetrics
+import android.util.Log
 import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.MotionEvent
@@ -460,7 +461,46 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             source
         }
 
-        filters = arrayOf(emptyEditTextBackspaceDetector)
+        val detectEnterKeyInputFilter = InputFilter { source, start, end, dest, dstart, dend ->
+            if (isTextChangedListenerDisabled() || isHandlingEnterEvent || !isViewInitialized) {
+                // If the view is not initialized do nothing and accept the changes
+                null
+            } else if (end > 1 && start == 0 && dstart == 0 && dend == 0) {
+                // When the initial content is set to Aztec accept the changes without checking
+                // This case is just an additional check that should never happen if
+                // you call `fromHTML` since isTextChangedListenerDisabled does the trick
+                null
+            } else
+            //  You sometimes get a SpannableStringBuilder, sometimes a plain String in the source parameter
+                if (source is SpannableStringBuilder) {
+                    isHandlingEnterEvent = true
+                    for (i in end - 1 downTo start) {
+                        val currentChar = source[i]
+                        Log.d("Danilo", "Type : "+Character.getType(currentChar))
+                        if (currentChar == '\n' /*&& onKeyListener?.onEnterKey() == true*/) {
+                            source.replace(i, i + 1, "")
+                        }
+                    }
+                    isHandlingEnterEvent = false
+                    source
+                } else {
+                    isHandlingEnterEvent = true
+                    val filteredStringBuilder = StringBuilder()
+                    for (i in start until end) {
+                        val currentChar = source[i]
+                        Log.d("Danilo", "Type : "+Character.getType(currentChar))
+                        if (currentChar == '\n' /*&& onKeyListener?.onEnterKey() == true*/) {
+                            // nothing
+                        } else {
+                            filteredStringBuilder.append(currentChar)
+                        }
+                    }
+                    isHandlingEnterEvent = false
+                    filteredStringBuilder.toString()
+                }
+        }
+
+        filters = arrayOf(emptyEditTextBackspaceDetector, detectEnterKeyInputFilter)
     }
 
     private fun handleBackspaceAndEnter(event: KeyEvent): Boolean {
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -753,7 +753,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
 
     /**
      * Sets the Aztec key listener to be used with this AztecText.
-     * Please note that this lister does hold a copy of the whole text in the editor
+     * Please note that this listener does hold a copy of the whole text in the editor
      * each time a key is pressed.
      */
     fun setAztecKeyListener(listenerAztec: OnAztecKeyListener) {
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -476,7 +476,6 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                     isHandlingEnterEvent = true
                     for (i in end - 1 downTo start) {
                         val currentChar = source[i]
-                        Log.d("Danilo", "Type : "+Character.getType(currentChar))
                         if (currentChar == '\n' /*&& onKeyListener?.onEnterKey() == true*/) {
                             source.replace(i, i + 1, "")
                         }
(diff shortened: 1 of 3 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/watchers/EnterPressedWatcher.kt
```diff
@@ -0,0 +1,49 @@
...
+class EnterPressedWatcher(aztecText: AztecText) : TextWatcher {
+
+    private val aztecTextRef: WeakReference<AztecText?> = WeakReference(aztecText)
+    private var textBefore : SpannableStringBuilder? = null
+    private var start: Int = -1
+
+    override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) {
+        val aztecText = aztecTextRef.get()
+        if (aztecText?.getAztecKeyListener() != null && !aztecText.isTextChangedListenerDisabled()) {
+            // we need to make a copy to preserve the contents as they were before the change
```
mzorz: Given we really only need to extend the behavior of detecting `Enter` for our Gutenberg purposes for now, I wonder if there is a way for us to know we're in "Gutenberg mode" at this level, and just only do this copy / enable (or even install) the `EnterPressedWatcher` when needed. The reason behind this is I'm thinking we may have long content on usual Aztec, and making a copy in memory will probably become a problem (as opposed to Gutenberg blocks which should be shorter).
I understand that if the listener is not set, then this won't be run, but it also means the `onBackspaceKey` will have to be disabled. If these two are closely related to Gutenberg, should we then make that explicit? And if on the contrary these 2 are not closely related, can we make a separation by splitting the interface into 2 different ones (one for `onBackspaceKey` and another one for `onEnterKey`)?
What do you think?

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -465,12 +466,17 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
-            return onKeyListener?.onEnterKey() ?: false
+            // Check if the external lister has consumed the enter pressed event
+            // In that case stop the execution
+            if (onAztecKeyListener?.onEnterKey() == true) {
+                return true
+            }
         }
+
         if (event.action == KeyEvent.ACTION_DOWN && event.keyCode == KeyEvent.KEYCODE_DEL) {
             // Check if the external lister has consumed the backspace pressed event
```
mzorz: minor typo, not strictly related to the change but let's fix this as we're on it: let's change `lister` for `listener`

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -465,12 +466,17 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
 
     private fun handleBackspaceAndEnter(event: KeyEvent): Boolean {
         if (event.action == KeyEvent.ACTION_DOWN && event.keyCode == KeyEvent.KEYCODE_ENTER) {
-            return onKeyListener?.onEnterKey() ?: false
+            // Check if the external lister has consumed the enter pressed event
```
mzorz: same typo as indicated below here:  `lister` for `listener`

aztec/src/main/kotlin/org/wordpress/aztec/watchers/EnterPressedWatcher.kt
mzorz: Let me clarify my thoughts: I think we should only [`install()` the `EnterPressedWatcher`](https://github.com/wordpress-mobile/AztecEditor-Android/pull/763/files#diff-7ca71cd1f009ab8438e0901205001b14R507) when the `onAztecKeyListener` is set (so the watcher is not run every time). Also, we should probably split the interface into 2 different interfaces.

khaykov:
I tested the 1.3.10 with JP input to confirm the issue, then 1.3.11 and this branch to confirm that there is no issue with JP input.

aztec/src/main/kotlin/org/wordpress/aztec/watchers/EnterPressedWatcher.kt
daniloercoli: Yes, we can `install` the enter watcher when the key listener is set on Aztec, but it will be added at the end of the list of watchers, instead we want it at the beginning to prevent other watchers to modify the text before it.
Most probably there is no difference at the moment, but who knows in the future?
Also, this enter.key watcher is always installed, but nothing is copied in memory if the key-listener is null. It's only copied when the listener is set in the AztecWrapper for React Native. 
Same is for the backspace listener, it does intercept the key pressed, only when the listener is set.

Note that Aztec has already a big pile of watchers, that do copy the current content in memory each time the user presses a key.


> Also, we should probably split the interface into 2 different interfaces.

We had 2 distinct interfaces, but merged it in one, since it's all about intercepting a key pressed event. There is no big difference for me, but i prefer to have 1 interface and 2 or 3 methods, each for a different key.

mzorz: > but it will be added at the end of the list of watchers, instead we want it at the beginning to prevent other watchers to modify the text before it.

I see what you mean. Can we then add a comment in code, to make sure to warn any other developer seeing that does not change the place of the `install()` call and leaves it as the very first TextWatcher to be added?

> nothing is copied in memory if the key-listener is null. 

Gotcha,  thanks for the clarification @daniloercoli.

> Note that Aztec has already a big pile of watchers, that do copy the current content in memory each time the user presses a key.

A clarification here: while it's true there are a lot of TextWatchers (and some recursion in it), AFAICT the only ones that copy content are the HistoryLogger (which seems to be the heaviest in terms of memory usage) and the ObservationQueue but this latter one only copies at the event (textChanged) level and for a very limited size (then longest is as long as a known pattern size is, which is max 4 events currently). The other watchers only perform specific in-line modifications on spans as long as I can see.

Also, maybe we should add a comment (or lint warning?) to the interface so the developer knows that by setting a listener for such interface a copy of the text will be generated on each `enter` pressed

daniloercoli: > Can we then add a comment in code, to make sure to warn any other developer seeing that does not change the place of the install() call and leaves it as the very first TextWatcher to be added?

> Also, maybe we should add a comment (or lint warning?) to the interface so the developer knows that by setting a listener for such interface a copy of the text will be generated on each enter pressed

Right, will add a comment on both places. 👍 

> A clarification here: while it's true there are a lot of TextWatchers (and some recursion in it), AFAICT the only ones that copy content are the HistoryLogger (which seems to be the heaviest in terms of memory usage) and the ObservationQueue but this latter one only copies at the event (textChanged) level and for a very limited size (then longest is as long as a known pattern size is, which is max 4 events currently). The other watchers only perform specific in-line modifications on spans as long as I can see.

I quickly checked yesterday, when started working on this PR, and it seems there are other watchers that does copy the content. `EndOfParagraphMarkerAdder` is one of those for example.

mzorz: > I quickly checked yesterday, when started working on this PR, and it seems there are other watchers that does copy the content. EndOfParagraphMarkerAdder is one of those for example.

Oh thanks for pointing that out. Just saw that 👍

daniloercoli: `SuggestionWatcher` and `InlineTextWatcher` are other examples of watchers that hold a  copy of the whole text in `beforeTextChanged`.

mzorz: yep I see `TextChangedEvent` holding the copy there 👍 . I imagine we could have 1 instance of the TextChangedEvent and pass it through to all listeners 🤔 (Not in this PR, just thinking aloud)

daniloercoli: Yes, we can probably open an issue for this to keep track of it.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -748,6 +751,11 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         return this.onAztecKeyListener
     }
 
+    /**
+     * Sets the Aztec key listener to be used with this AztecText.
+     * Please note that this lister does hold a copy of the whole text in the editor
```
mzorz: can we change `lister`  for `listener` 😉

aztec/src/main/kotlin/org/wordpress/aztec/watchers/EnterPressedWatcher.kt
```diff
@@ -0,0 +1,49 @@
...
+            textBefore = SpannableStringBuilder(text)
+            this.start = start
+        }
+    }
+
+    override fun afterTextChanged(text: Editable) {
+        val aztecText = aztecTextRef.get()
+        val aztecKeyLister = aztecText?.getAztecKeyListener()
```
mzorz: also let's rename `aztecKeyLister` for `aztecKeyListener`

daniloercoli: Not sure if it's me, or some sort of spell checker. 🤦‍♂️

PR #775, opened 2019-01-13
Issue/774 caption alignment

### LINKED ISSUE #774: Set caption span align property
When assigning attributes, the `align` property of the caption span is not synced with the alignment attributes. This can cause bugs when clients rely on them to be both up to date.

0nko:
Fixes #774.

This PR keeps the caption attributes and the align property of the caption span in sync. This is necessary to fix the image alignment property bug (wordpress-mobile/WordPress-Android#8569).

This PR can be tested by a related PR in WPAndroid (trying to change the alignment of an image with a caption).

### DIFF
--- a/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/extensions/CaptionExtensions.kt
+++ b/wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/extensions/CaptionExtensions.kt
@@ -45,11 +45,12 @@ fun AztecText.removeImageCaption(attributePredicate: AztecText.AttributePredicat
 
 fun AztecText.hasImageCaption(attributePredicate: AztecText.AttributePredicate): Boolean {
     this.text.getSpans(0, this.text.length, AztecImageSpan::class.java)
-        .firstOrNull {
-            val wrapper = SpanWrapper<AztecImageSpan>(text, it)
-            return text.getSpans(wrapper.start, wrapper.end, CaptionShortcodeSpan::class.java)
-                    .map { span -> span as CaptionShortcodeSpan }
-                    .any { span -> attributePredicate.matches(span.attributes) }
+            .firstOrNull {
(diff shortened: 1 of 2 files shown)


wordpress-shortcodes/src/main/java/org/wordpress/aztec/plugins/shortcodes/extensions/CaptionExtensions.kt
```diff
@@ -44,17 +45,20 @@ fun AztecText.removeImageCaption(attributePredicate: AztecText.AttributePredicat
 
 fun AztecText.hasImageCaption(attributePredicate: AztecText.AttributePredicate): Boolean {
     this.text.getSpans(0, this.text.length, AztecImageSpan::class.java)
-        .firstOrNull {
-            val wrapper = SpanWrapper<AztecImageSpan>(text, it)
-            return text.getSpans(wrapper.start, wrapper.end, CaptionShortcodeSpan::class.java).isNotEmpty()
+            .firstOrNull {
+                attributePredicate.matches(it.attributes)
```
hypest: Hey @0nko , just to be sure, does the filtering contribute to the fix of #775 ? Asking just to understand if there's some more code paths I should review and test. Thanks!

0nko: No. I just noticed that the function is not using the parameter and it wasn't working as expected. It's not even being used anywhere at the moment, just in the unit tests. I know PRs should be specific but since I was already working on the file I wanted to make a quick fix. We can remove it if you prefer.

hypest: Ah, cool. No, we don't need to remove it, it's fine. Just wanted to make sure I'm not missing something. Looks good to me!

0nko:
Thanks @hypest!

PR #780, opened 2019-01-24
Revert #776

marecar3:
Revert #776 as we don't need it


hypest:
For context, #776 is merged but the relevant PR that uses it (https://github.com/wordpress-mobile/gutenberg-mobile/pull/275) isn't, and it might even end up in a different approach. Under that light, leaving in #776 is kinda risky since it's not fully tested with the editor and doesn't make perfect sense to ship it in today's release. So, we chatted with @Tug to leave it out for now and re-introduce it later.

marecar3:
Thanks, @hypest for the context.

marecar3:
Thanks @hypest , will do that! :)

PR #790, opened 2019-03-06
Performance improvement on parsing from html

hypest:
This PR introduces a couple of speed of optimizations to address the slow start of the editor when the content is large. Relevant issue on wpandroid: https://github.com/wordpress-mobile/gutenberg-mobile/issues/672

What's included:
1. A new, additional `parseHtml...` method (`parseHtmlForInspection`). This one is forked from `fromHtml()` and has these optimizations:
  a. Use of `SpannableString` instead of `SpannableStringBuilder`. We don't want the (slower) `...Builder` variant here since we're not planning on modifying the text content of the String while processing it in this method. Some other functions were changed from expecting `SpannableStringBuilder` to expecting a `Spannable` since they didn't actually needed the editability `SpannableStringBuilder`.
  b. Don't use the very expensive `addVisualNewlinesToBlockElements()` since the parsed content is not going to be used for rendering it in the editor.
  c. Don't use `markBlockElementsAsParagraphs()` since the spannable will not be used for _editing_ at all so, no need for the PARAGRAPH_SPAN span behavior.
  d. Don't use `cleanupZWJ()` as it doesn't matter
2. Added a LIFO `tagStack` to optimize the handling of end-tags in `AztecTagHandler.kt`. Up to now, `AztecTagHandler.end()` was using `getLast()` which scans the spans to find the most recent "open" tag, but that can be sped up since we know about the nested nature of the html tags. So, a LIFO tracks the opening/closings of the tags and we normally expect the most recent open tag to be exactly at the top of the stack.

### Test A
1. Editing in the demo app should work as normal

### Test B
1. Use [this wpandroid PR](https://github.com/wordpress-mobile/WordPress-Android/pull/9370) and verify that the starting of the editor is now faster on long posts.

### Review
@[USER_NAME]


aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -163,8 +166,8 @@ class AztecTagHandler(val context: Context, val plugins: List<IAztecPlugin> = Ar
             start(output, AztecMediaClickableSpan(mediaSpan))
             output.append(Constants.IMG_CHAR)
         } else {
-            end(output, mediaSpan.javaClass)
             end(output, AztecMediaClickableSpan::class.java)
+            end(output, mediaSpan.javaClass)
```
mzorz: wondering, was this place interchange a bug before?

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -177,11 +180,23 @@ class AztecTagHandler(val context: Context, val plugins: List<IAztecPlugin> = Ar
...
     private fun start(output: Editable, mark: Any) {
+        tagStack.add(mark)
+
         output.setSpan(mark, output.length, output.length, Spanned.SPAN_MARK_MARK)
     }
 
     private fun end(output: Editable, kind: Class<*>) {
-        val last = output.getLast(kind)
+        // Get most recent tag type from the stack instead of `getLast()`. This is a speed optimization as `getLast()`
+        //  doesn't know that the tags are in fact nested and in pairs (since it's html)
```
mzorz: not sure I'm following correctly, the comment here seems to assume there can't be self-closing tags, for instance?

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
hypest: Hah, it wasn't a bug because on the spans-space, there is no meaning to the particular order the spans will be added to the Spannable or not.

But, because now the `start()` and `end()` methods use a LIFO Stack to keep track of the nesting of spans, the out-of-order `end()` calls fail to find the expected tag/span at the top of the Stack.

In this particular case, there are 2 `start` calls when opening the tag (https://github.com/wordpress-mobile/AztecEditor-Android/blob/bd1c9fa37ff5f332720186c3f5d1dee5f4e0944d/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt#L165-L166) where the first one pushes a `AztecMediaSpan` to the stack and the second pushes a `AztecMediaClickableSpan`. So, when ending the tag, we need to call the `end`s in reverse order. That's the change introduced here.

Does that make sense?

mzorz: oh perfect - yes that makes sense 👌 thanks for clarifying!

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
hypest: I guess the wording is not the clearest it could be. Self-closing tags (empty elements in html parlance) are treated as pairs of tags by tagsoup (the parsing library we're using) anyway. So for instance, a self-closing tag will cause both a call to the `start()` and the `end()` methods.

Does this address your concern?

I will update the wording to include this detail too.

mzorz: > Self-closing tags (empty elements in html parlance) are treated as pairs of tags by tagsoup 

👌 perfect! thanks :)

hypest:
Mario mentioned to me over Slack that he has no further review comments to add so, this is now 👍 to merge when CI passes.

PR #793, opened 2019-03-13
Clickable URL Spans

adrielcafe:
Hi! 
The current implementation of EnhancedMovementMethod only handles AztecMediaClickableSpan and UnknownClickableSpan.
I had to make the following changes to fit my app needs (to also handle AztecURLSpan) and thought this might be useful to others.

### Toggle clickable URL spans
Use case: an editor in **read-only** mode that needs to open urls
```kotlin
visualEditor.setUrlClickable(true)
```

### Set custom onClick listener (optional)
If null, will call the original `onClick()` method.
Use case: If the editor needs to open a Chrome Custom Tab (the default behavior is to open in a external app)
```kotlin
visualEditor.setOnUrlClickListener(object : EnhancedMovementMethod.OnUrlClickListener {
    override fun onClick(widget: View, url: String) {
        openInChromeTab(url)
    }
})
```

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -1765,4 +1765,12 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
     override fun dispatchHoverEvent(event: MotionEvent): Boolean {
         return if (accessibilityDelegate.onHoverEvent(event)) true else super.dispatchHoverEvent(event)
     }
+
+    fun setUrlClickable(urlClickable: Boolean){
+        EnhancedMovementMethod.clickableUrlSpan = urlClickable
+    }
+
+    fun setOnUrlClickListener(listener: EnhancedMovementMethod.OnUrlClickListener){
+        EnhancedMovementMethod.urlClickListener = listener
+    }
 }
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -1770,7 +1770,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         EnhancedMovementMethod.clickableUrlSpan = urlClickable
     }
 
-    fun setOnUrlClickListener(listener: EnhancedMovementMethod.OnUrlClickListener){
+    fun setOnUrlClickListener(listener: EnhancedMovementMethod.OnUrlClickListener?){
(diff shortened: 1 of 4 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/EnhancedMovementMethod.kt
```diff
@@ -72,14 +77,21 @@ object EnhancedMovementMethod : ArrowKeyMovementMethod() {
...
+                    return true
                 }
-                return true
             }
         }
 
         return super.onTouchEvent(widget, text, event)
     }
+
+    interface OnUrlClickListener {
```
0nko: All the tap (and other) listener interfaces live inside the `AztecText`. Could we move it there, for consistency?

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1778,4 +1778,12 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
     override fun dispatchHoverEvent(event: MotionEvent): Boolean {
         return if (accessibilityDelegate.onHoverEvent(event)) true else super.dispatchHoverEvent(event)
     }
+
+    fun setUrlClickable(urlClickable: Boolean) {
+        EnhancedMovementMethod.clickableUrlSpan = urlClickable
+    }
+
+    fun setOnUrlClickListener(listener: EnhancedMovementMethod.OnUrlClickListener?) {
```
0nko: Would you mind adding a configuration method to the `Aztec` class, similar to [this](https://github.com/adrielcafe/AztecEditor-Android/blob/f5c71e91f338aaf83554f2c75e5eadb4cc9da629/aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt#L101-L105)?

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1778,4 +1778,12 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
     override fun dispatchHoverEvent(event: MotionEvent): Boolean {
         return if (accessibilityDelegate.onHoverEvent(event)) true else super.dispatchHoverEvent(event)
     }
+
+    fun setUrlClickable(urlClickable: Boolean) {
```
0nko: This method could also exist in the `Aztec` fluid configuration class.

adrielcafe:
Hey @0nko, thanks for reviewing my code!

Just made the suggested changes, please let me know if I need to improve anything else.

0nko:
Looks great, thanks! :shipit:

PR #795, opened 2019-03-19
Fix for crash on Samsung Clipboard ListView on Android 8.X

daniloercoli:
This PR fixes a crash that happens when pasting text on Samsung Devices running Android 8.
The issue only happens if you use the custom Samsung `Clipboard` feature - 
Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/8827

Steps to reproduce:

Make sure you have Samsung device with Android 8.x
- Start the demo app
- Hold down finger over the screen until a menu with: PASTE, CLIPBOARD doesn't show up.
- Click on CLIPBOARD
- See the Toast message and that the custom list view doesn't show

The solution implemented in this PR is the less invasive, and easy to remove later, other viable fixes are listed here https://github.com/wordpress-mobile/WordPress-Android/issues/8827#issuecomment-454520085 

Note: we will update the Aztec ref in the Android app soon, with the next GB-mobile release, so no need to update it now on both projects.

cc @marecar3  since you've already investigate the problem.

Note 2: If we find a nice way to disable HW acceleration, on selected devices only, we may need to remove this fix, and try if it's crashing again when HW acceleration is OFF. Since this could be similar to #8828.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -47,6 +47,7 @@ import android.text.TextWatcher
 import android.text.style.SuggestionSpan
 import android.util.AttributeSet
 import android.util.DisplayMetrics
+import android.util.Log
 import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.MotionEvent
@@ -1385,6 +1386,9 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             max = Math.max(0, Math.max(selectionStart, selectionEnd))
         }
 
+        //
+        var clipboardIdentifier = resources.getIdentifier("android:id/clipboard", "id", context.packageName)
+
         when (id) {
             android.R.id.paste -> paste(text, min, max)
             android.R.id.pasteAsPlainText -> paste(text, min, max, true)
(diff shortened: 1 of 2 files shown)


malinajirka:
Thanks for the fix! Unfortunately, I have access just to a Samsung with Android 4.4. @marecar3 Could you please review this PR? Thanks!

marecar3:
Hey @malinajirka, yup I can do that, will respond with a proper comment.

marecar3:
Hey @malinajirka @daniloercoli , On my device `Samsung SM-T585` Android 8.1 this solution isn't working as it has different value for the Clipboard `16908872`.

Generally not sure if it's good to put the hardcoded value from the Android platform, as it can be different on various devices (like in this case) and also if it's maybe working now, it's not guaranteed that Google/Samsung/... won't change it in the future.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1400,7 +1401,18 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                     deleteInlineStyleFromTheBeginning()
                 }
             }
-            else -> return super.onTextContextMenuItem(id)
+            // Fix for crash when pasting text on Samsung Devices running Android 8.
+            // Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/8827
+            16908904,
```
marecar3: This value on my device : Samsung SM-T585 Android 8.1 is different -> 16908872.

daniloercoli: Interesting.... `16908874` --> was from Android 8, and `16908904` from Android 9. I just added it since don't know if 8.X will get the latest version of their clipboard.

I guess i can add your other value to the list, but it's like going per-case. Anyway don't have better idea, since the src code of their clipboard isn't available.

marecar3: @daniloercoli let me try something, and if doesn't work I will give the Approve so that we can move on.

marecar3:
Hey @daniloercoli if you don't mind I have changed initial solution and put some dynamic solution now, it's working on my device, please can you try on your device?

daniloercoli:
Nice improvement @marecar3! Worked as expected!

Issue #729, opened 2018-06-21
Add a character before the picture in demo app crashes Aztec


daniloercoli:
This is weird. 
I tried to add some character before the first picture in demo app and Aztec crashed with the following stack trace:
```
                  java.lang.ArrayIndexOutOfBoundsException: length=39; index=-1
                      at android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:646)
                      at android.widget.Editor.drawHardwareAccelerated(Editor.java:1787)
                      at android.widget.Editor.onDraw(Editor.java:1756)
                      at android.widget.TextView.onDraw(TextView.java:7723)
                      at android.view.View.draw(View.java:20338)
                      at android.view.View.updateDisplayListIfDirty(View.java:19283)
                      at android.view.View.draw(View.java:20061)
                      at android.view.ViewGroup.drawChild(ViewGroup.java:4421)
                      at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4207)
                      at android.view.View.updateDisplayListIfDirty(View.java:19274)
                      at android.view.View.draw(View.java:20061)
                      at android.view.ViewGroup.drawChild(ViewGroup.java:4421)
                      at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4207)
                      at android.view.View.draw(View.java:20341)
                      at android.widget.ScrollView.draw(ScrollView.java:2731)
                      at android.view.View.updateDisplayListIfDirty(View.java:19283)
                      at android.view.View.draw(View.java:20061)
                      at android.view.ViewGroup.drawChild(ViewGroup.java:4421)
                      at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4207)
                      at android.view.View.updateDisplayListIfDirty(View.java:19274)
                      at android.view.View.draw(View.java:20061)
                      at android.view.ViewGroup.drawChild(ViewGroup.java:4421)
                      at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4207)
                      at android.view.View.updateDisplayListIfDirty(View.java:19274)
                      at android.view.View.draw(View.java:20061)
                      at android.view.ViewGroup.drawChild(ViewGroup.java:4421)
                      at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4207)
                      at android.view.View.updateDisplayListIfDirty(View.java:19274)
                      at android.view.View.draw(View.java:20061)
[...]

PR #801, opened 2019-03-28
Android 8.x dymanic layout crash preventer

mzorz:
### Fix
This PR tackles the symptom (not the culprit) identified as the cause in #729.
The culprit having been identified as laying in the lower level within the Android OS itself https://android-review.googlesource.com/c/platform/frameworks/base/+/634929, and for which a real solution means to fix it at that level (see https://github.com/wordpress-mobile/WordPress-Android/issues/8828#issuecomment-474876535 for alternatives explored around a real solution to the specific problem), we can't do much about it.
As discussed elsewhere, I decided to try and approach the issue from the _symptom_ point of view, and try and workaround the problem without affecting functionality.

The problem was narrowed down and isolated (symptom-wise) as per the following statement:

```
"the Editor crashes when inserting anything right before an image."
```

One of the proposed approaches was to just overwrite `onDraw()` (where the crash is about to happen) and just filter out any non-acceptable values (that is, anything below 0) for coordinates as suggested in https://github.com/wordpress-mobile/AztecEditor-Android/issues/729#issuecomment-430105339. I think at that point however, it would mean that previous calculations were wrong for some reason, and while I haven't dove in too deep it might also mean some state variables (i.e. class members in `DynamicLayout`) could end up keeping a wrongly calculated state, thus generating another problem later on.

Due to these reasons, I decided a different approach would be to:

- a) first, see if we can programmatically detect the exact moment when the problem statement above is `true`
- b) then, somehow make the `TextView` to reflow / recalculate the view elements as appropriate.

One of the first ideas that came to mind was to make use of the already in place mechanism of `TextWatchers` to go ahead and make any needed changes. However, `TextWatcher` only is an observer that cannot affect the flow of events - that means that whatever is going to happen is still going to happen, you only get to listen to what's going on and only do something with it _after the fact_.

Then, we have `InputFilter`. This effectively is a mechanism in place that allows you to look into what's about to happen (text change) and decide on what should be replaced and what not.

I observed that initializing `AztecText` with actually _anything_ before an Image is not a problem (only insertion is), so I thought "ok, what if we replace not the specific part that normally would get replaced, but make it look like a block and replace _both_ the thing being modified _and_ the Image together?".

Unfortunately that's not exactly possible from within an `InputFilter`. When an insertion happens, the `Editable` will call its whatever `InputFilter`s have been set, passing them the place where the insertion needs to happen, and what the source string (or `Spannable`) to insert is. What this means is, even if we return the block of "inserted char" plus the Image right next to it, the whole block still will try to get inserted before the original image, causing the crash anyway.

#### What this PR does
So, we'd have to first keep a copy of the image span, then delete the original image span, then insert the block of "inserted text plus image" and, for the run of this inputFilter (when the case is first detected), just return an empty string, actually preventing the original change from happening. This is exactly what the `InputFilter` in this PR is doing and, so far so good, seems to be working.

We're effectively artificially re-creating the change as an insertion of "original insertion plus Image" and then making a substitution, if that makes sense.

### Test
1. start the demo app
2. place the cursor at the  beginning, right before the image
3. enter anything
4. observe it doesn't crash

Other things I've tested:
- [x] test with other pluggable keyboards (tried 3rd party such as SwiftKey)
- [x] test with different types of content in between image and insertion (heading, list, quote, another image, etc)
- [x] test pasting paragraphs (not just typing characters)
- [x] tested integration with WPAndroid (used [these steps and can't reproduce](https://github.com/wordpress-mobile/WordPress-Android/issues/9470#issuecomment-476782575))

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -420,7 +420,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         // triggers ClickableSpan onClick() events
         movementMethod = EnhancedMovementMethod
 
-        setupBackspaceAndEnterDetection()
+        setupKeyListenersAndInputFilters()
 
         //disable auto suggestions/correct for older devices
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
@@ -440,10 +440,42 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
     // Setup the keyListener(s) for Backspace and Enter key.
     // Backspace: If listener does return false we remove the style here
     // Enter: Ask the listener if we need to insert or not the char
-    private fun setupBackspaceAndEnterDetection() {
+    private fun setupKeyListenersAndInputFilters() {
         //hardware keyboard
         setOnKeyListener { _, _, event ->
             handleBackspaceAndEnter(event)
+        }
+
+        // This InputFilter created only for the purpose of avoiding crash described here:
+        // https://android-review.googlesource.com/c/platform/frameworks/base/+/634929
+        // https://github.com/wordpress-mobile/AztecEditor-Android/issues/729
+        // the rationale behind this workaround is that the specific crash happens only when adding/deleting text right
+        // before an AztecImageSpan, so we detect the specific case and re-create the contents only when that happens.
+        // This is indeed tackling the symptom rather than the actual problem, and should be removed once the real
+        // problem is fixed at the Android OS level as described in the following url
+        // https://android-review.googlesource.com/c/platform/frameworks/base/+/634929
+        val dynamicLayoutCrashPreventer = InputFilter { source, start, end, dest, dstart, dend ->
+            var temp = source
+            if (dest.length > dend+1) {
+                // if there are any images right after the destination position, hack the text
+                val spans = dest.getSpans(dstart, dend+1, AztecImageSpan::class.java)
+                if (spans.isNotEmpty()) {
+                    // test: is this an insertion?
+                    if (dstart == dend) {
+                        // take the source (that is, what is being inserted), and append the Image to it. We will delete
+                        // the original Image later so to not have a duplicate.
+                        // use Spannable to copy / keep the current spans
+                        temp = SpannableStringBuilder(source).append(dest.subSequence(dend, dend+1))
+
+                        // delete the original AztecImageSpan
+                        text.delete(dend, dend+1)
+                        // now insert both the new insertion _and_ the original AztecImageSpan
+                        text.insert(dend, temp)
+                        temp = "" // discard the original source parameter as an ouput from this InputFilter
+                    }
+                }
+            }
+            temp
         }
 
         val emptyEditTextBackspaceDetector = InputFilter { source, start, end, dest, dstart, dend ->
@@ -462,7 +494,12 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             source
         }
 
-        filters = arrayOf(emptyEditTextBackspaceDetector)
+        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.O || Build.VERSION.SDK_INT == Build.VERSION_CODES.O_MR1) {
+            // dynamicLayoutCrashPreventer needs to be first in array as these are going to be chained when processed
+            filters = arrayOf(dynamicLayoutCrashPreventer, emptyEditTextBackspaceDetector)
+        } else {
+            filters = arrayOf(emptyEditTextBackspaceDetector)
+        }
     }
 
     private fun handleBackspaceAndEnter(event: KeyEvent): Boolean {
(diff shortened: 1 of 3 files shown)


malinajirka:
Great job, thanks for the PR @mzorz! I'm behind a schedule so I won't have time to review it today. I'll make sure to review it on Monday/Tuesday.

daniloercoli:
Hi @mzorz 👋 - Would you mind to delete the branch `issue/729-dymanic-layout-crash-preventer` on the remote site if it's not required?
There are chances that reviewers will pick it up, instead the correct one ;)

daniloercoli:
Quickly tested on Android 8 and it does fixes the problem, I need to test it more extensively though.

In the meanwhile, I noticed that once the image is deleted and re-inserted the "Media deleted"  Toast message " is shown twice on the screen. I guess the delete callback is called on the Media obj.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -440,12 +440,44 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
+        // problem is fixed at the Android OS level as described in the following url
+        // https://android-review.googlesource.com/c/platform/frameworks/base/+/634929
+        val dynamicLayoutCrashPreventer = InputFilter { source, start, end, dest, dstart, dend ->
+            var temp = source
+            if (dest.length > dend+1) {
+                // if there are any images right after the destination position, hack the text
+                val spans = dest.getSpans(dstart, dend+1, AztecImageSpan::class.java)
+                if (spans.isNotEmpty()) {
+                    // test: is this an insertion?
+                    if (dstart == dend) {
```
daniloercoli: Is it possible to move this check before ` val spans = dest.getSpans(dstart, dend+1, AztecImageSpan::class.java)` ?
It should improve performances a bit, considering that this filter is called at any key pressed.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -440,12 +440,44 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
+        // https://android-review.googlesource.com/c/platform/frameworks/base/+/634929
+        // https://github.com/wordpress-mobile/AztecEditor-Android/issues/729
+        // the rationale behind this workaround is that the specific crash happens only when adding/deleting text right
+        // before an AztecImageSpan, so we detect the specific case and re-create the contents only when that happens.
+        // This is indeed tackling the symptom rather than the actual problem, and should be removed once the real
+        // problem is fixed at the Android OS level as described in the following url
+        // https://android-review.googlesource.com/c/platform/frameworks/base/+/634929
+        val dynamicLayoutCrashPreventer = InputFilter { source, start, end, dest, dstart, dend ->
+            var temp = source
+            if (dest.length > dend+1) {
```
daniloercoli: Can we move this check ` if (dest.length > dend+1` as first line in the filter, and do not create the `temp` variable since it's not needed in that case?

mzorz:
> Would you mind to delete the branch issue/729-dymanic-layout-crash-preventer on the remote site if it's not required?
There are chances that reviewers will pick it up, instead the correct one ;)

oops sorry :D just deleted  it - also this:

> In the meanwhile, I noticed that once the image is deleted and re-inserted the "Media deleted" Toast message " is shown twice on the screen. I guess the delete callback is called on the Media obj.

hah, I forgot to copy the code I put in the first experimental branch, that takes care of this. I put the code back in 79e17406 🙇 - thanks for the heads up!

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mzorz: gotcha, nice improvement, made change as suggested in 1ff08ac1

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mzorz: It wasn't actually creating a new variable (the object referenced to by `temp` and `source` where the same), but nevertheless thanks to your comment I looked into the docs again, and made an adjustment in 59ff4c08 as per the docs (initializing temp with `null` by default so the original change gets accepted if we don't need to make any other changes here).

Then we only override it in the case we actually go forward with the text change, by making `temp = ""` in [this other line ](https://github.com/wordpress-mobile/AztecEditor-Android/pull/801/files#diff-7ca71cd1f009ab8438e0901205001b14R481), to actually discard the original change.

Ref: [InputFilter docs](https://developer.android.com/reference/android/text/InputFilter#filter(java.lang.CharSequence,%20int,%20int,%20android.text.Spanned,%20int,%20int))

mzorz:
> Nice work @mzorz 🔆 LGTM!

Thank you @daniloercoli 😄 ! Glad you think it's worth merging 🙇 

> Wondering if it's necessary to test this PR on wp-android, since the Calypso mode should be ON there. I guess there are no differences, but a quick check would be nice.

Good idea, I've tested it there before and found no problems, and just tested it again with latest commit and seems to be working alright as well :) - also I think Gutenberg won't be a problem as well given we're not using text and images combined in a single Aztec instance so, should be good to go.

Thanks for the review @daniloercoli !

malinajirka:
Thanks for the fix @mzorz! I've tested several different use cases and everything works like a charm, great job!!!

I'm still not sure `the Editor crashes when inserting anything right before an image` is the only case when the editor crashes, but it's definitely the only case we are able to reproduce. So it'll be interesting to see whether we fix all the crashes or just some. 

LGTM  :shipit:

PR #806, opened 2019-04-09
Remove some unneeded logging

mzorz:
Small PR that removes unneeded logging that convolutes the logs being sent to Fabric, sometimes occupying space that some other information would find useful.


mzorz:
Thank you @daniloercoli! 🙇

PR #807, opened 2019-04-13
Don't bail when ZWJ cleanup found end-of-text marker

hypest:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/845

This PR fixes an Aztec issue where the ZWJ cleanup function was bailing too early if the end-of-text marker was found. Instead, cleanup needs to continue and remove the rest of the ZWJs if any.

Gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/868

### To test
Use the gutenberg-mobile PR and follow the steps there, thanks!

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
@@ -335,8 +335,9 @@ class AztecParser @JvmOverloads constructor(val plugins: List<IAztecPlugin> = li
         do {
             lastIndex = text.lastIndexOf(Constants.ZWJ_CHAR, lastIndex)
             if (lastIndex == text.length - 1) {
-                // ZWJ at the end of text will serve as end-of-text marker so, let it be and finish.
-                return
+                // ZWJ at the end of text will serve as end-of-text marker so, let it be and continue cleaning up ZWJs
+                lastIndex--
+                continue
             }
 
             if (lastIndex > -1) {


daniloercoli:
Are there any details about the reason to remove the ZWJ character within the text?
It's used to "combine" other characters and display those as single char. If we remove it we're messing up the original text, no?

hypest:
> Are there any details about the reason to remove the ZWJ character within the text?
(I mean, not your PR, but why the original code was there).

Yeah, so, the Constants.ZWJ is used as a "temporary" character during the parsing process to make empty block elements have some content. See https://github.com/wordpress-mobile/AztecEditor-Android/blob/7e968c36fca6d93857a972d5cb308afde93a1a11/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt#L211-L217 We set the span with `SPAN_EXCLUSIVE_EXCLUSIVE` at that point and that span flag doesn't allow the span to be 0-length.

Regarding disrupting multi-sequence emojis, I guess we got lucky here. In fact, we got ourselves a wrong naming of the constant... we have is as ZWJ, but the accurate one is ZWS (Zero Width Space) and it's a different code point. The typical ZWJ is `0x200D` but the ZWS is `0x200B`. So, our `cleanupZWJ()` doesn't actually affect the emojis. 😌

daniloercoli:
What do you think about changing the `ZWJ` constant to `ZWS`?

hypest:
> What do you think about changing the name of the ZWJ constant to ZWS?

We should do it :nod:

PR #812, opened 2019-04-25
Fix list handling in Gutenberg

hypest:
### Fix
This PR changes the "Enter" character detection and the messages sent to the React Native side logic. This description will get expanded, WIP.

### Background of changes

1. The `onEnterKey` callback now delivers the text that the Aztec processor (TextWatcher) was holding when the "Enter" key was detected, the `firedAfterTextChanged` flag that shows whether the key has been detected before or after the text has been changed and `selStart`, `selEnd` for the position of the caret in that text.
2. Introduce new variants of some methods (`toHtml`, `toPlainHtml`) that accept a param with the text that they should work on.
3. Exposed `Azteclog.logSpansDetails()` so it can be used by the AndroidStudio debugger.
4. Removed the `EnterPressedWatcher` and moved it to the react-native-aztec project so we can supply it with props set via the RN layer.
5. Removed the `text.replace(start, start + 1, "")` line that was deleting the "Enter" key when it got detected. ~Instead, the new assumption is that the key will stay in the text and the client listening (gutenberg-mobile for instance) should try to update its internal representation to match Aztec's behavior.~ The new assumption is that the deletion will happen conditionally: it won't for the List Gutenberg block and it *will* for the Paragraph Gutenberg block. Governed by a prop (see point No4).

### Test
Follow test plans on https://github.com/wordpress-mobile/gutenberg-mobile/pull/928


hypest:
Chatted over Slack and agreed to merge this now.

Issue #516, opened 2017-11-06
Crash when applying special comments on API 26/27


0nko:
Applying a special comment (more, page) results in a crash on Oreo 8.1 (API 27)

### Reproduced
1. Start the demo app (in Calypso mode)
2. Tap on the _More_ button
3. Observe the crash

```
    FATAL EXCEPTION: main
    Process: org.wordpress.aztec, PID: 6582
    java.lang.ArrayIndexOutOfBoundsException: length=39; index=-1
    at android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:648)
    at android.widget.Editor.drawHardwareAccelerated(Editor.java:1703)
    at android.widget.Editor.onDraw(Editor.java:1672)
    at android.widget.TextView.onDraw(TextView.java:6882)
    at android.view.View.draw(View.java:19192)
    at android.view.View.updateDisplayListIfDirty(View.java:18142)
    at android.view.View.draw(View.java:18920)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4236)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4022)
    at android.view.View.updateDisplayListIfDirty(View.java:18133)
    at android.view.View.draw(View.java:18920)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4236)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4022)
    at android.view.View.draw(View.java:19195)
    at android.widget.ScrollView.draw(ScrollView.java:1739)
    at android.view.View.updateDisplayListIfDirty(View.java:18142)
    at android.view.View.draw(View.java:18920)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4236)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4022)
    at android.view.View.updateDisplayListIfDirty(View.java:18133)
    at android.view.View.draw(View.java:18920)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4236)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4022)
    at android.view.View.updateDisplayListIfDirty(View.java:18133)
    at android.view.View.draw(View.java:18920)
    at android.view.ViewGroup.drawChild(ViewGroup.java:4236)
    at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4022)
    at android.view.View.updateDisplayListIfDirty(View.java:18133)
[...]

PR #813, opened 2019-04-29
Fix DynamicLayout Crash - Round 2

mkevins:
### Fix
This PR is an attempt to address the issue described here: https://github.com/wordpress-mobile/WordPress-Android/issues/8828. This issue is related to #516 and #729 which were addressed with PR #801. This is a workaround to attempt to avoid encountering the underlying Android OS bug [described here](https://issuetracker.google.com/issues/67102093) and [patched here](https://android-review.googlesource.com/c/platform/frameworks/base/+/634929).

This workaround utilizes an `InputFilter` as in #801. The code in that PR covers many scenarios that lead to a crash, however, there are still some that cannot be addressed with that approach. One limitation of the InputFilter approach is that we may only modify the text being inserted during a change, but cannot modify the text being removed. Also, in that PR, strict criteria must be met before the fix is applied. One such criterion is that the selection must be a caret (i.e. no selection). This means certain edge cases will still encounter the crash.

One odd note is that if we set up the Demo App content with text, followed by two images, and then some more text, we can trigger the crash by deleting the text preceding the image _only after we have edited that text at least once_. Also, even _after_ having edited that text, if we used the `undo` and `redo` buttons to attain the same state, we could then delete the text without a crash.

Somehow, setting the content in full (as occurs when the app first loads) will put the app into a "safe" state. I also noticed that when the first change happens above the images, the images move vertically by a very small amount.

In order to make use of this in a new workaround, we can set the text within the `InputFilter`. This is accomplished by utilizing the `fromHtml` method, passing in the output from `toFormattedHtml`. In order to do so with the new state of the content, we must overload the `toFormattedHtml` method to accept a `Spannable`, whereas the default behavior is to generate the html from the unedited content (which would merely revert our changes, disallowing modification by the user).

The nature of this PR is to expand the number of cases it addresses by relaxing the criteria for which it is applied. This also means that any side effects that such a workaround may introduce have the potential to affect more users. Since an unrecoverable crash is a fairly severe condition, it holds some weight, but, some consideration should also be given to the volume of users that are currently not experiencing this crash. It would be helpful if we could test these changes as thoroughly as possible so ensure we do not introduce any regressions.

### Test
Follow steps here: https://github.com/wordpress-mobile/WordPress-Android/issues/8828#issuecomment-485843440.
Also, here: https://github.com/wordpress-mobile/AztecEditor-Android/issues/516#issue-271419928.
And here: https://github.com/wordpress-mobile/AztecEditor-Android/issues/729#issue-334435899.

Any more tests in those threads would also be good candidates for testing this PR.

### Review
@mzorz

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -455,29 +455,27 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         // https://android-review.googlesource.com/c/platform/frameworks/base/+/634929
         val dynamicLayoutCrashPreventer = InputFilter { source, start, end, dest, dstart, dend ->
             var temp : CharSequence? = null
-            if (!bypassCrashPreventerInputFilter
-                    && dstart == dend && dest.length > dend+1
-                    && source != Constants.NEWLINE_STRING) {
-                // dstart == dend means this is an insertion
-                // avoid handling anything if it's a newline
+            if (!bypassCrashPreventerInputFilter) {
+
                 // if there are any images right after the destination position, hack the text
-                val spans = dest.getSpans(dstart, dend+1, AztecImageSpan::class.java)
+                val spans = dest.getSpans(dend, dend+1, AztecImageSpan::class.java)
                 if (spans.isNotEmpty()) {
+
                     // prevent this filter from running twice when `text.insert()` gets called a few lines below
                     disableCrashPreventerInputFilter()
                     // disable MediaDeleted listener before operating on content
                     disableMediaDeletedListener()
 
-                    // take the source (that is, what is being inserted), and append the Image to it. We will delete
-                    // the original Image later so to not have a duplicate.
-                    // use Spannable to copy / keep the current spans
-                    temp = SpannableStringBuilder(source).append(dest.subSequence(dend, dend+1))
+                    // create a new Spannable to perform the text change here
+                    var newText = SpannableStringBuilder(dest.subSequence(0, dstart))
+                            .append(source.subSequence(start, end))
+                            .append(dest.subSequence(dend, dest.length));
 
-                    // delete the original AztecImageSpan
-                    text.delete(dend, dend+1)
-                    // now insert both the new insertion _and_ the original AztecImageSpan
-                    text.insert(dend, temp)
-                    temp = "" // discard the original source parameter as an ouput from this InputFilter
+                    // force a history update to ensure the change is recorded
+                    history.beforeTextChanged(this@AztecText)
+
+                    // use HTML from the new text to set the state of the editText directly
+                    fromHtml(toFormattedHtml(newText), false)
 
                     // re-enable MediaDeleted listener
                     enableMediaDeletedListener()
@@ -1577,8 +1575,11 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                     (max == length || (length == 1 && text.toString() == Constants.END_OF_BUFFER_MARKER_STRING))) {
                 setText(Constants.REPLACEMENT_MARKER_STRING)
             } else {
+                // prevent changes here from triggering the crash preventer
+                disableCrashPreventerInputFilter()
                 editable.delete(min, max)
                 editable.insert(min, Constants.REPLACEMENT_MARKER_STRING)
+                enableCrashPreventerInputFilter()
             }
 
             // don't let the pasted text be included in any existing style
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -36,7 +36,14 @@ import android.support.graphics.drawable.VectorDrawableCompat
 import android.support.v4.content.ContextCompat
 import android.support.v7.app.AlertDialog
 import android.support.v7.widget.AppCompatEditText
-import android.text.*
+import android.text.Editable
+import android.text.InputFilter
+import android.text.InputType
+import android.text.Spannable
+import android.text.SpannableStringBuilder
+import android.text.Spanned
+import android.text.TextUtils
+import android.text.TextWatcher
 import android.text.style.SuggestionSpan
 import android.util.AttributeSet
 import android.util.DisplayMetrics
@@ -469,7 +476,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                     // create a new Spannable to perform the text change here
                     var newText = SpannableStringBuilder(dest.subSequence(0, dstart))
                             .append(source.subSequence(start, end))
-                            .append(dest.subSequence(dend, dest.length));
+                            .append(dest.subSequence(dend, dest.length))
 
                     // force a history update to ensure the change is recorded
                     history.beforeTextChanged(this@AztecText)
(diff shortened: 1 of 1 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -462,29 +455,27 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
-            if (!bypassCrashPreventerInputFilter
-                    && dstart == dend && dest.length > dend+1
-                    && source != Constants.NEWLINE_STRING) {
-                // dstart == dend means this is an insertion
-                // avoid handling anything if it's a newline
+            if (!bypassCrashPreventerInputFilter) {
+
                 // if there are any images right after the destination position, hack the text
-                val spans = dest.getSpans(dstart, dend+1, AztecImageSpan::class.java)
+                val spans = dest.getSpans(dend, dend+1, AztecImageSpan::class.java)
```
mzorz: I think we still need to check `dest.length > dend+1` before this line so `getSpans` doesn't try to look off the limits of the buffer

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -462,29 +455,27 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
+                    // force a history update to ensure the change is recorded
+                    history.beforeTextChanged(this@AztecText)
 
-                    // delete the original AztecImageSpan
-                    text.delete(dend, dend+1)
-                    // now insert both the new insertion _and_ the original AztecImageSpan
-                    text.insert(dend, temp)
-                    temp = "" // discard the original source parameter as an ouput from this InputFilter
+                    // use HTML from the new text to set the state of the editText directly
+                    fromHtml(toFormattedHtml(newText), false)
```
mzorz: Just thinking aloud here - we discussed this of course but I'm not sure about doing such a radical thing for _everyone_ with Android 8.x just as you already rightfully commented on the PR description:

> Also, in that PR, strict criteria must be met before the fix is applied. One such criterion is that the selection must be a caret (i.e. no selection). This means certain edge cases will still encounter the crash.

We may want to narrow down the cases so this doesn't introduce a regression for other users who maybe weren't experiencing the problem so, thinking in that direction it seems to me it'd be nice if we kept the "case detection logic" (such as the big `if` clause that was up there) and add / remove more constraints as we see fit. I'm not really  convinced on one way or the other, so please take this comment just as to keep the conversation going and see what we can get from it rather than an indication / suggestion of  "what seems right" in code 👍 

OTOH it may be the only way to cover "all" the possibilities where the issue may appear. I believe we're still spot on in that Images are the ones internally using `DynamicLayout` so it seems right to do the search for it (as in, the broader case).
I'll do some more smoke tests, another thing that comes to mind is maybe we also need to try and keep the caret position as well? Though we may want to address that as a separate PR if it starts to get convoluted

mzorz:
I've been testing this on WPAndroid with the following PRs:
https://github.com/wordpress-mobile/WordPress-Android/pull/9773
https://github.com/wordpress-mobile/gutenberg-mobile/pull/947

With the content in this report `fabric-5a76a4fa8cb3c2fa634286c6`, session `fabric-5CC8BCF7015E00017DBC2EE3F6DE56FF_DNE_0_v2` (the logs show a Post with several images in it), and can observe there's a lag when positioning the cursor typing fast, with entered sequence being lost in there. After giving it time, you can re-type (if not right before an Image, that is) and it works without lag. Fortunately the caret position is kept so, it's no big deal - of course much better than crashing so, I think this is bearable.
Here's a video of it https://cloudup.com/coQJdVLvDXA

I've tried with other long posts and can see the lag Aztec already had before so, not a pointer that this solution in particular may be problematic (or should I say, be more problematic than it was before).

I think we should test a bit more but, generally, looks good as to try it! 

Let's fix the lint issues and we can try ship a new version of Aztec at next cycle, sounds good @mkevins ?

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mkevins: I believe the `getSpans` method automatically adjusts the query boundaries to the constraints of the underlying `Spannable`, but I could not find documentation to confirm that. I've added a commit to address it, just in case.

I modified the "fence-post" slightly, as I believe the previous condition stopped short of the last character, since the span positions are counted by the space between characters. E.g. in the text below, if "!" was the image character, and "**Hello** world" was the selection, we would have `dstart == 0`, `dend == 11`, and `dest.length == 12`. In that case, the condition `dest.length > dend + 1` would avoid the filter, though a single character follows the selection.

hello-world-spans

In my observations, I have seen that image (obj.) characters are followed by a `\n`, so this may be a moot point, but I am unsure if this is guaranteed to be true at all times that text is being modified.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mkevins: > and add / remove more constraints as we see fit

I totally agree with this. So far, these are the conditions that must be met:

1. Android API version must be either 26 or 27.
2. The crash preventer must not already be "in progress" (i.e. no recursion).
3. At least one character must follow the selection / caret.
4. The character immediately following the selection / caret must have at least one `AztecImageSpan` (i.e. it is an image).

Conditions 2 and 3 are set in the outer-most conditional of the filter. Condition 4 requires querying of the `Spannable`, and thus lives inside that outer conditional. If there is a way to determine whether we are on a version of Oreo which has been patched, we could narrow condition 1 further, but I'm not sure this is currently possible.

If you have other conditions that come to mind, we can add them. Note: I intentionally left out the condition regarding the new line character because this approach was different enough, I thought it best to test the behavior separately. Do you know if the current state reproduces (regresses) the issue that condition was added for?

mkevins: > another thing that comes to mind is maybe we also need to try and keep the caret position as well? Though we may want to address that as a separate PR if it starts to get convoluted

I noticed this on `develop` as well, so I think a separate PR would be better, with an issue to track it.

mkevins:
> I think we should test a bit more but, generally, looks good as to try it!
> 
> Let's fix the lint issues and we can try ship a new version of Aztec at next cycle, sounds good @mkevins ?

Sure, I've pushed a commit to fix the lint errors. Thanks for setting up the related PRs @mzorz !

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mzorz: > Note: I intentionally left out the condition regarding the new line character because this approach was different enough, I thought it best to test the behavior separately. Do you know if the current state reproduces (regresses) the issue that condition was added for?

I think you're right - just tested that the conditions this PR here https://github.com/wordpress-mobile/AztecEditor-Android/pull/809 was fixing are not met with this PR anymore and wasn't able to reproduce the crash so, I think we're good 👍

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mzorz: Ah, gotcha - you're right. I was assuming too much re: selection start, end and length. I understand it's ok then 👍

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
mzorz: hey @mkevins I had to re-introduce the check in f77e5764 (hope you don't mind!) to fix the connected tests that were failing for this PR 👍

mzorz:
We had @daniloercoli's eyes on this one as well and approved merging so, merging

daniloercoli:
Tested and worked as expected. Also tried  tests on this PR 🎉

PR #821, opened 2019-06-04
[GB-Mobile Integration] Fix for Backspace does not outdent nested list item

### LINKED ISSUE #1027: Backspace does not outdent nested list item on android
**Description:**
Pressing backspace when the caret is at the beginning of a nested list item does not "outdent" the list item. Rather, it merges the list item to the preceding list item.

**Steps to reproduce:**
1. Create a nested list
2. Place caret at the start of an indented list item
3. Press [Backspace]

**Expected behavior:**
The indented list item should outdent one level with each press of [Backspace], and only merge with the preceding item after it is fully outdented.

**Screencast:**

**Actual behavior:**
The indented list item immediately merges with the preceding list item.

**Screencast:**

daniloercoli:
This PR is the first step required to fix https://github.com/wordpress-mobile/gutenberg-mobile/issues/1027

@mkevins has reported the problem with backspace and nested lists, and already proposed a solution [here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1027#issuecomment-495557004). In this PR I started from scratch, but ended up with similar code.

Please, test this PR by pointing the experimental branch here https://github.com/wordpress-mobile/gutenberg-mobile/tree/experimental/fix-backspace-outdent-on-nested-lists to the full hash `fbf98c1af64552a01f1ae203d295b79b8362c629` of this branch.

- Test both the Google GBoard and the AOSP keyboard.
- Also test on real devices, and emulators by using the external OS keyboard as input (in order to simulate HW keyboard).

### Review
@mkevins 

Note: I run both kotlin unit tests, and Java connected tests on this PR without finding any issue.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -501,10 +501,11 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             if (isHandlingBackspaceEvent) {
                 // It's already handling a backspace event. Do nothing.
             } else if (end != 0 || start != 0) {
-                // Not a backspace event
+                // Not a text erase event
             } else if (selectionStart != selectionEnd) {
                 // there is something selected on the editor, let's make Aztec do the work
             } else if (onAztecKeyListener != null) {
+                // There is a listener set, let's use it
                 isHandlingBackspaceEvent = true
 
                 // Prevent the forced backspace from being added to the history stack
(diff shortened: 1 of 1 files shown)


mchowning:
👋 @daniloercoli ! I'm assuming we should still work to merge this PR since [the related issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1027#issuecomment-495557004) is still open. Do you agree @mkevins, @daniloercoli  ?

mkevins:
>  @daniloercoli ! I'm assuming we should still work to merge this PR since [the related issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1027#issuecomment-495557004) is still open. Do you agree @mkevins, @daniloercoli ?

Hi @mchowning :wave: :smile: 

I somehow missed the original ping and just saw this one, so my apologies for the long delay. I agree, this is worth pursuing. I'm not sure off-hand whether the proposed solution needs to be updated to be current with all the changes from then to now. Do you have an idea of the priority of the original issue?

PR #825, opened 2019-06-10
Android X migration

marecar3:
This PR migrates android support dependencies to [AndroidX](https://developer.android.com/jetpack/androidx)

It's used by [gutenberg-mobile](https://github.com/wordpress-mobile/gutenberg-mobile/pull/1112) PR


aztec/src/test/kotlin/org/wordpress/aztec/AztecParserTest.kt
```diff
@@ -92,7 +92,7 @@ class AztecParserTest : AndroidTestCase() {
      */
     @Before
     fun init() {
-        context = MockContext()
+        context = ShadowApplication.getInstance().applicationContext
```
marecar3: Hey @loremattei @jtreanor I did spend some time to figure out why MockContext() is crashing during initialization but didn't find a proper reason why that was happening. Instead, I have added `ShadowApplication.getInstance().applicationContext` which is working good.

cc: @hypest

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -170,7 +170,7 @@ open class MainActivity : AppCompatActivity(),
 
         private val isRunningTest: Boolean by lazy {
             try {
-                Class.forName("android.support.test.espresso.Espresso")
+                Class.forName("androidx.test.espresso.Espresso")
```
marecar3: Lost some time to figure out that this was the problem why tests are failing.
Probably we can find some other approach to check if we are testing or not so that other colleagues don't lose too much time on this.
@loremattei @jtreanor @hypest

marecar3:
Hey, @jtreanor should we increment also this line: https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/jitpack.yml#L2 to `28` in this PR?

marecar3:
Hey, @loremattei @jtreanor @hypest I have fixed all problems with tests which were reproducible on a local machine. 

After that, I have experienced problems with failing tests in `MixedTextFormattingTests`. Problem is that tests are failing randomly and I didn't have any luck to reproduce them locally. 

One solution was to add `threadSleep` in some places and it gave result, but it seems that tests can't be fixed with that approach.

Any ideas what could be the problem here? Thanks!

hypest:
It seems that even on PRs without code changes (Example: #827), the connected tests fail at the moment with timeouts. So, probably false negatives at this stage. I think it might make sense to revert the commits that add delays... I'll test that out and report back.

jtreanor:
@hypest @marecar3  You can increase the timeout for connected tests by updating [this line](https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/.circleci/config.yml#L61) to something like `15m`. Lets make sure these are running on CI before merging.

jtreanor:
> Hey, @jtreanor should we increment also this line: https://github.com/wordpress-mobile/AztecEditor-Android/blob/develop/jitpack.yml#L2 to 28 in this PR?

@marecar3 yes, that should be updated too.

hypest:
>  You can increase the timeout for connected tests by updating this line to something like 15m. Lets make sure these are running on CI before merging.

Thanks for the pointer @jtreanor ! OK, I'll try that but I'd say let's not block on this, especially since it doesn't seem to be specific to this PR anyway.

PR #826, opened 2019-06-22
Update gradle version to match latest stable recommended in Android Studio

Doko-Demo-Doa:
This PR aims to update Gradle wrapper to latest version being used by latest current stable Android Studio version.


maxme:
Thanks, but we'll bump sdk and build tool versions later.

PR #827, opened 2019-07-09
Update PR template

loremattei:
This PR updates the PR template to add a heads up about the translation for the new strings.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,6 +9,6 @@
 ### Review
 @[USER_NAME]
 
-Make sure strings can be translated:
+Make sure strings will be translated:
 
-- [ ] If there are new strings that have to be translated, I'm adding them to the client's `strings.xml` as a part of the integration PR.
+- [ ] If there are new strings that have to be translated, I have added them to the client's `strings.xml` as a part of the integration PR.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,3 +8,7 @@
 
 ### Review
 @[USER_NAME]
+
+Make sure strings can be translated:
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,4 +7,8 @@
 3. [STEP_3]
 
 ### Review
-@[USER_NAME]
\ No newline at end of file
+@[USER_NAME]
+
+Make sure strings can be translated:
```
hypest: Nit but, can we change the `can` to `will`? With the `can` it's not clear whether the author needs to do something special with the string formats perhaps, or something.

While at it, I understand that the copy here is probably intentionally generalized but I wonder, how about we mention the WordPress app specifically?

Suggestion: "Make sure strings will be translated by the WordPress Android app"

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,4 +7,8 @@
...
+
+- [ ] If there are new strings that have to be translated, I'm adding them to the client's `strings.xml` as a part of the integration PR.
```
hypest: What do you think about having it in past tense like the WPAndroid case?

Like: "If there are new strings that have to be translated, I have added them to the client's `strings.xml` as a part of the integration PR."

.github/PULL_REQUEST_TEMPLATE.md
loremattei: I updated it, but I omitted the last part because this is a library and could be used by other clients... wdyt?

hypest: Thanks for the updates!

I think that the whole prompt to update the client's `strings.xml` is very specific to the WordPress app anyway (other clients might have a different infrastructure for translations) so generalizing it might be confusing for the PR author. That said, I think this is not a blocker for the PR so, let's run with the current version 👍.

PR #835, opened 2019-07-31
Span overlapping on Chromebooks

marecar3:
### Summary: 
Text entered into a post below an image is misplaced after free-form resizing the window

Steps to reproduce:
1. Start the Demo Aztec application
2. Play a little bit with window resizing (resize the app couple of times)

Result: Span text will overlap span image

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecDynamicImageSpan.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecDynamicImageSpan.kt
@@ -80,9 +80,10 @@ abstract class AztecDynamicImageSpan(val context: Context, protected var imageDr
         if (sizeRect.width() > 0) {
             return sizeRect.width()
         } else {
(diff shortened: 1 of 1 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecDynamicImageSpan.kt
```diff
@@ -109,6 +116,14 @@ abstract class AztecDynamicImageSpan(val context: Context, protected var imageDr
             height = (width / aspectRatio).toInt()
         }
 
+        // Note: This is not a solution just a temp code
+        // to demonstrate that for some reason value 36 ( which is got
+        // from imageDrawable?.intrinsicHeight ) is causing overlap problem
+        // or I think it's causing :D
+        if (height == 36) {
```
marecar3: This is the magic number from the logs. I am not sure why `imageDrawable?.intrinsicHeight` at some point is returning this value and also not sure if it's the case only for emulator or general.

PR #838, opened 2019-08-01
Fix breaking layout on slow window resizing.

### LINKED ISSUE #837: Phantom image size increase on slow window resize
This issue got uncovered after #835 got merged.

When you slowly resize window sometimes image bounding box gets too large, which results in this (all the space at the bottom is the image box)
image

To test, slowly resize the window.

khaykov:
### Test
1. Open the Aztec demo app on Chromebook or Emulator
2. Slowly resize the window.
3. Make sure text layout is not breaking in weird ways.


khaykov:
@marecar3 Thanks for taking a look! I fixed the test, and ready for another round :)

marecar3:
Thanks @khaykov!
I think that everything is working as expected!

PR #840, opened 2019-08-02
Fixed cursor size on new paragraph line

### LINKED ISSUE #839: Cursor increases in size on the new empty line
### Expected
When you press Enter after typing a paragraph, the cursor will have a height appropriate for the text size.

### Actual
Cursor is huge, and then shrinks when you start typing
image1

khaykov:
### Fixes #839

This PR fixes the issue with cursor size when you press Enter after typing a text.

### Test
1. Set editor into Calypso mode.
2. Open demo app and clear content of the editor.
3. Type some text and press Enter
4. Notice that the cursor on a new line is of the correct size.


marecar3:
Btw, thanks for helping on this one!

PR #843, opened 2019-08-13
Add the ability to signal redo/undo events via History listener

daniloercoli:
This PR does add the ability to signal undo/redo events via History listener to host apps.

Context: It will be used in wp-android to bump analytics about undo/redo actions in the editor

### Test
1. Start the demo app
2. Add a breakpoint [here](https://github.com/wordpress-mobile/AztecEditor-Android/compare/add/undo-redo-events-to-historyListener?expand=1#diff-3af4b1228dfbdd8f17a66315f599bfc0R147)
3. Write something in the editor, and tap Undo. 
4. Make sure Undo is properly called


0nko:
Looks good! 👍

Issue #854, opened 2019-10-02
Can't properly insert text on Android 7


daniloercoli:
This is related to https://github.com/wordpress-mobile/AztecEditor-Android/issues/853 and it's very noticeable on my Nexus 5x running Android 7.1.

Video here: https://cloudup.com/c9B1MVYh9lz (I was trying to type "Hello There").

Introduced since `v1.3.29`. I'm kind of surprised that we've not caught it during tests, maybe it's a problem on my device only?  
I just tested 2 previous versions, and those worked fine on my testing device. (The other testing device of mine, running Android 9 with Samsung keyboard, is not affected by this problem).

cc @marecar3 @khaykov  @hypest

marecar3:
Thanks, @daniloercoli!
I will see with @khaykov where could be the problem.

marecar3:
It seems that both this and #853 are exposed because of this change: https://github.com/wordpress-mobile/AztecEditor-Android/pull/838/files#diff-7ca71cd1f009ab8438e0901205001b14R455-R465

marecar3:
>Introduced since v1.3.29. I'm kind of surprised that we've not caught it during tests
Hey @daniloercoli, as a reviewer on that PR, I am taking full responsibility for those two bugs.

I was testing with the less Post `content` which test case didn't expose this kind of bug.
Now tested on Nexus 5x, 8.1 Android, and manage to reproduce it with full example content.

daniloercoli:
>> Introduced since v1.3.29. I'm kind of surprised that we've not caught it during tests

>> Hey @daniloercoli, as a reviewer on that PR, I am taking full responsibility for those two bugs.

I think the responsibility should be shared among all of us, not only you @marecar3 , since we all are using the editor, and it's now included in GB mobile. 

Btw, I encourage you all to run `app->tests` on your PRs, since also tests started failing in .29. That would have been an indicator that something was going off.

marecar3:
>Btw, I encourage you all to run app->tests on your PRs, since also tests started failing in .29. That would have been an indicator that something was going off.

Not sure why tests passed on this PR with the last commit: https://github.com/wordpress-mobile/AztecEditor-Android/pull/838/commits ?

designsimply:
[...]

Issue #853, opened 2019-10-02
Keeping Backspace pressed doesn't delete all the text


daniloercoli:
### Expected
To be able to keep Backspace pressed on the virtual keyboard and be able to keep deleing text.

### Observed
See the video on my nexus 5x: https://cloudup.com/cPoCbHUTMr4

- On a Nexus 5x with `Google Keyboard`
The caret stops almost immediately and I see the following in the log
```
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: beginBatchEdit on inactive InputConnection
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: beginBatchEdit on inactive InputConnection
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: finishComposingText on inactive InputConnection
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: deleteSurroundingText on inactive InputConnection
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: endBatchEdit on inactive InputConnection
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: setComposingRegion on inactive InputConnection
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: endBatchEdit on inactive InputConnection
2019-10-02 17:08:16.356 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: beginBatchEdit on inactive InputConnection
2019-10-02 17:08:16.357 30234-30234/org.wordpress.aztec W/IInputConnectionWrapper: setComposingText on inactive InputConnection

```

- On a Samsung S9 with `Samsung Keyboard`
The caret stops at the end of the previous line and `sendKeyEvent on inactive InputConnection` is available in the log. No video available atm, but I can make it.
```
2019-10-02 16:53:48.260 28980-28980/org.wordpress.aztec D/InputMethodManager: prepareNavigationBarInfo() DecorView@ef89c9a[MainActivity]
2019-10-02 16:53:48.260 28980-28980/org.wordpress.aztec D/InputMethodManager: getNavigationBarColor() -855310
[...]

PR #855, opened 2019-10-03
Issue/fix 854 cant properly insert text on android 7

marecar3:
### Fix https://github.com/wordpress-mobile/AztecEditor-Android/issues/854 and #853 

This quick fix is revert of: https://github.com/wordpress-mobile/AztecEditor-Android/pull/838
as I think that at this point the bug in the editor has higher priority then fix for Chromebook device.

### Test
Device: Nexus 5x Android 7 or Android 8

1. Open AztecEditor app
2. Make sure that you have full test content (which is provided by default)
3. Put the cursor on e.g. 'Heading 2' text and try deleting it with backspace or adding a new text

Expected result: Backspace should work as expected and there shouldn't be any lagging/problems when adding a new text.

### Review
@daniloercoli @khaykov 

### DIFF
--- a/aztec/src/test/kotlin/org/wordpress/aztec/AztecToolbarTest.kt
+++ b/aztec/src/test/kotlin/org/wordpress/aztec/AztecToolbarTest.kt
@@ -648,7 +648,7 @@ class AztecToolbarTest {
     @Throws(Exception::class)
     fun emptySelection() {
         editText.fromHtml("<b>bold</b><b><i>italic</i></b>")
-        editText.fromHtml("", false)
+        editText.fromHtml("")
 
         Assert.assertTrue(TestUtils.safeEmpty(editText))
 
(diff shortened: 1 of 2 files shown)


daniloercoli:
Tested on Nexus 5x Android 7.1, and Samsung S9 Android 9. 
Connected tests finished without problems.

:shipit:

aztec/src/test/kotlin/org/wordpress/aztec/AztecToolbarTest.kt
```diff
@@ -648,7 +648,7 @@ class AztecToolbarTest {
     @Throws(Exception::class)
     fun emptySelection() {
         editText.fromHtml("<b>bold</b><b><i>italic</i></b>")
-        editText.fromHtml("", false)
+        editText.fromHtml("")
```
daniloercoli: I see the previous code was using `setText` here. https://github.com/wordpress-mobile/AztecEditor-Android/pull/838/files#diff-0767c9a83c90683bf4b3661b324d2ad8L651

PR #858, opened 2019-10-04
release/1.3.30.1 (hotfix)

daniloercoli:
This is an hotfix that will be included in wp-android 13.3 to fix the delay and lag on Aztec.

Make sure the delay is not present anymore on demo content on Android 8 devices.


daniloercoli:
Thanks @malinajirka and @hypest for the time spent testing this PR. We've decided to not hot-fixing Aztec, because the fix above doesn't make big enough improvement to the Editor. The same fix will land to end users through the next Beta cycle as normal.

PR #861, opened 2019-10-16
[WP-Android 8828] - Detect ArrayIndexOutOfBoundsException exception

daniloercoli:
This PR does add the ability to detect the `ArrayIndexOutOfBoundsException` that may happen on Android 8 at the following line of code `android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:646)` due to a bug present in the Android Platform.
When this exact exception is detected, Aztec reports it to the parent app, by wrapping it into a custom Exception.

The reporting to the parent app happens by re-using the external logging capabilities of Aztec, instead of adding another layer of communication between Aztec and the parent app.
The parent app should just check the type of the exception in the "external" logger implementation passed to Aztec.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
@@ -47,10 +47,17 @@ class AztecExceptionHandler(private val logHelper: ExceptionHandlerHelper?, priv
 
         // Detect ArrayIndexOutOfBoundsException on Android 8, and report it to the parent app
         // See: https://github.com/wordpress-mobile/WordPress-Android/issues/8828
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && Build.VERSION.SDK_INT <= Build.VERSION_CODES.O_MR1
-                && ex is ArrayIndexOutOfBoundsException) {
+        if (ex is ArrayIndexOutOfBoundsException) {
             val stackTrace = Log.getStackTraceString(ex)
-            if (stackTrace.contains("android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:646)")) {
+            var detected = false
+            if (Build.VERSION.SDK_INT == Build.VERSION_CODES.O &&
+                    stackTrace.contains("android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:646)")) {
+                detected = true
+            } else if (Build.VERSION.SDK_INT == Build.VERSION_CODES.O_MR1 &&
+                    stackTrace.contains("android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:648)")) {
+                detected = true
+            }
+            if (detected) {
                 visualEditor.externalLogger?.logException(DynamicLayoutGetBlockIndexOutOfBoundsException("Error #8828", ex))
             }
         }
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
@@ -3,7 +3,7 @@ package org.wordpress.aztec
 import android.os.Build
 import android.util.Log
 import org.wordpress.android.util.AppLog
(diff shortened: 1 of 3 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
```diff
@@ -42,6 +45,16 @@ class AztecExceptionHandler(private val logHelper: ExceptionHandlerHelper?, priv
             }
         }
 
+        // Detect ArrayIndexOutOfBoundsException on Android 8, and report it to the parent app
+        // See: https://github.com/wordpress-mobile/WordPress-Android/issues/8828
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && Build.VERSION.SDK_INT <= Build.VERSION_CODES.O_MR1
+                && ex is ArrayIndexOutOfBoundsException) {
+            val stackTrace = Log.getStackTraceString(ex)
+            if (stackTrace.contains("android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:646)")) {
```
mzorz: Nice checking for the line number string here. It makes it fine-grained so we avoid having false positives treated mistakingly 👍 
Also, was wondering whether `stackTrace` could be null, so double checked `getStackTraceString()` and it seems it will never return `null` as per [this](https://developer.android.com/reference/android/util/Log.html#getStackTraceString(java.lang.Throwable)) so 👍 as well.

aztec/src/main/kotlin/org/wordpress/aztec/util/DynamicLayoutGetBlockIndexOutOfBoundsException.kt
```diff
@@ -0,0 +1,5 @@
+package org.wordpress.aztec.util
```
mzorz: What do you think about moving this class to a package of its own like `org.wordpress.aztec.exception` or `org.wordpress.aztec.util.exception`?

mzorz:
FYI just wanted to leave a note that this is easily reproduceable (so the code in this PR can be tested) by disabling the `dynamicLayoutCrashPreventer` InputFilter like this:

Comment out lines 528 in AztecText.kt like this:
```
//        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.O || Build.VERSION.SDK_INT == Build.VERSION_CODES.O_MR1) {
//            // dynamicLayoutCrashPreventer needs to be first in array as these are going to be chained when processed
//            filters = arrayOf(dynamicLayoutCrashPreventer, emptyEditTextBackspaceDetector)
//        } else {
            filters = arrayOf(emptyEditTextBackspaceDetector)
//        }
```

Only leave `filters = arrayOf(emptyEditTextBackspaceDetector)` (line 532) uncommented.

I've tested it and confirmed the crash is captured by the code in this branch and the call to the external logger is correctly made 👍

aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
mzorz: I think I detected an issue - this line will be different on Android 8.1 (API 27), see this https://github.com/wordpress-mobile/AztecEditor-Android/issues/516

it will be `java.lang.ArrayIndexOutOfBoundsException: length=39; index=-1
    at android.text.DynamicLayout.getBlockIndex(DynamicLayout.java:648)`

If you follow the steps to reproduce in the comment below on Android 8.1 (API 27) it will fail the string comparison @daniloercoli  - can we add this other string for the case where `Build.VERSION.SDK_INT == Build.VERSION_CODES.O_MR1`?

daniloercoli: Nice catch! I've updated the code in 86a000026a6de29e178e0fa4901c0d68e3adb30d

Tried to be more verbose as possibile to let future devs to understand quickly why there are those if cases.

mzorz: Awesome! Thank you 🙇

jrcacd:
How can we catch the ArrayIndexOutOfBoundsException exception?

PR #866, opened 2019-10-23
[GB-Mobile]  SetSpan ends beyond length

daniloercoli:
This PR does fix a GB-mobile problem (https://github.com/wordpress-mobile/WordPress-Android/issues/9832) where AztecWrapper tries to set the selection out of the bounds.

More details about the problem are in the comment here: https://github.com/wordpress-mobile/WordPress-Android/issues/9832#issuecomment-545564526 together with testings steps.

In this PR i've just disabled cleaning utils over the HTML passed as input, since the "logic" should all be JS side. Otherwise there could be a miss-match between passed HTML and what is "stored" inside Aztec.

GB-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1490
WP-Android PR: wordpress-mobile/WordPress-Android#10672

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt
@@ -75,8 +75,8 @@ class AztecParser @JvmOverloads constructor(val plugins: List<IAztecPlugin> = li
         return spanned
     }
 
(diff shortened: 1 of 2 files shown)


daniloercoli:
Closing this PR since it introduced more problems than it fixes.
We will fix the problem JS side on GB-mobile.

PR #867, opened 2019-10-30
Add simple tests for BR tags before P closing tag

daniloercoli:
Add tests to make sure the underlying parser is working as expected when one or more BR tag(s) are available in the HTML code right before the Paragraph closing tag.

Currently, this html `<p>Hello There!<br></p>` after being parsed to span and back to html will become `<p>Hello There!</p>`.
This is not a bug, this is how we originally implemented the function that cleans the HTML input in [AztecParser->tidy](https://github.com/wordpress-mobile/AztecEditor-Android/blob/b1fad439d56fa6d4aa0b78526fef355c59d00dd3/aztec/src/main/kotlin/org/wordpress/aztec/AztecParser.kt#L650) method.
Since we're using this editor in Gutenberg Mobile project, where the selection could be sent from the JS side to the native, we needed to take in consideration this behavior of Aztec in GB-mobile, and modify the logic that does set the selection accordingly.

Ref GB mobile PR: https://github.com/WordPress/gutenberg/pull/18138

### DIFF
--- a/aztec/src/test/kotlin/org/wordpress/aztec/AztecParserTest.kt
+++ b/aztec/src/test/kotlin/org/wordpress/aztec/AztecParserTest.kt
@@ -1164,6 +1164,38 @@ class AztecParserTest : AndroidTestCase() {
         Assert.assertEquals(input, output)
     }
 
+    /**
+     * Currently, this html <p>Hello There!<br></p> after being parsed to span and back to html will become
+     * <p>Hello There!</p>.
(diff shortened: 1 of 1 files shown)


hypest:
I will update this PR from develop, just to make sure that all tests are green.

hypest:
@daniloercoli has switched to other tasks these days so, I'll go ahead and merge this PR.

PR #873, opened 2019-11-20
Add default values to vertical space

maxme:
Use defaults for `block_vertical_padding` and `heading_vertical_padding` from `dimens.xml` like it's done for other parameters (like `spacing_extra` and `spacing_multiplier`)

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -388,8 +388,11 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
 
         commentsVisible = styles.getBoolean(R.styleable.AztecText_commentsVisible, commentsVisible)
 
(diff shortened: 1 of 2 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -388,8 +388,11 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
 
         commentsVisible = styles.getBoolean(R.styleable.AztecText_commentsVisible, commentsVisible)
 
-        verticalParagraphMargin = styles.getDimensionPixelSize(R.styleable.AztecText_blockVerticalPadding, 0)
-        verticalHeadingMargin = styles.getDimensionPixelSize(R.styleable.AztecText_headingVerticalPadding, 0)
+
```
theck13:
```suggestion
```

It looks like the continuous build server is complaining about this extra line.  Removing it should get the build to pass.

PR #875, opened 2019-11-28
Update glide to the latest version

anitaa1990:
### Fix
Updated the Glide version to 4.10.0 from 3.7.0. There have been a lot of changes in the v4 version of Glide. I followed this guide [here](https://bumptech.github.io/glide/doc/migrating.html#using-modelloader-streammodelloader) to help with the migration.

### Testing
- Run AztecEditor Demo app and verify that photos are loaded and displayed correctly. 
- Run AztecEditor Demo app and verify that videos are loaded and video thumbnail is displayed correctly. 
- Added the commit hash from this PR to the WordPress app to verify that images and videos are loaded correctly.

### Testing
@0nko 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         kotlinCoroutinesVersion = '1.1.0'
         supportLibVersion = '27.1.1'
         tagSoupVersion = '1.2.1'
-        glideVersion = '3.7.0'
+        glideVersion = '4.10.0'
         picassoVersion = '2.5.2'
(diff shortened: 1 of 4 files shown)


anitaa1990:
Thanks for the quick review @0nko!  

> the existing images from the demo text were never displayed

I'm not able to reproduce this. I tested in emulators 23, 26, 28 and 29. Could this be related to the device you are testing on? Or maybe the api version? Are you facing this issue in multiple devices? Would you be able to let me know the device version?

PR #878, opened 2019-11-29
Resolved clipboard issue on Android 7.0

jd-alexander:
### Fix
https://github.com/wordpress-mobile/WordPress-Android/issues/10872

This solution shows a warning instead of crashing when clipboard functionality is utilized on Samsung devices running Android 7.0.

### Test
Use testing instructions as described [here](https://github.com/wordpress-mobile/WordPress-Android/issues/8827#issuecomment-453113443)

**Note**

This hasn't been tested on a device as yet as I was unable to secure a Samsung device with the specifications and doing it using an emulator wasn't causing the same result. 

### Review
@malinajirka 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -1559,8 +1559,9 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                     deleteInlineStyleFromTheBeginning()
                 }
             }
-            // Fix for crash when pasting text on Samsung Devices running Android 8 & 7.
-            // Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/8827
+            // Fix for crash when pasting text on Samsung Devices running Android 7 & 8.
+            // Android 7 Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/10872
+            // Android 8 Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/8827
             clipboardIdentifier -> {
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && Build.VERSION.SDK_INT < Build.VERSION_CODES.P
                         && Build.MANUFACTURER.toLowerCase().equals("samsung")) {
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -1562,7 +1562,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             // Fix for crash when pasting text on Samsung Devices running Android 8 & 7.
             // Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/8827
             clipboardIdentifier -> {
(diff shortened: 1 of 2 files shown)


jd-alexander:
I saw some other members utilizing https://www.browserstack.com/ for testing. I am thinking to create a WP Android branch with this as an artifact so that an APK can be generated that would make testing possible on those platforms.

aztec/src/main/res/values/strings.xml
```diff
@@ -2,7 +2,7 @@
 
     <!-- GENERAL -->
     <string name="cursor_moved">Cursor moved</string>
-    <string name="samsung_disabled_custom_clipboard">Sorry, this feature is disabled on Android 8. Please use the Paste action instead.</string>
+    <string name="samsung_disabled_custom_clipboard">Sorry, this feature is disabled on Android %1$s. Please use the Paste action instead.</string>
```
malinajirka: @aforcier I'm not 100% sure so I'll better ask. Consider the following scenario
1. `samsung_disabled_custom_clipboard` is created
2. It gets translated into Spanish
3. It gets released
4. We update the string in English, but we don't translated it yet
5. It gets released
6. What version of the string will be used on a Spanish device?

I'm asking, because if we use version from step `2.` the app will crash on `context.getString(R.string.samsung_disabled_custom_clipboard, Build.VERSION.RELEASE)` as the original string doesn't contain any placeholder.

malinajirka: ` If there are more arguments than format specifiers, extra arguments are ignored.`
So it seems the app won't crash after all, but I'm curious how the translations work anyway:).

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1559,13 +1559,13 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                     deleteInlineStyleFromTheBeginning()
                 }
             }
-            // Fix for crash when pasting text on Samsung Devices running Android 8.
+            // Fix for crash when pasting text on Samsung Devices running Android 8 & 7.
             // Ref: https://github.com/wordpress-mobile/WordPress-Android/issues/8827
```
malinajirka: We might want to consider adding a link to the new issue - https://github.com/wordpress-mobile/WordPress-Android/issues/10872. Wdyt?

jd-alexander: Yes, that's a good point. Will do!

jd-alexander: Resolved in af1e734 Let me know if this works. I was gonna convert this to a multiline comment but didn't see any other usages in this manner outside of for methods so just left it with the same format.

aztec/src/main/res/values/strings.xml
aforcier: You know, I'm not sure 🤔 it should depend on whether the GlotPress job updates translation files or completely replaces them each time, I think I once knew but forgot. @loremattei do you know the answer to this one? 🙂

loremattei: Sure! So, when the new English string is uploaded to GlotPress during code freeze, the translation is invalidated. When we download the strings two weeks later to build the final binary, if the updated string has not not been translated yet, it's removed from the Spanish (or whatever) strings.xml and the app will show the English version. 

It works that the translated files are completely replaced by the newest ones, so untranslated strings are not in them.

malinajirka: Thank you for the clarification Lorenzo! ;)

PR #879, opened 2019-11-29
Implement start and reversed in lists

marecar3:
This PR adds visual support for the reversed and start properties in list.

- [ ] Update tests with new test cases

It does three things:
 - Add to TextList paragraph attributes properties for reversed and start
 - Adds support for parsing of the attributes when reading the HTML
 - Adds support to show visually the attributes effect on display

list

To test:
 - Start the demo app
 - Check that all  lists are showed correctly, specially the new ones with reversed and start attributes.
 - Play around on HTML by changing the attributes of list to see if all is displayed correctly
- Test scenarios
 - Ordered list without start and reversed
 - Ordered list with reversed attribute: check if numbers are updated correctly has members are added
 - Ordered list with start attributed: check if numbers start on the correct number indicated
 - Ordered list with start and reversed: check if numbers are correct
 - In all cases delete lines and see if updates are done correctly.

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -115,6 +115,9 @@ open class MainActivity : AppCompatActivity(),
                 "    <li>Nine</li>\n" +
                 "    <li>Eight</li>\n" +
                 "</ol>"
+        private val ORDERED_REVERSED_WITH_START_IDENT = "<h4>Reversed Start in 4 List:</h4>" +
+                "<ol reversed>" +
+                "<li>dkfdskfdf\\sd\\</li><li>dfds</li><li>fdf</li><li>dfd</li><li>f</li><li>dsfs<ol><li>fdfd</li><li>f</li><li>dsf</li><li>ds</li><li>d</li><li>g</li><li>sdg</li><li>ds</li></ol></li></ol>\n"
         private val LINE = "<hr />"
         private val UNORDERED = "<ul><li style=\"color:darkred\">Unordered</li><li>Should not have color</li></ul>"
         private val QUOTE = "<blockquote>Quote</blockquote>"
@@ -175,6 +178,7 @@ open class MainActivity : AppCompatActivity(),
                         ORDERED_REVERSED +
                         ORDERED_REVERSED_WITH_START +
                         ORDERED_REVERSED_NEGATIVE_WITH_START +
+                        ORDERED_REVERSED_WITH_START_IDENT +
                         LINE +
                         UNORDERED +
                         QUOTE +
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -109,15 +109,28 @@ open class MainActivity : AppCompatActivity(),
                 "    <li>Nine</li>\n" +
                 "    <li>Eight</li>\n" +
                 "</ol>"
-        private val ORDERED_REVERSED_NEGATIVE_WITH_START = "<h4>Reversed Start in 10 List:</h4>" +
+        private val ORDERED_REVERSED_NEGATIVE_WITH_START = "<h4>Reversed Start in 1 List:</h4>" +
                 "<ol reversed start=\"1\">\n" +
-                "    <li>Ten</li>\n" +
-                "    <li>Nine</li>\n" +
-                "    <li>Eight</li>\n" +
+                "    <li>One</li>\n" +
+                "    <li>Zero</li>\n" +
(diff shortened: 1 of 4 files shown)


hypest:
👋 @marecar3 , let's add automated unit tests to exercise the new features in this PR, thanks!

marecar3:
> 👋 @marecar3 , let's add automated unit tests to exercise the new features in this PR, thanks!

Yup, I already added a subtask in the description for that purpose.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecOrderedListSpan.kt
```diff
@@ -55,7 +55,22 @@ class AztecOrderedListSpan(
...
+        var lineIndex = if (start > 0) {
+            if (isReversed) start - (getIndexOfProcessedLine(text, end) - 1)
+            else start + (getIndexOfProcessedLine(text, end) - 1)
+        } else {
+            val number = getNumberOfItemsInProcessedLine(text)
+            if (isReversed) number - (getIndexOfProcessedLine(text, end))
+            else getIndexOfProcessedLine(text, end)
+        }
+
         val textToDraw = if (lineIndex > -1) {
```
maxme: I think we could remove this check. I tested on the Web and negative number are shown. Here is the diff:

```diff
-        val textToDraw = if (lineIndex > -1) {
-            if (dir >= 0) lineIndex.toString() + "."
+        val textToDraw = if (dir >= 0) lineIndex.toString() + "."
             else "." + lineIndex.toString()
-        } else {
-            ""
-        }
```

This is kind of weird if we only show the empty string as the prefix:

Screenshot 2019-12-02 at 11 53 44

hypest:
> Writing unit tests will be hard for this one as we're only changing the drawText function.

True. What I had in mind were tests for at least making sure that the attributes are preserved after some trivial editing and serializing back to html.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecOrderedListSpan.kt
marecar3: Hey @maxme, nice catch there. 
I have fixed it now and you can try when you have time :)

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -91,6 +91,33 @@ open class MainActivity : AppCompatActivity(),
...
+                "</ol>"
+        private val ORDERED_REVERSED_NEGATIVE_WITH_START = "<h4>Reversed Start in 10 List:</h4>" +
+                "<ol reversed start=\"1\">\n" +
+                "    <li>Ten</li>\n" +
+                "    <li>Nine</li>\n" +
+                "    <li>Eight</li>\n" +
+                "</ol>"
+        private val ORDERED_REVERSED_WITH_START_IDENT = "<h4>Reversed Start in 4 List:</h4>" +
+                "<ol reversed>" +
+                "<li>dkfdskfdf\\sd\\</li><li>dfds</li><li>fdf</li><li>dfd</li><li>f</li><li>dsfs<ol><li>fdfd</li><li>f</li><li>dsf</li><li>ds</li><li>d</li><li>g</li><li>sdg</li><li>ds</li></ol></li></ol>\n"
```
maxme: Looks good but can we make this a bit prettier 😅

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -91,6 +91,33 @@ open class MainActivity : AppCompatActivity(),
...
+                "    <li>Two</li>\n" +
+                "    <li>One</li>\n" +
+                "</ol>"
+        private val ORDERED_REVERSED_WITH_START = "<h4>Reversed Start in 10 List:</h4>" +
+                "<ol reversed start=\"10\">\n" +
+                "    <li>Ten</li>\n" +
+                "    <li>Nine</li>\n" +
+                "    <li>Eight</li>\n" +
+                "</ol>"
+        private val ORDERED_REVERSED_NEGATIVE_WITH_START = "<h4>Reversed Start in 10 List:</h4>" +
```
maxme: Nitpicking: We should change this to "Reversed Start in **1** List"

marecar3: Yup, you are right! 
Fixed.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
marecar3: ahahahaha! 
you are 100% right!

maxme:
Looks good, :shipit:

PR #882, opened 2019-12-05
Revert "Fix/ime iob fix"

mzorz:
Reverts wordpress-mobile/AztecEditor-Android#834

According to what has been observed (no significant contribution to lowering the number of occurrences of the crash observed in https://github.com/wordpress-mobile/WordPress-Android/issues/9904), the patch is being reverted to avoid introducing this unneeded complexity.


mzorz:
Checked a few paths after merge conflict resolution, seems to work well and tests passed, merging

PR #883, opened 2019-12-06
Accessibility : Improve TalkBack behavior of the formatting toolbar

jd-alexander:
### Fix
 https://github.com/wordpress-mobile/WordPress-Android/issues/10893

As described in the issue, there are some of the toolbar actions that get rendered as `ToggleButtons` but behave as a `Button`. An example of this is the button that is used to extend or collapse the media toolbar. Its state is announced as checked or not checked unnecessarily. This PR seeks to address all the buttons that exist in the toolbar that are not expected to toggle it's checked state but are used to prompt another action such as a popup window. 
  
### Test
1. Run the Aztec demo app with TalkBack enabled.
2. Hover actions such as Heading, List, Link, and Media Collapse/Expand.
3. Listen to or see the TalkBack announcement and verify the properties of the action are being read as that of a button since that's the primary behavior and no toggling is actually taking place.

### Other thoughts. 
1. To verify that the right buttons are actually being modified. i.e ensure the selected buttons are appropriate and point out any that are missing. 

2. There's one button, The More Button I wasn't able to add to the list of targets to be modified because it actually exists in comments project within Aztec so I am not able to reference it in the logic that determines which buttons should be modified. Any thoughts on this are welcome. 
https://github.com/wordpress-mobile/AztecEditor-Android/blob/ad68cc9b640e7f50d7e21e7e459083b45fa1a7d7/wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/toolbar/MoreToolbarButton.kt#L20
After discovering this, an approach that came to mind would be to modify the interface that governs the `ToolBarAction` so the `AztecToolbar` wouldn't have to know which buttons should get this behavior that would be determined at the design level for the button. 

3. I chose this solution as it seems like the approach that's least intrusive and would make the least modification to the business logic that powers the editing toolbar. Also modifying the info node of the element before passing it back to `TalkBack` allows it to still continue to process all the properties to give the best experience to the user. 

### Review
@shiki @malinajirka 


jd-alexander:
Thanks for this review @khaykov 🙇 

> We remove ACTION_LONG_CLICK from some of the buttons, but not the other. What is the logic behind this?

I didn't remove it from the others because the accessibility behavior that's attached to them is that of the `ToggleButton` and for the target buttons in this PR I was trying to mimic the `Button` behavior which doesn't have a long click. That was my rationale 😅 

> just leave it to developers to implement it in inflateButton method.  

This is a good idea! It's within the same project so I just did the implementation myself.

wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/toolbar/PageToolbarButton.kt
```diff
@@ -49,7 +51,9 @@ class PageToolbarButton(val visualEditor: AztecText) : IToolbarButton {
     }
 
     override fun inflateButton(parent: ViewGroup) {
-        LayoutInflater.from(context).inflate(R.layout.page_button, parent)
+        val rootView = LayoutInflater.from(context).inflate(R.layout.page_button, parent)
+        val button = rootView.findViewById<ToggleButton>(R.id.format_bar_button_page)
```
jd-alexander: A quick question @khaykov I was wondering if just putting the logic here is breaking the single responsibility principle. I was gonna put this in another function but it seemed a bit trivial. Nonetheless, let me know so I can adhere to code standards 🙇

khaykov: I see your point, but at the same time, we are kind of limited by past architectural decisions :) In my opinion, `inflateButton` can include the code related to making the button ready for usage (which includes setting up accessibility). We can change the architecture, but it might get tricky from there, and considering our eventual move to block editor, I think this is an acceptable approach.

khaykov:
> > We remove ACTION_LONG_CLICK from some of the buttons, but not the other. What is the logic behind this?
> 
> I didn't remove it from the others because the accessibility behavior that's attached to them is that of the ToggleButton and for the target buttons in this PR I was trying to mimic the Button behavior which doesn't have a long click. That was my rationale 😅

Behavior wise all the toolbar buttons have long-press actions (that displays hint). I'm not sure how helpful it is for accessibility since the button name is announced already, but for consistency, I think we should either leave it be for all the buttons or remove it from all of them :) What do you think?

jd-alexander:
> Behavior wise all the toolbar buttons have long-press actions (that displays hint). I'm not sure how helpful it is for accessibility since the button name is announced already, but for consistency, I think we should either leave it be for all the buttons or remove it from all of them :) What do you think?

That's a good point. I think it's best to leave it for consistency.

wordpress-comments/src/main/java/org/wordpress/aztec/plugins/wpcomments/toolbar/PageToolbarButton.kt
jd-alexander: Okay great. Glad we are on the same page 😄

PR #398, opened 2017-06-29
Fixing toolbar highlight issues

khaykov:
#395 need's to be merged before this one.

Fixes #345
Fixes #357 
Preserves selected style when deleting the last character on 0 index, and allows the selected style to be removed with extra backspace.

### Test 1

1. Open the demo app.
2. Remove the demo content.
3. Close the device keyboard. (Note that the cursor remains in the text view.)
4. Tap a format toolbar button for inline style (e.g. Bold) to enable it.
5. Tap back into the text view. 
6. Confirm that previously selected style in the toolbar is still highlighted.
7. Type text. Confirm that it's correctly styled.

### Test 2

1. Open the demo app.
2. Remove the demo content.
3. Tap a format toolbar button for inline style(e.g. Bold) to enable it.
4. Type text and press enter.
5. Deselect style and type text.
6. Notice that the style button is not highlighted and style is not applied.

### Test 3

1. Open the demo app.
2. Remove the demo content.
3. Tap a format toolbar button for inline style(e.g. Bold) to enable it.
4. Type text.
5. Press backspace to delete all the text.
6. Confirm that selected style is still highlighted.
7. Press backspace.
8. Confirm that selected style is not highlighted.

There are a couple of changes to inline style handling, so it would be awesome if both @0nko and @theck13 could play with inline styles a bit to make sure everything is ok.

Issue #362, opened 2017-06-08
Deleting All Text Containing Inline Style with Backspace Leaves Empty Tags


theck13:
### Expected
Deleting all text from the visual view using the backspace key deletes all text and formatting from the HTML view.

### Observed
Deleting all text from the visual view using the backspace key leaves empty tags of the inline style(s) applied to the last deleted text.

### Reproduced
1. Enter "none bold none italic none strike" into the demonstration app.
2. Select "bold" text.
3. Tap ***Bold*** format button.
4. Select "italic" text.
5. Tap ***Italic*** format button.
6. Select "strike" text.
7. Tap ***Strike*** format button.
8. Place cursor at end of text.
9. Tap ***Backspace*** key until text is gone.
10. Tap ***HTML*** format button.
11. Notice `<b></b>` tags remain.

#### Tested
Google Pixel on Android 7.1.2 with AztecDemo 1.0

khaykov:
Fixed in #398

Issue #357, opened 2017-05-30
Inline style in toolbar erroneously highlighted 


khaykov:
### Reproduced
1. Load editor without any text.
2. Press Bold button in toolbar.
3. Type somethign and press enter.
4. Deselect Bold and type somethign.
5. Notice that the Bold button is highlighted (but not applied, as expected).

#### Tested
Genymotion on Nexus 6P (7.0.0) and Nesux 5X (6.0.0)

PR #892, opened 2020-02-07
Last char is when the length is now zero

hypest:
Experimental PR to fix the gutenberg-mobile issue: https://github.com/wordpress-mobile/gutenberg-mobile/issues/1865

### From https://github.com/wordpress-mobile/AztecEditor-Android/pull/398:

Test 1 (testing the format while the keyboard is closed): ✅
Test 2 (format on the next line): ❌ This test seems to fail anyway even on `develop` (d2baefd3512e3c2e19a6b5a183befb4c8f108949) so, ignoring for this PR
Test 3 (Backspace last char, backspace to clear format): ✅

### From https://github.com/wordpress-mobile/AztecEditor-Android/issues/357:

Test: ⚠️ issue not happening but, the format did not continue to the next line as the test steps mention.

### From https://github.com/wordpress-mobile/AztecEditor-Android/issues/362:

Test: ✅

## To test
Using the gutenberg-mobile PR https://github.com/wordpress-mobile/gutenberg-mobile/pull/1867 confirm that the issue of merging two paragraph blocks (https://github.com/wordpress-mobile/gutenberg-mobile/issues/1865) is not happening anymore.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/watchers/ZeroIndexContentWatcher.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/watchers/ZeroIndexContentWatcher.kt
@@ -27,7 +27,10 @@ class ZeroIndexContentWatcher(aztecText: AztecText) : TextWatcher {
 
         val aztecText = aztecTextRef.get()
         // last character was removed
-        if (aztecText != null && textChangedEventDetails.inputEnd == 0 && textChangedEventDetails.inputStart == 1) {
+        if (aztecText != null
+                && text.length == 0
+                && textChangedEventDetails.inputEnd == 0
+                && textChangedEventDetails.inputStart == 1) {
             aztecText.disableOnSelectionListener()
         }
     }


hypest:
The unit tests seem to fail on CI (all of them, which is weird) but they are green locally.

cameronvoell:
Confirmed this changes fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1865 on the basic case, however I found one way to still have merge on delete not work, when you add a style to the text that extends farther to the right from what you delete at the front of the text before the attempted merge. (the merge on delete never happens in the attached GIF, don't be fooled by GIF restart :-P )

I think the remaining issue would be pretty rare, and this fix is still an improvement, so we could document the remaining issue, and still merge this fix. What do you think @hypest ?

styled-text-no-merge-on-delete

hypest:
Thanks Cameron, yeah, we could tackle the subcase in a separate PR, but since I'm not even 100% about the current fix, let's also hear from @khaykov about the fix, since there might be more context.

hypest:
Since we need this fix on the v1.22 release of the block editor, let's move forward, wdyt @cameronvoell ?

mchowning:
Going ahead and merging since the tests are passing locally and this is "blocking" the 1.22 block editor release. 🥁

PR #895, opened 2020-02-20
Add a condition that prevents invalid span bounds

0nko:
Prevents wordpress-mobile/WordPress-Android#11309 (and possibly a few others) by checking for < 0 and > text.length span bounds. If it's detected the state is logged and the style is not applied.


hypest:
All connected tests fail, most certainly not because of any change in this PR. The fails happen on develop anyway. I'll go ahead and merge this PR.

Issue #910, opened 2020-05-25
BackgroundColorSpan support?


arctouch-felipevaladares:
Does the lib support BackgroundColorSpan? i need to create a custom toolbar and one of the options should be to paint the text background like this:
![image](https://user-images.githubusercontent.com/59581302/82846196-10242300-9ebe-11ea-8c47-fddbd6be1b02.png)

felipevaladares:
i did by my self on this PR https://github.com/wordpress-mobile/AztecEditor-Android/pull/911

PR #911, opened 2020-05-26
Added backgroundColorSpan support

felipevaladares:
- Added CssBackgroundColorPlugin based on CssUnderlinePlugin
- Added AztecBackgroundColorSpan
- General changes to implement and use BackgroundColorSpan

### Test
1. Run the app
2. Select some text
3. Click on the new button that is located between Bold and Quote

### Review
@koke @maxme @bummytime @shiki @Tug 

device-2020-05-26-184604

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
@@ -2,7 +2,6 @@ package org.wordpress.aztec.formatting
 
 import android.graphics.Typeface
 import android.text.Spanned
-import android.text.style.BackgroundColorSpan
 import android.text.style.StyleSpan
 import org.wordpress.android.util.AppLog
 import org.wordpress.aztec.AztecAttributes
@@ -116,6 +115,11 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
             return
         }
 
+        if (textFormat == AztecTextFormat.FORMAT_BACKGROUND){
+            //clear previous background before applying a new one to avoid problems when using multiple bg colors
+            removeBackgroundInSelection(selectionStart, selectionEnd)
+        }
+
         var precedingSpan: IAztecInlineSpan? = null
         var followingSpan: IAztecInlineSpan? = null
 
@@ -134,10 +138,10 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
 
                 if (spanEnd > start) {
                     // ensure css style is applied
-                    (precedingSpan as IAztecInlineSpan).applyInlineStyleAttributes(editableText, start, end)
+                    spanToApply.applyInlineStyleAttributes(editableText, start, end)
                     return@applyInlineStyle // we are adding text inside span - no need to do anything special
                 } else {
-                    applySpan(precedingSpan as IAztecInlineSpan, spanStart, end, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+                    applySpan(spanToApply, spanStart, end, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
                 }
             }
         }
@@ -153,8 +157,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
 
             if (followingSpan != null) {
                 val spanEnd = editableText.getSpanEnd(followingSpan)
-                applySpan(followingSpan as IAztecInlineSpan, start, spanEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
-                editableText.setSpan(followingSpan, start, spanEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+                applySpan(spanToApply, start, spanEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
             }
         }
 
@@ -177,8 +180,25 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
 
             applySpan(spanToApply, start, end, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
         }
-
         joinStyleSpans(start, end)
+    }
+
+    private fun removeBackgroundInSelection(selStart: Int, selEnd: Int) {
+        val spans = editableText.getSpans(selStart, selEnd, AztecBackgroundColorSpan::class.java)
+        spans.forEach { span ->
+            if (span != null) {
+                val currentSpanStart = editableText.getSpanStart(span)
+                val currentSpanEnd = editableText.getSpanEnd(span)
+                val color = span.backgroundColor
+                editableText.removeSpan(span)
+                if (selEnd < currentSpanEnd) {
+                    editableText.setSpan(AztecBackgroundColorSpan(color), selEnd, currentSpanEnd, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+                }
+                if (selStart > currentSpanStart) {
+                    editableText.setSpan(AztecBackgroundColorSpan(color), currentSpanStart, selStart, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
+                }
+            }
+        }
     }
 
     private fun applySpan(span: IAztecInlineSpan, start: Int, end: Int, type: Int) {
@@ -218,6 +238,11 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
     fun removeInlineStyle(spanToRemove: IAztecInlineSpan, start: Int = selectionStart, end: Int = selectionEnd) {
         val textFormat = spanToTextFormat(spanToRemove) ?: return
 
+        if (textFormat == AztecTextFormat.FORMAT_BACKGROUND){
+           // removeBackgroundInSelection(start, end)
+         //   return
+        }
+
         val spans = editableText.getSpans(start, end, IAztecInlineSpan::class.java)
         val list = ArrayList<AztecPart>()
 
@@ -252,7 +277,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
             return firstSpan.style == secondSpan.style
         }
         // special check for BackgroundSpan
-        if (firstSpan is BackgroundColorSpan && secondSpan is BackgroundColorSpan) {
+        if (firstSpan is AztecBackgroundColorSpan && secondSpan is AztecBackgroundColorSpan) {
             return firstSpan.backgroundColor == secondSpan.backgroundColor
         }
 
@@ -261,6 +286,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
 
     // TODO: Check if there is more efficient way to tidy spans
     fun joinStyleSpans(start: Int, end: Int) {
+
         // joins spans on the left
         if (start > 1) {
             val spansInSelection = editableText.getSpans(start, end, IAztecInlineSpan::class.java)
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
@@ -115,7 +115,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
             return
         }
 
-        if (textFormat == AztecTextFormat.FORMAT_BACKGROUND){
+        if (textFormat == AztecTextFormat.FORMAT_BACKGROUND) {
             //clear previous background before applying a new one to avoid problems when using multiple bg colors
             removeBackgroundInSelection(selectionStart, selectionEnd)
         }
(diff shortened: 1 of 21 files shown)


duyvt88:
@felipevaladares Hi, can we change the text color of the selected text?
Many thanks!

felipevaladares:
> 
> 
> @felipevaladares Hi, can we change the text color of the selected text?
> Many thanks!

This is to change the background color, not the text color, but you can use both

felipevaladares:
the idea is to use in cases like the image on this https://github.com/wordpress-mobile/AztecEditor-Android/issues/910

duyvt88:
> > @felipevaladares Hi, can we change the text color of the selected text?
> > Many thanks!
> 
> This is to change the background color, not the text color, but you can use both

Thanks for your reply. I'm trying to customize your code to change the text color.

felipevaladares:
> 
> 
> > > @felipevaladares Hi, can we change the text color of the selected text?
> > > Many thanks!
> > 
> > 
> > This is to change the background color, not the text color, but you can use both
> 
> Thanks for your reply. I'm trying to customize your code to change the text color.

I'm not sure if you will be able to do that using my code, you should probably create a new Span and Plugin based on my changes, i believe it will be very similar

duyvt88:
> > > > @felipevaladares Hi, can we change the text color of the selected text?
> > > > Many thanks!
> > > 
> > > 
> > > This is to change the background color, not the text color, but you can use both
> > 
> > 
> > Thanks for your reply. I'm trying to customize your code to change the text color.
> 
> I'm not sure if you will be able to do that using my code, you should probably create a new Span and Plugin based on my changes, i believe it will be very similar

I have done it. Thanks for your support.
image

Do you think we can change the font size of the selected text?
Thanks,

felipevaladares:
> 
> 
> > > > > @felipevaladares Hi, can we change the text color of the selected text?
> > > > > Many thanks!
> > > > 
> > > > 
> > > > This is to change the background color, not the text color, but you can use both
> > > 
> > > 
> > > Thanks for your reply. I'm trying to customize your code to change the text color.
> > 
> > 
> > I'm not sure if you will be able to do that using my code, you should probably create a new Span and Plugin based on my changes, i believe it will be very similar
> 
> I have done it. Thanks for your support.
> image
> 
> Do you think we can change the font size of the selected text?
> Thanks,

Nice, i'm curious to see your code, have you opened a PR?

felipevaladares:
@koke @maxme @bummytime @shiki @Tug could you guys please review this PR?

cameronvoell (changes requested):
Nice work here @felipevaladares. Changes seem on the whole consistent with the existing code base, so nice job with that. However, I did notice a few minor issues that would be good to fix before merging this in. 

1. While testing your PR, I noticed an existing issue that becomes more noticeable with your changes: https://github.com/wordpress-mobile/AztecEditor-Android/issues/914 . If you toggle text with background color bold or italic etc., and then toggle back off, the background color disappears unexpectedly. Ideally this is fixed before merging.

2. In the past we had other issues where certain text patterns did not show a visual for highlighting, or show a cursor when it is placed in the text. I think we are seeing similar issues with text with a backgroundColorSpan, highlighting shows no color change, and the cursor is not visible. Perhaps previous fixes for these types of issues will provide hints on how to fix that. (https://github.com/wordpress-mobile/AztecEditor-Android/pull/244 , https://github.com/wordpress-mobile/AztecEditor-Android/pull/249)

3. Perhaps for at least this first iteration we can add Background Color button to the Plugin buttons section of the toolbar, so that it is easy to not include the button if we do not add the new CssBackgroundColorPlugin. This way existing apps using Aztec will not have the new button show up unless they want to use it. (For example, notice if you comment this line https://github.com/wordpress-mobile/AztecEditor-Android/blob/a1386beed8d8215fe8ab79c501c1ec237ae8a785/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L443 the More Button will disappear from the toolbar, but this is not the case if we remove your `aztec.addPlugin(CssBackgroundColorPlugin())` because it is in the main layout sections of simple and advanced toolbars instead of the plugin section.)

Thanks again for the changes here.

felipevaladares:
> 
> 
> Nice work here @felipevaladares. Changes seem on the whole consistent with the existing code base, so nice job with that. However, I did notice a few minor issues that would be good to fix before merging this in.
> 
>     1. While testing your PR, I noticed an existing issue that becomes more noticeable with your changes: #914 . If you toggle text with background color bold or italic etc., and then toggle back off, the background color disappears unexpectedly. Ideally this is fixed before merging.
> 
>     2. In the past we had other issues where certain text patterns did not show a visual for highlighting, or show a cursor when it is placed in the text. I think we are seeing similar issues with text with a backgroundColorSpan, highlighting shows no color change, and the cursor is not visible. Perhaps previous fixes for these types of issues will provide hints on how to fix that. (#244 , #249)
> 
>     3. Perhaps for at least this first iteration we can add Background Color button to the Plugin buttons section of the toolbar, so that it is easy to not include the button if we do not add the new CssBackgroundColorPlugin. This way existing apps using Aztec will not have the new button show up unless they want to use it. (For example, notice if you comment this line https://github.com/wordpress-mobile/AztecEditor-Android/blob/a1386beed8d8215fe8ab79c501c1ec237ae8a785/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L443
>         the More Button will disappear from the toolbar, but this is not the case if we remove your `aztec.addPlugin(CssBackgroundColorPlugin())` because it is in the main layout sections of simple and advanced toolbars instead of the plugin section.)
> 
> 
> Thanks again for the changes here.

Hi @cameronvoell thanks for reviewing, i'll try to handle all the issues and improvements as soon as possible :)

duyvt88:
> > > > > > @felipevaladares Hi, can we change the text color of the selected text?
> > > > > > Many thanks!
> > > > > 
> > > > > 
> > > > > This is to change the background color, not the text color, but you can use both
> > > > 
> > > > 
> > > > Thanks for your reply. I'm trying to customize your code to change the text color.
> > > 
> > > 
> > > I'm not sure if you will be able to do that using my code, you should probably create a new Span and Plugin based on my changes, i believe it will be very similar
> > 
> > 
> > I have done it. Thanks for your support.
> > image
> > Do you think we can change the font size of the selected text?
> > Thanks,
> 
> Nice, i'm curious to see your code, have you opened a PR?

Hi @felipevaladares,
Sorry for replying late. Please check my code here https://github.com/duyvt88/AztecEditorCustom.

Thank you very much!

arctouch-felipevaladares:
> 
> 
> Nice work here @felipevaladares. Changes seem on the whole consistent with the existing code base, so nice job with that. However, I did notice a few minor issues that would be good to fix before merging this in.
> 
>     1. While testing your PR, I noticed an existing issue that becomes more noticeable with your changes: #914 . If you toggle text with background color bold or italic etc., and then toggle back off, the background color disappears unexpectedly. Ideally this is fixed before merging.
> 
>     2. In the past we had other issues where certain text patterns did not show a visual for highlighting, or show a cursor when it is placed in the text. I think we are seeing similar issues with text with a backgroundColorSpan, highlighting shows no color change, and the cursor is not visible. Perhaps previous fixes for these types of issues will provide hints on how to fix that. (#244 , #249)
> 
>     3. Perhaps for at least this first iteration we can add Background Color button to the Plugin buttons section of the toolbar, so that it is easy to not include the button if we do not add the new CssBackgroundColorPlugin. This way existing apps using Aztec will not have the new button show up unless they want to use it. (For example, notice if you comment this line https://github.com/wordpress-mobile/AztecEditor-Android/blob/a1386beed8d8215fe8ab79c501c1ec237ae8a785/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L443
>         the More Button will disappear from the toolbar, but this is not the case if we remove your `aztec.addPlugin(CssBackgroundColorPlugin())` because it is in the main layout sections of simple and advanced toolbars instead of the plugin section.)
> 
> 
> Thanks again for the changes here.

@cameronvoell can you verify if the 1st and 3th items are ok? I believe i fixed the #914 also i removed the button from the toolbar and changed to be a plugin, still trying to figure out the second item(highlighting problem)

arctouch-felipevaladares:
@cameronvoell highlight problem fixed 👍

arctouch-felipevaladares:
@cameronvoell did you managed to review again? i really need this changes to be merged asap

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -238,13 +262,6 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
         joinStyleSpans(start, end)
     }
 
-    fun removeInlineCssStyle(start: Int = selectionStart, end: Int = selectionEnd) {
```
cameronvoell, line 241: I am a little suspicious that removing this code might have side effects, though I'm not sure what they are yet. Also this does not seem to fix all cases of https://github.com/wordpress-mobile/AztecEditor-Android/issues/914. For example, if you highlight the first red italic text in the sample text and turn off italic, and then open html you will see the color attribute is gone, and returning from html the color will be gone in the display text. 

I did see that background color no longer disappears when you remove a text format so that is an improvement. Perhaps we leave this code as it was since it does not affect background, and it does not seem to completely fix https://github.com/wordpress-mobile/AztecEditor-Android/issues/914 either.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -221,8 +247,6 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
                 editableText.removeSpan(it)
             }
         }
-        // remove the CSS style span
-        removeInlineCssStyle()
```
cameronvoell, line 225: Suggest we keep this for now since it does not affect background color, see previous https://github.com/wordpress-mobile/AztecEditor-Android/pull/911/files#r463370943

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecBackgroundColorSpan.kt
```diff
@@ -0,0 +1,24 @@
+package org.wordpress.aztec.spans
+
+import android.graphics.Color
+import android.text.TextPaint
+import android.text.style.BackgroundColorSpan
+import org.wordpress.aztec.AztecAttributes
+
+class AztecBackgroundColorSpan(
+        val color: Int,
+        val alpha: Int = 220,
```
cameronvoell, line 10: I think only `color` is ever used as a constructor parameter from here. The other parameters can probably be moved to be class properties. What do you think?

cameronvoell:
Hello @arctouch-felipevaladares , thanks for the updates, and sorry for the delay with the review. 
> @cameronvoell highlight problem fixed 👍

Nice fix here. 

> @cameronvoell can you verify if the 1st and 3th items are ok? I believe i fixed the #914 also i removed the button from the toolbar and changed to be a plugin

I left some comments about the 1st item, along with a few other comments on the code. The 3rd item for making the background button functionality an optional plugin is looking good as well 👍 

I did find one other issue that is related to the highlight state of the Background button. It looks like after highlighting some text and then navigating elsewhere the button stays highlighted. The button should function like the other buttons in that it is highlighted only when placed above or highlighting text that has that property, and then it should automatically go to inactive state when moved elsewhere (see bold or italic button for example).

maxme:
Just a note that we're changing the project license, since this haven't landed yet it's not a problem, but if the patch land after the switch, it will be done under the MPL v2.

Ref. https://github.com/wordpress-mobile/AztecEditor-Android/pull/922

yuvalgnessin-qz:
Hi @felipevaladares :) do you think you might be able to respond to @cameronvoell's comments in the near term? If not, I might be interested in taking over this effort. Let me know!

felipevaladares:
> 
> 
> Hi @felipevaladares :) do you think you might be able to respond to @cameronvoell's comments in the near term? If not, I might be interested in taking over this effort. Let me know!

Hi @yuvalgnessin-qz i will do that this week :)

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
felipevaladares: This code is weird, it just remove everything without any rules, not sure if it will break something, maybe it's a work around to something? it really doesn't make any sense to me

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
felipevaladares: ok, if really doesn't affect bg color i think we can keep, but its something to take a look in the future, that method doesn't make much sense

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecBackgroundColorSpan.kt
felipevaladares: the color is used for the super class, BackgroundColorSpan(color) constructor

felipevaladares: no problem, i'll take a look into this

felipevaladares:
> 
> 
> Just a note that we're changing the project license, since this haven't landed yet it's not a problem, but if the patch land after the switch, it will be done under the MPL v2.
> 
> Ref. #922

No problem, i agree

starshipcoder:
I suggest the background icon take the current background color, so moving cursor could modify this color too

And it will be ready for a color picker

yuvalgnessin-qz:
Hello @cameronvoell and others! Since @felipevaladares was not available to address the changes you requested, I have opened a new PR #934 implemented by @lvcasasanta to replace this one. It addresses all the changes you have requested. 

Please take a look as soon as you can. Thank you!

SiobhyB:
I'm going ahead to close this PR, as it's been superseded by https://github.com/wordpress-mobile/AztecEditor-Android/pull/1041.

PR #916, opened 2020-06-22
Add possibility to use a custom layout for the toolbar

starshipcoder:
Permits to ui client to change icon order, hide icon, use 2 lines...

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
@@ -55,6 +55,7 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
     private var sourceEditor: SourceViewEditText? = null
     private var dialogShortcuts: AlertDialog? = null
     private var isAdvanced: Boolean = false
+    private var hasCustomLayout: Boolean = false
     private var isMediaToolbarAvailable: Boolean = false
     private var isExpanded: Boolean = false
     private var isMediaToolbarVisible: Boolean = false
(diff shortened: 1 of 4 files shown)


maxme:
Just a note that we're changing the project license, since this haven't landed yet it's not a problem, but if the patch land after the switch, it will be done under the MPL v2.

Ref. https://github.com/wordpress-mobile/AztecEditor-Android/pull/922

starshipcoder:
No problem for me, it is even better!

maxme:
@starshipcoder that works fine, but the logic is a bit weird to me. What about keeping the `advanced` option and add a `customLayout` attribute that, if set, would override the `advanced` option. 

I think @anitaa1990 worked on that, what's your opinion about this one?

starshipcoder:
You are right ! I will do that. Thank you for the review

anitaa1990:
> @starshipcoder that works fine, but the logic is a bit weird to me. What about keeping the advanced option and add a customLayout attribute that, if set, would override the advanced option.
I think @anitaa1990 worked on that, what's your opinion about this one?

That sounds like a good idea to me!

dimanych:
Please 👍 and merge this pr )

starshipcoder:
I apply @maxme suggestion, would you mind review the PR ?

oguzkocer:
@anitaa1990 It looks like this PR might have slipped through the cracks, could you take a look at it when you get a chance?

anitaa1990:
I can take a look at this one today 👍

anitaa1990 (approved):
This looks good! Thank you for working on this @starshipcoder. :shipit:

anitaa1990:
@oguzkocer there are a few checks failing in the PR (related to Buildkite). Is there something I need to do (maybe create the same branch `customize_toolbar_layout` in order to reinitiate the checks again?

oguzkocer:
@anitaa1990 The CI checks don't work in forked PRs, a separate PR needs to be created to run the checks. I'll send you a Slack message with a link to a guide for this.

PR #919, opened 2020-07-21
Update strikethrough to default to <s> tag

cameronvoell:
Fixes part of https://github.com/wordpress-mobile/gutenberg-mobile/issues/729

### Description
This change updates the strikethrough button to default to `<s> `tag instead of `<del>` tag. This brings AztecEditor-Android inline with [AztecEditor-iOS](https://github.com/wordpress-mobile/AztecEditor-iOS) and [HTML5](https://html.com/tags/s/#:~:text=is%20used%20to%20strike,that%20is%20no%20longer%20correct.) standards. 

### To Test
Case 1:
1. Press Strikethrough button and add some text, note strikethrough works correctly
2. Switch to HTML mode, note that Aztec shows `<s></s>` tag around strikethrough text where it would previously have a `<del>` tag

Case 2:
1. Switch to HTML mode, add some text with a `<del></del>` tag around it (or update text with `<s></s>` around it to be `<del></del>` instead).  
2. Switch back to visual mode and note that text is still displayed as crossed out, as expected.


mchowning (approved):
Tested and everything works well. Thanks @cameronvoell ! 👍

Issue #917, opened 2020-07-10
Contributor Agreement for re-licensing Aztec-Android from GPL v2 to MPL v2.0


maxme:
[WordPress-Mobile organization](https://github.com/wordpress-mobile) is relicensing Aztec-Android code base from [GNU General Public License, v2 (GPL v2)](https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html) to [Mozilla Public License v2.0 (MPL v2.0)](https://www.mozilla.org/en-US/MPL/2.0/). In this ticket, we're asking all contributors to give their permission for this change.

# Why Switching from the GPL v2?

The GPL v2.0 license is a blocker for distributing Aztec-Android in proprietary mobile apps or in open source apps that are not GPL-compatible. Rich text editor libraries in the mobile space are lacking. There is no well known open source rich text editor for Android and we believe that Aztec-Android could be used more and get more contributions with a less restrictive license than the GPL v2.

# Why the MPL v2.0?

- It's less restrictive than the GPL and LGPL. MPL licensed software can be combined with other material if the files are separated (see paragraph 1.7 `Larger Work` and 3.3 `Distribution of a Larger Work` in the MPL v2.0). This allows proprietary mobile apps to integrate Aztec-Android.
- It's a copyleft license, like the GPL, every derivative work must be licensed under the same license. Unlike the BSD or MIT licenses, no one can distribute the code under different terms. It forces anyone modifying Aztec-Android. to contribute back with their changes (bug fix, new feature, etc.).
- It's compatible with the GPL v2 (see paragraph 1.12 `Secondary License` and 3.3. `Distribution of a Larger Work` in the MPL v2.0), so integration into WordPress and the WordPress apps won't be a problem.

# What happen if I don't agree?

We'll remove or rewrite your contributions so that the relicensing can go forward while respecting your copyright.

# What does this change cover?

The relicensing covers the code in Aztec-Android repository: https://github.com/wordpress-mobile/AztecEditor-Android

# What happen with previous versions of Aztec-Android?

Aztec-Android versions prior to the license switch will remain under the GPL v2.

# Contributors

If you're in the list below, you contributed to Aztec-Android as an external contributor. We ask you to post the following comment on this ticket if you agree with the change:

```
I agree that my past contributions to Aztec-Android are relicensed from
the GNU General Public License v2 to the Mozilla Public License version 2.0.
```

[...]

PR #922, opened 2020-08-03
Change the license from GPL v2 to MPL v2

maxme:
All contributors agreed for the license switch in the ticket https://github.com/wordpress-mobile/AztecEditor-Android/issues/917. We're now changing the license from GPL v2 to MPL v2.


hypest (approved):
Looks good to me! Looking forward to the new contributions this change can enable! :shipit:

PR #927, opened 2020-09-15
Accessibility Improvement : Remove focusing of AztecText control as it's unneeded in TalkBack mode.

### LINKED ISSUE #926: Accessibility : AztecText focuses on hover
### Expected
That the user has to tap the AztecText control for it to gain focus when using TalkBack.

Parent issue https://github.com/wordpress-mobile/gutenberg-mobile/issues/918

### Observed
Once the user hovers on the AztecText control when TalkBack is on, the view is focused. 

### Reproduced
1. Enable TalkBack.
2. Hover over the example content within the AztecText view.
3. See the blinking cursor. 

Tested on Google Pixel 4 XL, Android 10, WPAndroid `develop` ()

jd-alexander:
Fixes #926 
## Findings
The `AztecText` view was focusing on hover during accessibility mode. 

## Solution
The solution was to modify the `AztecTextAccessibilityDelegate` that was intercepting the `MotionEvent` on hover since that triggered the delegate's logic which involved requesting the focus on the `AztecText` view. 

## Testing

1. Open the example app. 
2. Enable TalkBack by long-pressing both the volume up & down key at the same time. 
3. Once it's enabled, hover over the example content of the `AztecText` view.
I modified the string content from `EXAMPLE` to `QUOTE` so testing would be easier. 
https://github.com/wordpress-mobile/AztecEditor-Android/blob/f65933fa6b597b29a8f6552e7b524c79bc165ccd/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L462
4. TalkBack should read out the entire content, but the view should not be focused until you double-tap it. 
5. Before TalkBack would say long-press the control as the first option but now it says tap as seen below.
6. Ensure that tapping actually causes the cursor to become visible since the view is no longer going to be focused automatically.

Before | After 
--------|-------
<kbd></kbd|       <kbd></kbd>

## Screenshot
Before | After 
--------|-------
<kbd></kbd>    |       <kbd></kbd>

## Video
Before | After 
--------|-------
 [Cursor blinking automatically](https://drive.google.com/file/d/1S-PTepDF_bXwueXfZzSyuCqCtfXA2rrL/view?usp=sharing)       | [Tap to activate cursor](https://drive.google.com/file/d/1RysRKR3exXv_dNA-McGngJGGGyL22YJH/view?usp=sharing) 

## Reviewing 
Only 1 reviewer is needed but anyone can review. 

## Submitter Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.
- [x] I have considered adding accessibility improvements for my changes.
- [x] If it's feasible, I have added unit tests. 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextAccessibilityDelegate.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextAccessibilityDelegate.kt
@@ -62,7 +62,6 @@ class AztecTextAccessibilityDelegate(private val aztecText: EditText) {
     private fun announceLine(lineOffset: Int) {
         if (!aztecText.isFocused || (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP && !aztecText.isAccessibilityFocused)) {
             aztecText.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)
-            aztecText.requestFocus()
         } else {
             val announcement = getTextAtLine(lineOffset).replace(Constants.IMG_STRING, mediaItemContentDescription)
             accessibilityManager.interrupt()


mchowning (approved):
Tested and working well. Looks good! Nice work @jd-alexander ! 🎉 

I also went through the test steps for https://github.com/wordpress-mobile/AztecEditor-Android/pull/646, which is the PR adding the line you're removing, and I found no issues. :shipit:

jd-alexander:
Oh nice. Thanks @mchowning 🙇

PR #940, opened 2021-10-07
Upgrade Gradle to 7.1.1 & Android Gradle Plugin to 4.2.2

oguzkocer:
This PR updates Gradle to 7.1.1 and Android Gradle Plugin to 4.2.2 which matches our major libraries & clients. Here are a few changes that were made as part of this change:
* Added plugin DSL support which is something we have been adding to all our projects
* Removes `enableUnitTestBinaryResources` from `gradle.properties` flag in d9927bdd69c0f96f2e692310107bafeae8ee8e57 which Gradle said is not supported - and didn't make a difference
* Removes unnecessary `buildToolsVersion` in fc6eb05252ac748093f024b2125d65fef414cc24 as it's now part of Android Gradle Plugin and has been stripped from our build files for other projects for some time
* Upgrades Kotlin to match current WPAndroid and fixes a couple very minor lint errors in 0b0a07d89871c2a7972fe706065e4544c0930653 - I could have upgraded the Kotlin version to the latest, but there were more lint errors and I didn't want to muddy the PR too much with Kotlin changes
* Updates Robolectric version to `4.4` which matches current WPAndroid and fixes an issue with the build f73f1f026dca0e7f8e9164d6a0412b8a861e4af3
* Updates `targetSdkVersion` to 29 - this was a necessary change to get the library building with the other changes. We probably want to upgrade it to 30 to match the latest changes in other projects. @ravishanker Could you take a look at that since you have recently worked on this? I am not sure if there is anything that needs to be changed in the library to support that change. _This is the only change I am a little bit worried about as I don't have enough knowledge/experience about target sdk changes._
* Git ignores `.idea/compiler.xml` which is something we have done a while back for other projects

These changes were made with the major goal of supporting composite builds, and minor goal of bringing the library up to speed with other projects. As always, there is still more to be improved.

**To test:**
Follow the test instructions in https://github.com/wordpress-mobile/WordPress-Android/pull/15432


build.gradle
```diff
@@ -1,27 +1,17 @@
...
-
-    repositories {
-        google()
-        jcenter()
-    }
-
-    dependencies {
-        classpath "com.android.tools.build:gradle:$gradlePluginVersion"
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
+        commonTargetSdkVersion = 29
```
ravishanker: Please bump it up to 30, given everything else is upgraded now.

oguzkocer: @ravishanker I've left a note for you in the PR description about this:

> Updates targetSdkVersion to 29 - this was a necessary change to get the library building with the other changes. We probably want to upgrade it to 30 to match the latest changes in other projects. @ravishanker Could you take a look at that since you have recently worked on this? I am not sure if there is anything that needs to be changed in the library to support that change. This is the only change I am a little bit worried about as I don't have enough knowledge/experience about target sdk changes.

Unfortunately your comment is not clear enough to tell whether you've read the PR description or if there is anything to be considered for upgrading this library's `targetSdkVersion`.

mchowning:
Just bringing a discussion from Slack to here. Before merging this we should ensure that we can still generate artifacts for this library. Currently, we're using [Jitpack](https://jitpack.io/#wordpress-mobile/AztecEditor-Android) for that, but the build from this PR [failed](https://jitpack.io/com/github/wordpress-mobile/AztecEditor-Android/650836934d/build.log). 

You mentioned that you might go ahead and just migrate this from jitpack to s3, which sounds like a fine solution to me. Of course, fixing the Jitpack issue is also fine, but if the migration isn't too hard it might be easier to maintain in the future. No strong preferences from my end though.

mchowning (approved):
Changes look good, and with [your other PR moving us from jitpack to S3](https://github.com/wordpress-mobile/AztecEditor-Android/pull/941), publishing shouldn't be a problem either.

PR #941, opened 2021-10-11
Publish to s3

oguzkocer:
This PR adds support for publishing library artifacts to S3. It works the same way as our other libraries. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2083 should be referenced to understand some of the decisions made for this setup.

### Test
* Test https://github.com/WordPress/gutenberg/pull/35606
* Test https://github.com/wordpress-mobile/WordPress-Android/pull/15457


mchowning (approved):
Code looks good to me and CI is happy. :shipit:

PR #939, opened 2021-10-06
Feature: Add support for the mark HTML tag

geriux:
### Fix

This PR adds support for the `mark` HTML tag.

It currently supports setting the text color and there are plans to add support for background-color as well.

<kbd></kbd>

### Test
1. Open the demo app
2. Scroll down to the bottom to find the new tag
3. Expect to see the text with the colors

### Review
@0nko 

Make sure strings will be translated:

- [x] If there are new strings that have to be translated, I have added them to the client's `strings.xml` as a part of the integration PR.

PR #943, opened 2021-11-16
Add Mark inline formatting

geriux:
This PR is a follow-up of https://github.com/wordpress-mobile/AztecEditor-Android/pull/939 to wrap up adding `Mark` formatting support.

To test this feature, use the iOS build from this [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/15504) that uses this feature.

### Fix

Adds Mark formatting capabilities.

### Test
For test cases, you can check the description of this [PR](https://github.com/WordPress/gutenberg/pull/36028).

### Review
@antonis 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
@@ -21,6 +21,7 @@ import org.wordpress.aztec.spans.AztecStyleStrongSpan
 import org.wordpress.aztec.spans.AztecStyleSpan
 import org.wordpress.aztec.spans.AztecUnderlineSpan
 import org.wordpress.aztec.spans.IAztecInlineSpan
+import org.wordpress.aztec.spans.MarkSpan
 import org.wordpress.aztec.watchers.TextChangedEvent
 import java.util.ArrayList
 
@@ -72,6 +73,14 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
                     AztecTextFormat.FORMAT_UNDERLINE,
                     AztecTextFormat.FORMAT_CODE -> {
                         applyInlineStyle(item, textChangedEvent.inputStart, textChangedEvent.inputEnd)
+                    }
+                    AztecTextFormat.FORMAT_MARK -> {
+                        // For cases of an empty mark tag, either at the beginning of the text or in between
+                        if (textChangedEvent.inputStart == 0 && textChangedEvent.inputEnd == 1) {
+                            applyMarkInlineStyle(textChangedEvent.inputStart, textChangedEvent.inputEnd)
+                        } else {
+                            applyInlineStyle(item, textChangedEvent.inputStart, textChangedEvent.inputEnd)
+                        }
                     }
                     else -> {
                         // do nothing
@@ -176,6 +185,13 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
         joinStyleSpans(start, end)
     }
 
+    private fun applyMarkInlineStyle(start: Int = selectionStart, end: Int = selectionEnd, attrs: AztecAttributes = AztecAttributes()) {
+        val previousSpans = editableText.getSpans(start, end, MarkSpan::class.java)
+        previousSpans.forEach {
+            it.applyInlineStyleAttributes(editableText, start, end)
+        }
+    }
+
     private fun applySpan(span: IAztecInlineSpan, start: Int, end: Int, type: Int) {
         if (start > end || start < 0 || end > editableText.length) {
             // If an external logger is available log the error there.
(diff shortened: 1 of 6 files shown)


geriux:
Hey there @hypest @mchowning 👋 

I'm having an issue with this feature and I saw you have worked on this project before, I was wondering if any of you might have some insight 😄 

Currently, if you have a word typed in, and an empty `<mark>` tag at the end, if you focus at the end to write something within the tag, it will let you add content but while you do so, the first word is wrapped by another `<mark>` tag with the CSS attributes of the original `mark` tag. The desired result would be to leave the first word as it was and be able to type content within the `mark` tag that's at the end.

https://user-images.githubusercontent.com/4885740/142177453-f041e8d7-a618-43ac-bfce-13e5590c4447.mp4

 Have you seen this before? I updated the demo app content to show the current issue. Any guidance would be appreciated. Thanks!

hypest:
👋 @geriux , thanks for the ping!

Haven't really dug deeply into this yet, but thought of sharing this in case it helps in the meantime: I wonder if there's some mixup with inline and block level elements handling here that leads to that separate <mark> to appear. For example, I see that it's the `blockFormatter` used here instead of the `inlineFormatter`. I understand that the `mark` tag is an inline-level HTML tag, right? So, I'd expect to only see it being handled by the inline elements related classes. Again, I haven't tried to really debug this so, take the above with a grain of salt :)

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -73,6 +74,14 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
                     AztecTextFormat.FORMAT_CODE -> {
                         applyInlineStyle(item, textChangedEvent.inputStart, textChangedEvent.inputEnd)
                     }
+                    AztecTextFormat.FORMAT_MARK -> {
+                        // For cases of an empty mark tag, either at the beginning of the text or in between
+                        if (textChangedEvent.inputStart == 0 && textChangedEvent.inputEnd == 1) {
+                                applyMarkInlineStyle(textChangedEvent.inputStart, textChangedEvent.inputEnd)
```
antonis: I think there is extra indentation on this line though the linter did not complain 🤔

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -176,6 +185,15 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
         joinStyleSpans(start, end)
     }
 
+    private fun applyMarkInlineStyle(start: Int = selectionStart, end: Int = selectionEnd, attrs: AztecAttributes = AztecAttributes()) {
+        val previousSpans = editableText.getSpans(start, end, MarkSpan::class.java)
+        previousSpans.forEach {
+            (it as MarkSpan).applyInlineStyleAttributes(editableText, start, end)
```
antonis: I think we could omit the casting here (`it as MarkSpan`) since `previousSpans` contains an array of `MarkSpan!`

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -176,6 +185,15 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle) : AztecFormat
...
+
+            return@forEach
```
antonis: We can safely omit the `return@forEach` here since it is a local return to the caller of the forEach loop that moves to next element of the array ([ref](https://kotlinlang.org/docs/returns.html#return-to-labels)).

antonis (approved):
Great work @geriux 👍 
I tested with the Gutenberg PR and everything works as expected. The code also LGTM 🎉 

I added a couple of comments that do not affect the code functionally :)

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -179,9 +179,8 @@ open class MainActivity : AppCompatActivity(),
         private val VIDEOPRESS_2 = "[wpvideo OcobLTqC w=640 h=400 autoplay=true html5only=true3]"
         private val QUOTE_RTL = "<blockquote>לְצַטֵט<br>same quote but LTR</blockquote>"
         private val MARK = "<p>Donec ipsum dolor, <mark style=\"color:#ff0000\">tempor sed</mark> bibendum <mark style=\"color:#1100ff\">vita</mark>.</p>"
-        private val MARK_EXAMPLE = "<p>Hello <mark style=\"background-color:rgba(0, 0, 0, 0);color:#151853\" class=\"has-inline-color has-tertiary-color\"></mark></p>"
 
-/*        private val EXAMPLE =
+        private val EXAMPLE =
```
hypest: Nit: what do you think about squashing this commit with [the one that introduced the commenting](https://github.com/wordpress-mobile/AztecEditor-Android/pull/943/commits/5fb3c5dedaaf3e1e1f1459a89999f9b626ef6153#diff-f560a91308e1ade3461de1cb5c543283d28801832ecf0a1475be13ce4eae23ecR184) to avoid adding extra steps to [the git-blame history](https://github.com/wordpress-mobile/AztecEditor-Android/blame/688a5bb8b10525f051ebd405af65a0818dd8a32b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L183)? No strong feeling on this.

geriux: Of course! I'll do that, thanks! 👍

PR #944, opened 2021-11-22
Add flag to control whether applying Aztec styles to HTML

fluiddot:
This PR introduces a new resource bool (`use_aztec_style_html`) as a flag to control whether, when parsing HTML, it should apply Aztec styles ([reference](https://github.com/wordpress-mobile/AztecEditor-Android/blob/5d983f84e74d0d6db4297e846ef010a3d5146612/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L1481-L1532)). This change shouldn't produce regressions as the flag by default is set to `true`, which is the default behavior.

Related to https://github.com/wordpress-mobile/gutenberg-mobile/issues/3136.

#### Verify that no regressions are introduced
1. Open the demo app.
2. Observe that the content is properly rendered with expected styles.

### Review
[TBD]


fluiddot:
This change is no longer required after https://github.com/wordpress-mobile/AztecEditor-Android/pull/945 being merged.

PR #945, opened 2021-11-24
Preserve link, code span instances when styling

hypest:
### Fix

PR to preserve the span and tag order for `link` and `code` tags. Although HTML doesn't dictate that the inline element order is of importance, there's no good reason for Aztec to alter the order when parsing.

In this PR, the code that applies style to the `a` and `code` spans is refactored to be able to set the style on the span without replace the span itself in the spannable.

This PR is an alternative approach of https://github.com/wordpress-mobile/AztecEditor-Android/pull/944.

### Test
1. In `MainActivity.kt` use the following html as value to the [EXAMPLE val](https://github.com/wordpress-mobile/AztecEditor-Android/blob/01da97c191e42e2289ba2170325bb64b2681d6c9/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L183): `<p>This is a <a href="http://wordpress.com"><strong>link</strong></a></p>`
2. Compile and run the demo app
3. Switch to html mode and notice that the markup is the same as the one in `EXAMPLE`
4. Switch back to visual mode
5. Type a new character in the beginning of the line
6. Switch to html mode again
7. Notice that the `strong` is still nested inside the `a`

The above steps fail without the changes in this PR.


fluiddot (approved):
LGTM 🎊 ! Thanks @hypest for addressing this issue and adding unit tests 🙇 .

I verified the changes in the Aztec demo app and WPAndroid app as well as, check that the new unit tests passed 🟢 .

_Tested on Samsung Galaxy S20 FE 5G (Android 11)._

PR #946, opened 2021-12-09
Drop jetifier

malinajirka:
Make sure https://github.com/wordpress-mobile/AztecEditor-Android/pull/947 is merged first -> update the target branch to `develop` unless GitHub does it automatically.

This PR is part of [the effort to drop Jetifier from WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/pull/15675). Aztec directly doesn't rely on the android support library, however, it is dependant on utils version which is not migrated to AndroidX. This PR is technically not necessary to unblock the PR in WPAndroid as gradle can resolve the utils version conflict and use the newer version in WPAndroid. However, since we are using composite builds and having inconsistent versions could bite us, I opted for updating it everywhere for consistency.

Changes
- Bumps utils version to a version which is migrated to AndroidX
- Bumps minSdkVersion from 16 to 21 => WordPress Utils requires version 18. However, since we never test any of the changes on Android < 21 I decided to bump the version to 21.
- Disables jetifier

To test
Make sure to disable Jetifier in your local gradle.properties  => Smoke test the example app

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,7 +8,7 @@ android {
 
     defaultConfig {
         applicationId "org.wordpress.aztec"
-        minSdkVersion 16
+        minSdkVersion commonMinSdkVersion
         targetSdkVersion commonTargetSdkVersion
         versionCode 1
         versionName "1.0"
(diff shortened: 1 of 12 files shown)


oguzkocer (approved):
Thank you for taking care of this @malinajirka 🙇‍♂️

PR #947, opened 2021-12-10
Remove xamarin.testcloud library

malinajirka:
This PR removes `com.xamarin.testcloud:espresso-support` which is not being used anymore. The library is not migrated to AndroidX and is blocking our progress of the "Drop Jetifier" effort.

~This is still a draft since we are waiting for confirmation from the rest of the team that this library is really unused.~

The library was added as part of POC/tool comparsion in 2017 and it hasn't been used since then.

### Test
Green CI checks should be enough.

### Review
@[USER_NAME]


oguzkocer (approved):
@malinajirka Thank you for taking care of this!

PR #949, opened 2021-12-31
Prevent update selected styles when text changes on Gutenberg mode

fluiddot:
**Related to Gutenberg PR: https://github.com/WordPress/gutenberg/pull/37676**

Gutenberg already controls the selected styles, so we need to prevent Aztec from modifying them when the text changes.

### Test
This change shouldn't affect the current behavior of the selected styles hence, we should just verify that Aztec works as expected regarding the selected styles when modifying text.

### Review
@guarani 
@antonis 


antonis (approved):
Great work @fluiddot 👍 
I tested the change with a local Android build and everything work as expected. The code change looks consistent to me 🎉

guarani (approved):
I see this is already approved but I gave it a quick test on Android and it works as expected 👍 Also the code changes on both platforms look very similar.
Thanks for fixing this issue, @fluiddot!

PR #952, opened 2022-01-27
Heading Span - Reset previous font metrics if the font spacing changes

geriux:
### Fix
This PR adds another check to take into account before resetting the font metrics, by checking if the font spacing changed..

### Test
Use the test instructions from this [Gutenberg PR](https://github.com/WordPress/gutenberg/pull/38205).

### Review
@antonis 


antonis (approved):
Thank you for you work on this @geriux 🙇 
I tested the solution via Approved in https://github.com/WordPress/gutenberg/pull/38205#pullrequestreview-866396688 and everything worked as expected. The code changes also look consistent to me 🎉

Issue #955, opened 2022-02-10
Enable custom ordering and filtering of toolbar buttons


planarvoid:
In order to improve functionality of the Aztec editor it would be great if we could allow only a subset of the toolbar actions (not all actions are supported by all usecases). In addition to this we should allow reordering of the toolbar items. 

This should work both in Basic and Advanced mode.

PR #956, opened 2022-02-10
Add reordering and filtering of toolbar buttons

planarvoid:
### Fix
https://github.com/wordpress-mobile/AztecEditor-Android/issues/955

This PR introduces functionality to reorder and filter toolbar buttons. By default it should work as it works now. In addition to this the `Aztec` class now accepts a configuration object `ToolbarOrder` which defines the order of actions in the bottom bar. In addition to this it supports moving the `plugins` block anywhere between the actions and vertical dividers. 

In the implementation I've extracted all the inner views from the toolbar into separate layouts which are later inflated based on the input. The default behaviour is the same as it is now. The changes included changing all the places which require a certain button to be present. 

When testing make sure the current interface is not broken!

### Test
1. Make sure the toolbar `advanced` mode is set to **false** (that's the default setting)
2. Run the test app
3. Test the bottom toolbar
4. Notice it contains the same actions as before
5. Notice all the actions work as expected

### Test
1. Make sure the toolbar `advanced` mode is set to **true** 
2. Run the test app
3. Test the bottom toolbar
4. Notice it contains the same actions as before
5. Notice all the actions work as expected

### Test
1. Make sure the toolbar `advanced` mode is set to **false** (that's the default setting)
2. Change the `Aztec` invocation by adding a `ToolbarOrder.BasicOrder` parameter
3. Add all the `ToolbarElements` you want in order you expect
4. Add the plugins block in the list
5. Run the test app
6. Notice the bottom bar contains the actions you defined in the order you set
7. Check the actions work as expected

### Test
1. Make sure the toolbar `advanced` mode is set to **false** (that's the default setting)
2. Change the `Aztec` invocation by adding a `ToolbarOrder.BasicOrder` parameter
3. Add all the `ToolbarElements` you want in order you expect
4. Do not add the plugins block in the list
5. Run the test app
6. Notice the bottom bar contains the actions you defined in the order you set
7. Notice the plugins are automatically added at the end of the list

### Test
1. Make sure the toolbar `advanced` mode is set to **true** (that's the default setting)
2. Change the `Aztec` invocation by adding a `ToolbarOrder.AdvancedOrder` parameter
3. Add all the `ToolbarElements` you want in order you expect
4. Notice you can add `collapsedItems` and `expandedItems` - expanded items are always visible, collapsed items are opened on the `expand` click
4. Add the plugins block to the `expandedItems` 
5. Run the test app
6. Notice the bottom bar contains the actions you defined in the order you set
7. Check the actions work as expected

### Test
1. Make sure the toolbar `advanced` mode is set to **true** (that's the default setting)
2. Change the `Aztec` invocation by adding a `ToolbarOrder.AdvancedOrder` parameter
3. Add all the `ToolbarElements` you want in order you expect
4. Notice you can add `collapsedItems` and `expandedItems` - expanded items are always visible, collapsed items are opened on the `expand` click
4. Do not add the plugins block
5. Run the test app
6. Notice the bottom bar contains the actions you defined in the order you set
7. Notice the plugins are automatically added at the end of the expanded list

### Review
@[USER_NAME]


hypest:
👋 @malinajirka , thanks for this PR!

> In addition to this the Aztec class now accepts a configuration object ToolbarOrder which defines the order of actions in the bottom bar.

I wonder if you perhaps have considered and turned down the idea of passing that configuration without the Aztec class' involvement? Like, performing the `setToolbarOrder()` outside and before [the Aztec instantiation](https://github.com/wordpress-mobile/AztecEditor-Android/blob/aca0d53b5f97bfc00c108e4383eb06774c9ef732/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L433). My rationale is to keep the button order concern out of the Aztec class as it doesn't seem to need to care about that?

Note: I haven't drilled deep into this and I'm not sure if there's a blocker somewhere, just wanted to check with you.

planarvoid:
hey @hypest ! Thanks for the suggestion. I was originally considering it but decided that I should use the setup currently used by the Aztec/Toolbar relationship (for example the toolbar listener is set through Aztec as well). However, I've changed it now so that the items are set on the Toolbar directly. I'm fine with both approaches 👍

malinajirka:
Thanks for working on this! Unless it's critical or a blocker for you, I will review it when I wrap up a project which is about to be released - at the end of this week or the beginning of the next week. If it's a blocker, please let me know and I'll make sure to find the time.

hypest:
Also, apologies for pinging Jirka above instead of you Vojta! 🤗

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/ToolbarItems.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    companion object {
+        fun addItems(container: LinearLayout, inflater: LayoutInflater, toolbarItemGroup: ToolbarItemGroup) {
+            var offset = 0
+            toolbarItemGroup.toolbarItems.forEachIndexed { index, toolbarElement ->
+                if (toolbarItemGroup.verticalDividerIndices.contains(index)) {
+                    val view = inflater.inflate(R.layout.format_bar_vertical_divider, null)
+                    container.addView(view, index + offset)
+                    offset += 1
+                }
```
danilo04: I think instead of using a different field to set the dividers (`verticalDividerIndices`), why don't you create a `ToolbarItem` for a divider. 

```kotlin
enum class ToolbarItem(val layout: Int? = null) {
        HEADING(R.layout.format_bar_button_heading),
        LIST(R.layout.format_bar_button_list),
        DIVIDER(R.layout.format_bar_vertical_divider),
        ...
```

You then would just add a `DIVIDER` item for the lines. That would simplify the logic.

danilo04 (commented):
Hi @planarvoid. If I understand correctly, this change tries to make more flexible the `AztecToolbar` implementation by allowing to have variable items on the toolbar. 

The only concern I have is that `ToolbarItems` seems to be disconnected from `ToolbarActions`. They are linked because the `ToolbarItem` view inflated contains the button defined for a `ToolbarAction`. This can be confusing for someone that tries to extend Aztec. The requirements that should be met to make `ToolbarItems` and `ToolbarActions` work together are:
1. The `ToolbarItem`'s layout should include a button.
2. The `ToolbarAction`'s button ID should match the ID in the `ToolbarItem`'s layout.

I don't think this is a blocking issue, but we might need to consider improving this aspect of `AztecToolbar`. 

I have also left a comment regarding the use of the field `verticalDividerIndices` in `ToolbarItemGroup` and how it can be improved to reduce the complexity of the implementation.

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/ToolbarItems.kt
planarvoid: I used to have a divider as a toolbar item but it didn't work well with the use of a set for the items (we shouldn't use one item more than once). On the other hand it might be neater to just check that manually and crash if it's used incorrectly. I'll change it back 👍

planarvoid:
thanks for the review @danilo04 , I think it makes a lot of sense! 

I've made some changes so let me know what you think. 

- I've added an interface `IToolbarItem` which has an optional layout. This interface is implemented by `ToolbarAction` (where it shouldn't affect current functionality) and the `PLUGINS` and `DIVIDER` items. I needed this to be able to have a single list of `items` into the layout. By doing this we can directly use the `ToolbarActions` as parameter of the layout. 
- There is now a `sanitize` function which removes duplicates and makes sure that the `PLUGINS` are always present in the correct layout. 

Let me know what you think about this approach! I didn't want to change the `IToolbarAction` interface because it is used in custom plugins. However, there is no way that the user can extend or change `ToolbarAction` enum.

danilo04 (approved):
Hi @planarvoid. I think the changes make sense and look good :shipit:!

Issue #957, opened 2022-02-16
Add support for code block


planarvoid:
The drawing support for inline `code` and `pre` block is already built in. We should add a toolbar action that would handle these actions. The expected behaviour is: 

### Inline code from selection
- Select a word in a sentence
- Click on `<>` icon
- The word gets highlighted with the `<code>` inline annotation
- The `<>` icon gets highlighted
- Click on the `<>` icon again
- The icon is not highlighted any more
- The word is no longer highlighted

### Block code from selection
- Select a block of text (multiple spans)
- Click on the `<>` icon
- The block of text gets highlighted with the `<pre><code>` block annotation
- Add a line
- The new line is still within the block (the behaviour is similar to `quote` 
- Click on `<>` icon again
- The highlighting is now removed

### Block code from cursor position
- Move the cursor into a paragraph
- Click on the `<>` icon
- The paragraph is converted to `<pre><code>` block
- Click on the `<>` again
- The paragraph is converted back and the code block is removed

### Block code from cursor position on an empty line
- Move the cursor on an empty line
- Click on the `<>` icon
- The line is converted to `<pre><code>` block
- Click on the `<>` again
- The line is converted back and the code block is removed

PR #958, opened 2022-02-16
Add support for code inline tag and for preformat block

planarvoid:
### Fix
https://github.com/wordpress-mobile/AztecEditor-Android/issues/957

This PR introduces the `code` and `preformat` functionality to the Aztec editor. The functionality to draw both waswas already implemented. In this PR I've introduces two new `ToolbarAction` which turns them on and off. Both actions are now hidden by default. You need to set toolbar items to enable them.

### Inline code from selection
- Select a word in a sentence
- Click on `<>` icon
- The word gets highlighted with the `<code>` inline annotation
- The `<>` icon gets highlighted
- Click on the `<>` icon again
- The icon is not highlighted any more
- The word is no longer highlighted

### Block code from selection
- Select a block of text (multiple spans)
- Click on the `preformat` icon
- The block of text gets highlighted with the `<pre><code>` block annotation
- Add a line
- The new line is still within the block (the behaviour is similar to `quote` 
- Click on `preformat` icon again
- The highlighting is now removed

### Block code from cursor position
- Move the cursor into a paragraph
- Click on the `preformat` icon
- The paragraph is converted to `<pre><code>` block
- Click on the `preformat` again
- The paragraph is converted back and the code block is removed


aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -559,7 +560,7 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
         }
 
         // if text is selected and action is styling - toggle the style
-        if (action.isStylingAction() && action != ToolbarAction.HEADING && action != ToolbarAction.LIST) {
+        if (action.isStylingAction() && action != ToolbarAction.HEADING && action != ToolbarAction.LIST && action != ToolbarAction.CODE) {
```
danilo04: If I understand correctly, we don't enter this block because `HEADING`, `LIST`, and now `CODE` have submenus when clicked. So we don't try to execute the action on the selected text or the whole text because the actual action will be executed in the submenu. 

The method `IToolbarAction.isStylingAction()` is used in `AztecToolbar` in two places:
1. To filter selected actions and add their text formats if there isn't any text selected. This is just if the action is `INLINE_STYLE`
```kotlin
      if (!editor!!.isTextSelected() && action.actionType == ToolbarActionType.INLINE_STYLE) {
            val actions = getSelectedActions()
            val textFormats = ArrayList<ITextFormat>()

            actions.filter { it.isStylingAction() }
                    .forEach { textFormats.add(it.textFormats.first()) }
```

2. If there are text selected, to toggle the style, with the exception of `HEADING`, `LIST`, and now `CODE`.
```kotlin
        // if text is selected and action is styling - toggle the style
        if (action.isStylingAction() && action != ToolbarAction.HEADING && action != ToolbarAction.LIST && action != ToolbarAction.CODE) {
            aztecToolbarListener?.onToolbarFormatButtonClicked(action.textFormats.first(), false)
            val returnValue = editor!!.toggleFormatting(action.textFormats.first())

            highlightAppliedStyles()

            return returnValue
        }
```

It seems to me that `HEADING`, `LIST`, and now `CODE` should not be styling actions because they won't be applied directly, just after the submenu is clicked or some specific logic to decide which format to apply (in case of `CODE` button). So instead of adding exceptions to this `if` statement for every toolbar action with a submenu, why don't we just declare these actions as not styling. 

```kotlin
enum class ToolbarAction constructor(
        override val buttonId: Int,
        override val buttonDrawableRes: Int,
        override val actionType: ToolbarActionType,
        override val textFormats: Set<ITextFormat> = setOf(),
        override val layout: Int? = null)
    : IToolbarAction, ToolbarItems.IToolbarItem {
       HEADING(R.id.format_bar_button_heading,
            R.drawable.format_bar_button_heading_selector,
            ToolbarActionType.LINE_BLOCK,
            setOf(AztecTextFormat.FORMAT_NONE),
            R.layout.format_bar_button_heading) {
        override fun isStylingAction(): Boolean {
            return false
        }
    },
     CODE(
            R.id.format_bar_button_code,
            R.drawable.format_bar_button_code_selector,
            ToolbarActionType.BLOCK_STYLE,
            setOf(AztecTextFormat.FORMAT_CODE, AztecTextFormat.FORMAT_PREFORMAT),
            R.layout.format_bar_button_code) {
        override fun isStylingAction(): Boolean {
            return false
        }                                             
    },
```

danilo04 (changes requested):
Hi @planarvoid. I have reviewed and tested the PR. 

I have left a comment regarding one aspect of the design that can be improved. The `Inline code from selection` test didn't work for me. Instead of changing the format of just the word, it changes the format of the whole sentence.

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
planarvoid: I think that this part:
```
      if (!editor!!.isTextSelected() && action.actionType == ToolbarActionType.INLINE_STYLE) {
            val actions = getSelectedActions()
            val textFormats = ArrayList<ITextFormat>()

            actions.filter { it.isStylingAction() }
                    .forEach { textFormats.add(it.textFormats.first()) }
```
could be applied even for heading/list/code if they are currently selected. The `getSelectedActions` takes whatever styling actions are now selected and applies them to the text format. 

> It seems to me that HEADING, LIST, and now CODE should not be styling actions because they won't be applied directly, just after the submenu is clicked or some specific logic to decide which format to apply (in case of CODE button). So instead of adding exceptions to this if statement for every toolbar action with a submenu, why don't we just declare these actions as not styling.

I think that it might be better to not rely on the `isStylingAction` method here because you're right, basically we skip that step if we want to manually handle the change (whether it's through a menu or in case of code where we want to be able to apply either block or inline action based on circumstances)

planarvoid: I've moved the `isStylingAction` block to the `else` part of the `when`. This way we'll only fall back to it if the action is not handled by anyone else.

planarvoid:
thanks for the review @danilo04 ! 
After discussing it with Sylvester I think I've misunderstood the behaviour. We're supposed to have 2 buttons: 
- `preformat` for the block code
- `code` for the inline code
This makes the functionality much simpler

danilo04 (approved):
Hi @planarvoid. Things got really simpler when you divided the two actions into two different buttons in the toolbar. I tested different scenarios and all of them worked :shipit: !!

Issue #959, opened 2022-02-23
Add support for a task list


planarvoid:
The editor currently supports ordered and unordered lists. We want to add support for task lists. This will be another type of a list on the same level as the other two (and we should be able to switch between these lists). The option will be disabled by default.

PR #960, opened 2022-02-24
Add support for a task list

planarvoid:
### Fix
https://github.com/wordpress-mobile/AztecEditor-Android/issues/959

In this PR I've added a support for task list. This option is available in the list menu. 
Changing the `checkbox` state was tricky. In the end the checkboxes are drawn within the new parent list span (not on the items themselves). I was considering another option. Drawing the checkbox as part of the `listItem` instead. This solution has a couple of disadvantages, the biggest one is that all the other bullets are drawn within the `list` container so switching the lists would've been hacky. 
The disadvantage of the current solution is that the whole list needs to be redrawn on checking a list item. If this solution is not good enough, we might want to try the other way. 
The `tasklist` is hidden by default. It needs to be enabled on the toolbar. 

### Test
1. Change the `MainActivity` to call `toolbar.enableTaskList()` 
2. Run the app
3. Notice the `lists` menu now contains a new option - task list
4. Try: 
    - Add a new task list
    - Check a few items
    - Check the toolbar icon state is `task list` 
    - Change the task list type to ordered or unordered list
    - Check the toolbar icon state has changed back
    - Change the list type back 

https://user-images.githubusercontent.com/1079756/155494268-7f308edc-fa9f-4371-8a14-b6d812836fa0.mp4


aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -144,6 +154,12 @@ class AztecTagHandler(val context: Context, val plugins: List<IAztecPlugin> = Ar
                 handleElement(output, opening, preformatSpan)
                 return true
             }
+            INPUT -> {
+                if (opening && attributes.getValue("type") == "checkbox") {
+                    return handleCheckboxInput(output, context, attributes)
```
planarvoid: In case there is a `checkbox` element in the `<ul><li>`, we change the type of the `<ul>` parent to `AztecTaskList`

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -93,7 +98,12 @@ class AztecTagHandler(val context: Context, val plugins: List<IAztecPlugin> = Ar
                 return true
             }
             LIST_UL -> {
-                handleElement(output, opening, createUnorderedListSpan(nestingLevel, alignmentRendering, AztecAttributes(attributes)))
+                val lastSpan = tagStack.lastOrNull()
+                if (!opening && lastSpan is AztecTaskListSpan) {
```
planarvoid: If we're closing the tag and the last span is `AztecTaskListSpan`, we know that the list contains a task list so we have to close the correct span.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
```diff
@@ -154,6 +170,29 @@ class AztecTagHandler(val context: Context, val plugins: List<IAztecPlugin> = Ar
         return false
     }
 
+    private fun handleCheckboxInput(output: Editable, context: Context, attributes: Attributes): Boolean {
```
planarvoid: Here we convert from structure like: 
```
<ul>
<li><input type="checkbox" checked />Item</li>
</ul>
```

to something like

```
<ul>
<li checked="true>Item</li>
</ul>
```
This means that we remove the `input` element and we set the `checked` attribute on the `li` container. We also need to change the parent `ul` type to task list in order to draw the checkboxes.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1521,6 +1526,37 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
+        taskLists.forEach { taskList ->
+            if (taskList.onRefresh == null) {
+                taskList.onRefresh = {
+                    refreshTaskListSpan(it)
+                }
+            }
+        }
+    }
+
+    private fun refreshTaskListSpan(taskList: AztecTaskListSpan) {
```
planarvoid, line 1540: This function removes and recreates the task list span. This is the only way to redraw anything in a Spannable. we cannot have animations for the redrawing.

aztec/src/main/kotlin/org/wordpress/aztec/EnhancedMovementMethod.kt
```diff
@@ -89,4 +94,15 @@ object EnhancedMovementMethod : ArrowKeyMovementMethod() {
 
         return super.onTouchEvent(widget, text, event)
     }
+
+    private fun handleTaskListClick(text: Spannable, off: Int): Boolean {
+        val clickedList = text.getSpans(off, off, AztecTaskListSpan::class.java).firstOrNull()
+        val clickedLine = text.getSpans(off, off, AztecListItemSpan::class.java).lastOrNull()
+        val spanStart = text.getSpanStart(clickedLine)
+        if (spanStart == off && clickedList != null && clickedLine != null && clickedList.canToggle()) {
```
planarvoid, line 101: the `spanStart == off` here makes sure that we only handle clicks that are before the line start. This means we'll only handle clicks on the checkboxes and around them.

aztec/src/main/kotlin/org/wordpress/aztec/handlers/ListItemHandler.kt
```diff
@@ -81,7 +84,12 @@ class ListItemHandler(
                 nestingLevel: Int,
                 alignmentRendering: AlignmentRendering
         ) {
-            set(text, createListItemSpan(nestingLevel, alignmentRendering), start, end)
+            val isInTaskList = !text.getSpans(start, end, AztecTaskListSpan::class.java).isNullOrEmpty()
```
planarvoid, line 87: We need to make sure that any new item added in the `AztecTaskListSpan` will be unchecked by default.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecListItemSpan.kt
```diff
@@ -23,10 +24,52 @@ fun createListItemSpan(nestingLevel: Int,
...
+                }
+                sb.append(">")
+                if (shouldReplaceCheckedAttribute) {
+                    sb.append("<input type=\"checkbox\" class=\"task-list-item-checkbox\"")
+                    if (attributes.getValue(CHECKED) == "true") {
+                        sb.append(" checked")
+                    }
+                    sb.append(" /")
+                }
+                return sb.toString()
```
planarvoid, line 64: This logic prints the missing `<input type="checkbox"/>` element when generating the HTML.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
danilo04: If I understand correctly, we cannot know if a `list` is a `task list`, until we go and find a `checkbox` inside the first element of the list. So, we go back and change the type of the list to task list. 

I was wondering if we can simplify this process by adding an attribute to the `<ul>` element itself. Something like `<ul type="task-list">` or `<ul class="task-list">`. With this change, we can set the correct span `list` or `task list` at the moment we see the list element `<ul>`. 

I can do that by adding the next override inside of `AztecTaskListSpan`:
```kotlin
override val startTag: String
       get() {
           if (!attributes.hasAttribute("type")) {
               attributes.setValue("type", "task-list")
           }
           return "$TAG $attributes"
       }
```


Then, the `LIST_UL` element can be handled like:

```kotlin
LIST_UL -> {
                val lastSpan = tagStack.lastOrNull()
                // We use the attribute [type] for opening tags and lastSpan for closing tags
                val span = if (attributes.getValue("type") == "task-list" ||  lastSpan is AztecTaskListSpan) {
                    createTaskListSpan(nestingLevel, alignmentRendering, AztecAttributes(attributes), context)
                } else {
                    createUnorderedListSpan(nestingLevel, alignmentRendering, AztecAttributes(attributes))
                }
                handleElement(output, opening, span)
                return true
            }
```

and `handleCheckboxInput(...)` just have to handle the `checked` attribute:

```kotlin
private fun handleCheckboxInput(output: Editable, context: Context, attributes: Attributes): Boolean {
        val wrappingListItem = tagStack.lastOrNull() as? AztecListItemSpan ?: return false
        val checkedAttribute = attributes.getValue("checked")
        val isChecked = checkedAttribute != null && checkedAttribute != "false"
        wrappingListItem.attributes.setValue(CHECKED, isChecked.toString())
        return true
    }
```

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
danilo04: Putting the possible change here as well (is in another comment down below). If we use an attribute (`type="task-list"), we could recognize the task list for both opening and closing tags. See comment below for complete implementation.

```kotlin
LIST_UL -> {
                val lastSpan = tagStack.lastOrNull()
                // We use the attribute [type] for opening tags and lastSpan for closing tags
                val span = if (attributes.getValue("type") == "task-list" ||  lastSpan is AztecTaskListSpan) {
                    createTaskListSpan(nestingLevel, alignmentRendering, AztecAttributes(attributes), context)
                } else {
                    createUnorderedListSpan(nestingLevel, alignmentRendering, AztecAttributes(attributes))
                }
                handleElement(output, opening, span)
                return true
            }
```

aztec/src/main/kotlin/org/wordpress/aztec/EnhancedMovementMethod.kt
danilo04: (💯 ) This is pretty cool!!

aztec/src/main/kotlin/org/wordpress/aztec/handlers/ListItemHandler.kt
danilo04: (👍 ) Awesome!

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecListItemSpan.kt
danilo04: I would add a comment to the start tag specifying that `AztecListItemSpan` handles items for all kinds of lists, including tasks lists that require a checkbox added. We check whether is a task list item by checking whether the item has an attribute `checked`.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecTaskListSpan.kt
```diff
@@ -0,0 +1,122 @@
...
+        var onRefresh: ((AztecTaskListSpan) -> Unit)? = null
+) : AztecListSpan(nestingLevel, listStyle.verticalPadding) {
+    private var toggled: Boolean = false
+    override val TAG = "ul"
+
+    override fun getLeadingMargin(first: Boolean): Int {
+        return listStyle.indicatorMargin + 2 * listStyle.indicatorWidth + listStyle.indicatorPadding
+    }
+
+    override fun drawLeadingMargin(c: Canvas, p: Paint, x: Int, dir: Int,
```
danilo04, line 79: (🤔 ) Interesting that this method also gets called when you enter a new list item. So this is the method that creates the checkboxes.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
danilo04: I would add this comment to the top of the method and the reasoning why. It helps understand the design decisions and if we want to implement something similar (another input type), we have to deal with this.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
danilo04: But according to one of your comments, we just refresh the task list if the user clicks on the checkboxes or nearby the checkboxes, so that is important to keep the refreshing rate low. I am correct on this statement @planarvoid ?

aztec/src/test/kotlin/org/wordpress/aztec/TaskListTest.kt
```diff
@@ -0,0 +1,959 @@
...
+import org.wordpress.aztec.watchers.EndOfBufferMarkerAdder
+
+/**
+ * Testing ordered and unordered lists.
+ *
+ * This test uses ParameterizedRobolectricTestRunner and runs twice - for ol and ul tags.
+ */
+@RunWith(ParameterizedRobolectricTestRunner::class)
+@Config(sdk = [23])
+class TaskListTest(listTextFormat: ITextFormat, listHtmlTag: String) {
```
danilo04, line 24: (❤️ ) Awesome!

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -711,6 +730,10 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
         this.toolbarItems = toolbarItems
     }
 
+    fun enableTaskList() {
```
danilo04, line 733: I don't like tying this to the toolbar, since it seems that this is a feature of the `LIST` element. But since the `ToolbarAction`'s are `enums`, then we cannot create specific methods for them. So I think this solution works.

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
```diff
@@ -238,6 +239,13 @@ class AztecToolbar : FrameLayout, IAztecToolbar, OnMenuItemClickListener {
                     return true
                 }
             }
+            KeyEvent.KEYCODE_T -> {
```
danilo04, line 242: (💯 ) Cool. I tested with my tablet and Bluetooth keyboard and it worked.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecTaskListSpan.kt
```diff
@@ -0,0 +1,122 @@
...
+        // Make sure the marker is correctly aligned on RTL languages
+        val markerStartPosition: Float = x + (listStyle.indicatorMargin * dir) * 1f
+        val d: Drawable = context.resources.getDrawable(R.drawable.ic_checkbox, null)
+        val leftBound = markerStartPosition.toInt()
+        if (isChecked(text, lineIndex)) {
+            d.state = intArrayOf(android.R.attr.state_checked)
+        } else {
+            d.state = intArrayOf()
+        }
+        d.setBounds(leftBound - 40, baseline - 40, leftBound + 8, baseline + 8)
```
danilo04: The bounds worked well for phones, but in tablets, the checkboxes don't have any padding. 

| Phone      | Tablet |
| ----------- | ----------- |
| image      | image       |

Where did you find those values (`40` and `8`), was through test and error?

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecTaskListSpan.kt
```diff
@@ -0,0 +1,122 @@
...
+        val lineIndex = getIndexOfProcessedLine(text, end) ?: return
+
+        val style = p.style
+        val oldColor = p.color
+
+        p.color = listStyle.indicatorColor
+        p.style = Paint.Style.FILL
+
+        // Make sure the marker is correctly aligned on RTL languages
+        val markerStartPosition: Float = x + (listStyle.indicatorMargin * dir) * 1f
```
danilo04, line 99: I tested it with RTL languages and it is working! But the bounds affect the look in tablets:
image

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
```diff
@@ -69,20 +71,39 @@ class BlockFormatter(editor: AztecText,
...
+            } else {
+                applyBlockStyle(AztecTextFormat.FORMAT_TASK_LIST)
+                editor.addRefreshListenersToTaskLists(editableText, selectionStart, selectionEnd)
+            }
+        } else {
+            if (containsList(AztecTextFormat.FORMAT_ORDERED_LIST, 0) || containsList(AztecTextFormat.FORMAT_UNORDERED_LIST, 0)) {
+                switchListType(AztecTextFormat.FORMAT_TASK_LIST)
+                editor.addRefreshListenersToTaskLists(editableText, selectionStart, selectionEnd)
+            } else {
+                removeBlockStyle(AztecTextFormat.FORMAT_TASK_LIST)
```
danilo04, line 102: I noticed that we just remove the style but the elements keep the checkbox item. This is useful if you change the list back to task list, but leave the checkboxes there is the list is kept unordered or ordered. 

Example
```html
<ol>
 <li><input type="checkbox" class="task-list-item-checkbox">hello</li>
 <li><input type="checkbox" class="task-list-item-checkbox">one more</li>
</ol>
```

Do you think this might be a problem?

danilo04 (commented):
Hi @planarvoid, awesome work on this PR. You apply some innovative stuff 🎉 .

I have reviewed and tested the changes. I have left some comments for your consideration regarding a change that might reduce the complexity of handling task lists (adding an attribute to distinguish an unordered list from a task list without going into its elements).

I also check that the bounds for checkboxes aren't working well in Tablets.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
planarvoid: thanks for this suggestion! As discussed in DM, I've implemented it in 5b264edb51b252d3ed4b5f48aacc45a96a60add0

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
planarvoid: I've implemented your solution. I've still used `!opening && lastSpan is AztecTaskListSpan` as the alternative because I think that otherwise it might affect a nested list. I'm not sure though.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
planarvoid: Good point, added in 42caab404109705e8ffc956fd58e5159027c57ca

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
planarvoid: yep, we only refresh it on user click on the checkbox or around it

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecListItemSpan.kt
planarvoid: good idea, I've added it in aee4f3b3250ed1e27b7870b6ab3fb0e07c0c89f5

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
planarvoid: This will actually change once another PR is merged (the indents). In our changes we want to have the toolbar list actions expanded so we'll have 3 toolbar actions - ordered list, unordered list and a task list - and we can remove this function.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecTaskListSpan.kt
planarvoid: it was test and error. I'll look into making this more reliable on bigger devices. Good catch! Do other lists work on tablets well?

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
planarvoid: good catch! I think this can be an issue. I'll remove the "checked" parameter when changing list style 👍

planarvoid: I've implemented that in 0bd6c7cd02e06342bc7b2f563087e7a2b91a9c44

aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt
danilo04: That is better I think

planarvoid:
thanks for the review @danilo04 I've fixed the tablet and RTL views: 

Screenshot 2022-03-03 at 16 51 33

and implemented all of your other suggestions. I think this is ready for another check!

danilo04 (approved):
Hi @planarvoid. Thank you for making some of the changes requested. I have tested them on phones and tablets and the checkboxes now are showing correctly on both devices. The only change required is that the test should be updated now that the implementation of the task list uses a parameter. Once that test is fixed, I think this PR is ready to merge.

Issue #965, opened 2022-03-03
Support nested lists


planarvoid:
### Expected
The editor is currently able to draw nested lists. However, there is no way to create them. In this issue we want to add support for two operations `indent` and `outdent`.
- Intent will take part of the list and create a nested list from it
- Outdent will do the opposite. 

These two operations will be optional.

PR #966, opened 2022-03-03
List indent and outdent

planarvoid:
### Fix
https://github.com/wordpress-mobile/AztecEditor-Android/issues/965

In this PR I've introduced 2 new actions: 
- indent 
- outdent
Both of these can be currently used only on lists and they are optional in the toolbar.
I've also extracted the `ordered` and `unordered` list into separate actions (again, both are hidden by default). 

https://user-images.githubusercontent.com/1079756/156547235-737b9ea8-6761-4da8-858e-d2aa569515be.mp4

### Test
- Make sure you add the necessary items to the toolbar in `MainActivity`. 
- Writing testing steps down can be tricky in this scenario. I think it's important to test the following usecases. Please try to test everything with both only a cursor and with multiple list items selected. The logic was extremely tricky to get right. 

To clarify terminology, in a list like this: 

- Test item 1
  - Test item 2
  - Test item 3

The item 1 is on the highest level, items 2 and 3 are on deeper levels, and are on the same level. Item 2 has previous item 1 and following item 3. 

### Indent 
- When previous item is missing is missing, you cannot indent
- When previous item is present but is on higher level, you cannot indent
- When previous item is present and is on the same or deeper level, you can indent - test with the following item missing, on the same level, on deeper level and on higher level. 
- You can only indent if all the selected items are on the same level
- When placing cursor, try extreme positions - end of line, end of list, beginning of a line. 
- Try to surround the list with other elements and indent then
- Generally test everything you can think of and please report any weirdness

### Outdent 
- When the following item is one level deeper, you cannot outdent. 
- When the current item is on the highest level, it should be removed from the list
- Otherwise it should be possible to outdent everything - test with the following item missing, on the same level, on deeper level and on higher level. 
- You can only outdent if all the selected items are on the same level
- When placing cursor, try extreme positions - end of line, end of list, beginning of a line. 
- Try to surround the list with other elements and outdent then
- Generally test everything you can think of and please report any weirdness


aztec/src/main/kotlin/org/wordpress/aztec/formatting/ListFormatter.kt
```diff
@@ -0,0 +1,245 @@
...
+        val directParentFlags = editableText.getSpanFlags(directParent)
+
+        val selectedItems = editableText.getSpans(selStart, selEnd, AztecListItemSpan::class.java).filterCorrectSpans(selectionStart = selStart, selectionEnd = selEnd)
+        if (!validateSelection(selectedItems, directParent)) return null
+        val selectedListItems = selectedItems.filter {
+            it.nestingLevel > directParent.nestingLevel
+        }
+        if (selectedListItems.isEmpty()) return null
+        val nestingLevel = selectedListItems.first().nestingLevel
+        selectedListItems.forEach {
```
AmandaRiu: np: unused?

aztec/src/main/kotlin/org/wordpress/aztec/formatting/ListFormatter.kt
```diff
@@ -0,0 +1,245 @@
...
+        val selectedParentItems = selectedItems.filter { it.nestingLevel < directParent.nestingLevel }
+        // This means multiple items on several list levels are selected
+        if (selectedParentItems.size > 1) {
+            return false
+        }
+        return true
+    }
+
+    private fun <T : IAztecBlockSpan> Array<T>.filterCorrectSpans(selectionStart: Int, selectionEnd: Int): List<T> {
+        return this.filterIndexed { index, span ->
```
AmandaRiu: np: `index` can be renamed to `_` since it is unused.

AmandaRiu (commented):
@planarvoid Still working through reviewing this one but it's close to EOD so I wanted to submit what I have now in case you want to work on it when you get in. The indent/outdent functionality works well. I was able to indent/outdent both ordered and unordered lists without any issue though I was at first a little confused because the option to indent/outdent is always enabled, even if it's not really allowed. 

The only big issue I've found so far has to do with undo/redo. It doesn't appear that indenting/outdenting is recorded in the history stack:

### Normal undo/redo

https://user-images.githubusercontent.com/5810477/160705398-010d572c-3c0e-4e16-b1fc-3573af222847.mp4

### Indented undo/redo

https://user-images.githubusercontent.com/5810477/160705415-97c2c063-fc91-42eb-85c2-1f9c26d96069.mp4

I did find that I was able to add indent/outdent actions to the history stack with these changes:

```
    fun indent() {
        history.beforeTextChanged(this@AztecText)
        blockFormatter.indent()
        contentChangeWatcher.notifyContentChanged()
    }

    fun outdent() {
        history.beforeTextChanged(this@AztecText)
        blockFormatter.outdent()
        contentChangeWatcher.notifyContentChanged()
    }
```

But did notice that while they successfully add them to the stack, unless there was some text change in the process it doesn't always undo/redo visually. Though if you switch to the code editor you can see the undo/redo is being successfully applied, it's just not updating visually. So there must be one more step missing:


https://user-images.githubusercontent.com/5810477/160705931-226fa5c6-353e-4dfb-8bac-eb068df826af.mp4

planarvoid:
thanks for the review @AmandaRiu ! this was a great catch 👍 . I've discovered that the issue was caused by incorrect wrapping of indented lists. Instead of: 
```
<ul>
    <li>First item
        <ul>
            <li>Second item</li>
        </ul>
    </li>
</ul>
```
it did 
```
<ul>
    <li>First item</li>
    <ul>
        <li>Second item</li>
    </ul>
</ul>
```
which is incorrect (but it felt logical). I wrote a lot of tests for this behaviour and I think I covered all the possible cases so let me know what you think. It was actually really tricky to implement.

AmandaRiu (approved):
Works beautifully! :shipit:

Issue #967, opened 2022-03-10
Make sure code inline style is exclusive


planarvoid:
### Expected
I'd expect the `code` inline style to be the only style selected (when selecting it after choosing `bold` or `italic` inline styles).

### Observed
You can currently select all of the inline style at once even though they are not visible. 

### Reproduced
1. Select a text
2. Click on "code" action
3. Click on "bold" action
4. Notice both styles are selected
5. Notice only "code" is visible

PR #968, opened 2022-03-10
Make sure code can be selected as the only current inline action and format

planarvoid:
### Fix
https://github.com/wordpress-mobile/AztecEditor-Android/issues/967

This PR makes sure that the code action behaves as an exclusive action - it's the only action that can be selected at once. 

### Test
1. Click in an empty line with no text selected
2. Click on "Code" action
3. Notice it gets checked
4. Type some text
5. Notice it's styled with the "code" style
6. Click on "bold" action
7. Notice the "code" action is no longer checked
8. Notice the "bold" action is now checked
9. Type some text
10. Notice it's styled with the "bold" style

### Test
1. Select some text
2. Click on "code" style
3. Notice the text is styled in "code" style
4. Click on "italic" 
5. Notice the "code" style is removed and "italic" style is applied
6. Click on "code" style
7. Notice the "italic" style is removed and the "code" style is applied

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
@@ -37,14 +37,5 @@ enum class AztecTextFormat : ITextFormat {
     FORMAT_MONOSPACE,
     FORMAT_CODE,
     FORMAT_MARK,
-    FORMAT_HIGHLIGHT;
-
-    companion object {
-        /**
-         * Marks all the inline text formats which are exclusive. When applied they cancel the other text formats.
-         */
-        val exclusiveInlineStyles: List<AztecTextFormat> = listOf(
-                FORMAT_CODE
-        )
-    }
+    FORMAT_HIGHLIGHT
 }
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
@@ -41,20 +41,6 @@ enum class AztecTextFormat : ITextFormat {
 
     companion object {
         /**
-         * Marks all the inline text formats which could be applied at once.
-         */
-        val inclusiveInlineStyles: List<AztecTextFormat> = listOf(
-                FORMAT_BOLD,
-                FORMAT_STRONG,
-                FORMAT_ITALIC,
-                FORMAT_EMPHASIS,
-                FORMAT_CITE,
-                FORMAT_STRIKETHROUGH,
-                FORMAT_UNDERLINE,
-                FORMAT_MARK
-        )
-
-        /**
          * Marks all the inline text formats which are exclusive. When applied they cancel the other text formats.
          */
         val exclusiveInlineStyles: List<AztecTextFormat> = listOf(
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
@@ -37,5 +37,28 @@ enum class AztecTextFormat : ITextFormat {
     FORMAT_MONOSPACE,
     FORMAT_CODE,
     FORMAT_MARK,
-    FORMAT_HIGHLIGHT
+    FORMAT_HIGHLIGHT;
+
+    companion object {
+        /**
+         * Marks all the inline text formats which could be applied at once.
+         */
+        val inclusiveInlineStyles: List<AztecTextFormat> = listOf(
+                FORMAT_BOLD,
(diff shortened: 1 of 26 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
```diff
@@ -37,5 +37,28 @@ enum class AztecTextFormat : ITextFormat {
...
+                FORMAT_UNDERLINE,
+                FORMAT_MARK
+        )
+
+        /**
+         * Marks all the inline text formats which are exclusive. When applied they cancel the other text formats.
+         */
+        val exclusiveInlineStyles: List<AztecTextFormat> = listOf(
+                FORMAT_CODE
+        )
```
AmandaRiu: I worry about this being such a manual process since it would mean that developers have to remember to add new exclusive/inclusive formats here as well. I'm not sure how feasible this is, but since there is already a `ToolbarActionType.INLINE_STYLE` that does not include the new `ToolbarActionType.EXCLUSIVE_INLINE_STYLE`, I wonder if it would make more sense to just return the list of `AztecTextFormat`s in those specific categories instead of manually listing them out here? WDYT?

AmandaRiu (commented):
@planarvoid Reviewed the code but without an actual "code" button in the example app I can't test it. Code changes look good though I did add one comment for your consideration.

AmandaRiu (commented):
@planarvoid I figured out how to add the code button and tested this functionality. All works well. Just have the one code comment for your review.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
planarvoid: I agree this is not perfect but I'm not sure this could be done in a better way. We can't tie `ToolbarAction` to `TextFormat` because. Toolbar action is only related to the toolbar and doesn't cover all the text formats. For example `Mark` cannot be added from the toolbar but it's still supported by the editor. On the other hand there are other text formats which are not inline. 
In the end I've changed the behaviour so that we only need to define the `exclusive` formats and the logic is that inclusive formats are inheriting from IAztecInlineStyle and are not `exclusive`. I think this solution is better.

planarvoid:
thanks for the review @AmandaRiu ! I've changed the logic. Let me know what you think

aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
AmandaRiu: @planarvoid I definitely like this simplified version, but since the `InlineFormatter` already uses `IAztecInlineStyle` to determine if a span is an inline span, why not extend `IAztecInlineSpan` to make a new `IAztecExclusiveInlineSpan` and have `AztecCodeSpan` implement it. This would make it easier to mark spans in the future as "exclusive" with two simple designations:
- Have the new span class implement `IAztecExclusiveInlineSpan`
- Define the new span as `ToolbarActionType.EXCLUSIVE_INLINE_STYLE` in `ToolbarAction`
This is nice because those are exactly the two places a new span would need to touch during the development process anyway thus ensuring that the new span will just "work" without having to do something differently for it just because it is exclusive. With that done we can just add a new method to `InlineFormatter` that works much the current `containsInlineStyle()` method, but instead checks if the span is exclusive. Since `IAztecExclusiveInlineSpan` inherits from `IAztecInlineSpan` the existing functionality will work as intended. 

If you choose to go this route I'd recommend adding `IAztecExclusiveInlineSpan` to the same file as `IAztecInlineSpan` so it's really clear what the options are and also highly recommend adding a class comment that clearly defines how this interface will work. For example:

```
package org.wordpress.aztec.spans

interface IAztecInlineSpan : IAztecSpan

/**
 * Extend this interface if an inline span should be exclusive, meaning
 * only a single inline format can be applied to this span.  
 */
interface IAztecExclusiveInlineSpan : IAztecInlineSpan
```

WDYT?

AmandaRiu (approved):
@planarvoid These updated changes work great, but I did leave an additional recommendation in the code comment thread for your review. I'll go ahead and approve this PR so that if you decide to keep it as-is you can go ahead and merge it.

aztec/src/main/kotlin/org/wordpress/aztec/AztecTextFormat.kt
planarvoid: good suggestion! I've used the `IAztecExclusiveInlineSpan` interface and I like this solution much better than before!

planarvoid:
thanks for the review @AmandaRiu  I've implemented your suggestion!

AmandaRiu (approved):
Works beautifully @planarvoid ! Really nice work here. :shipit:

Issue #969, opened 2022-03-11
Add option for block styles to be exclusive


planarvoid:
We want to have an option so that selecting a block style will disable the currently selected style
- Selecting a heading
- Clicking on "quote"
- Heading style is removed and "quote" style is set

PR #970, opened 2022-03-11
Enable exclusive blocks in AztecText.kt

planarvoid:
### Fix
https://github.com/wordpress-mobile/AztecEditor-Android/issues/969

In this PR I've added an optional setting which when set makes block styles exclusive. 

### Before test
Make sure you enable all the block styles in the MainActivity by calling:
```
        toolbar.setToolbarItems(ToolbarItems.BasicLayout(
                ToolbarAction.HEADING, ToolbarAction.LIST, ToolbarAction.QUOTE, ToolbarAction.PREFORMAT
        ))
```
Make sure you set this style on `AztecTextStyle`
```
<item name="exclusiveBlocks">true</item>
```

### Test a
1. Click on a heading
2. Select "Preformat" style
3. Notice that the heading style is removed and the preformat is applied
4. Select "heading" style
5. Notice the heading style is reapplied

### Test 2
1. Click on a list item (try first, last and an item in the middle)
2. Select "Heading" style
3. Notice that the item is removed from the list and the style is applied

### Test 3
1. Test various combinations of block styles

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
@@ -116,7 +116,7 @@ class BlockFormatter(editor: AztecText,
 
     fun togglePreformat() {
         if (!containsPreformat()) {
-                if (containsOtherHeadings(AztecTextFormat.FORMAT_PREFORMAT)) {
+                if (containsOtherHeadings(AztecTextFormat.FORMAT_PREFORMAT) && !exclusiveBlockStyles.enabled) {
                     switchHeadingToPreformat()
                 } else {
                     applyBlockStyle(AztecTextFormat.FORMAT_PREFORMAT)
@@ -135,7 +135,7 @@ class BlockFormatter(editor: AztecText,
             AztecTextFormat.FORMAT_HEADING_5,
             AztecTextFormat.FORMAT_HEADING_6 -> {
                 if (!containsHeadingOnly(textFormat)) {
-                    if (containsPreformat()) {
+                    if (containsPreformat() && !exclusiveBlockStyles.enabled) {
                         switchPreformatToHeading(textFormat)
                     } else if (containsOtherHeadings(textFormat)) {
                         switchHeaderType(textFormat)
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
@@ -46,12 +46,14 @@ class BlockFormatter(editor: AztecText,
                      private val quoteStyle: QuoteStyle,
                      private val headerStyle: HeaderStyle,
                      private val preformatStyle: PreformatStyle,
-                     private val alignmentRendering: AlignmentRendering
+                     private val alignmentRendering: AlignmentRendering,
+                     private val exclusiveBlockStyles: ExclusiveBlockStyles
 ) : AztecFormatter(editor) {
     data class ListStyle(val indicatorColor: Int, val indicatorMargin: Int, val indicatorPadding: Int, val indicatorWidth: Int, val verticalPadding: Int)
     data class QuoteStyle(val quoteBackground: Int, val quoteColor: Int, val quoteBackgroundAlpha: Float, val quoteMargin: Int, val quotePadding: Int, val quoteWidth: Int, val verticalPadding: Int)
     data class PreformatStyle(val preformatBackground: Int, val preformatBackgroundAlpha: Float, val preformatColor: Int, val verticalPadding: Int)
     data class HeaderStyle(val verticalPadding: Int)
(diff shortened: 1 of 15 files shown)


danilo04:
HI @planarvoid, I found a case that I am not sure should be allowed. 

1. On a Heading 2, tap on preformat
2. Then tap again to make it Heading 2
3. The block retains both styles, Heading 2 and Preformat.

Screen Capture on 2022-03-11 at 08-58-07

This is happening because of this:

```kotlin
fun togglePreformat() {
        if (!containsPreformat()) {
                if (containsOtherHeadings(AztecTextFormat.FORMAT_PREFORMAT)) {
                    switchHeadingToPreformat()
                } else {
                    applyBlockStyle(AztecTextFormat.FORMAT_PREFORMAT)
                }
        } else {
            removeEntireBlock(AztecPreformatSpan::class.java)
        }
    }
```

The `containsOtherHeadings(AztecTextFormat.FORMAT_PREFORMAT)` returns `true`, so it doesn't call `applyBlockStyle(AztecTextFormat.FORMAT_PREFORMAT)` which is the one that removes the previous block formatting.

planarvoid:
good catch @danilo04 , I've fixed it in 47bd6488634cbc5c0e12a2b34a4e00b322416450 where I've disabled the additional logic that was there for conversion between preformat and heading

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
```diff
@@ -212,6 +214,80 @@ class BlockFormatter(editor: AztecText,
...
+                editableText.setSpan(it, lineEnd, spanEnd, spanFlags)
+            } else if (spanStartsBeforeLineStart && !spanEndsAfterLineEnd) {
+                // If the selected line is at the end of a span, move the span end to the start of the line
+                editableText.removeSpan(it)
+                editableText.setSpan(it, spanStart, lineStart, spanFlags)
+            } else {
+                // In this case the line fully covers the span so we just want to remove the span
+                editableText.removeSpan(it)
+            }
+        }
```
danilo04, line 283: To be honest, I am not sure how this works, but from all the tests I ran, this is working correctly.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
```diff
@@ -212,6 +214,80 @@ class BlockFormatter(editor: AztecText,
...
+            // Only these supported blocks will be split/removed on block change
+            val format = when (it) {
+                is AztecHeadingSpan -> it.textFormat
+                is AztecOrderedListSpan -> AztecTextFormat.FORMAT_ORDERED_LIST
+                is AztecUnorderedListSpan -> AztecTextFormat.FORMAT_UNORDERED_LIST
+                is AztecTaskListSpan -> AztecTextFormat.FORMAT_TASK_LIST
+                is AztecListItemSpan -> AztecTextFormat.FORMAT_UNORDERED_LIST
+                is AztecQuoteSpan -> AztecTextFormat.FORMAT_QUOTE
+                is AztecPreformatSpan -> AztecTextFormat.FORMAT_PREFORMAT
+                else -> return@forEach
```
danilo04: This is the only concern I have about the PR. It seems that if I extend Aztec with a block span and I wanted to be exclusive, I have to come to this place and add it as well. This then becomes an [unknown unknown](https://medium.com/@nathan.fooo/notes-a-philosophy-of-software-design-f48206bb32de) because I need to know about this piece of code to make it work. 

Maybe if the `Span` has the information about whether or not the block is exclusive and also holds the `textFormat`. With this design, then it becomes obvious to the programmer who is going to add a new block and want it to be exclusive. 

WDYT?

danilo04 (commented):
Hi @planarvoid. I have reviewed and tested your PR. I have left some comments for your consideration. Let me know what do you think?

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
planarvoid: It's not obvious but I don't see how it could be done in another way 😢. That's why I try to comment the logic so it makes more sense

aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
planarvoid: that's a good point 👍. I agree and I'll try to improve it

AmandaRiu:
Hey @planarvoid 👋🏻 Since Danilo is on leave at the moment I'll be taking over this PR review. Let me know when it will be ready for the next round of reviews 👍🏻

AmandaRiu (approved):
@planarvoid Changes look good! I tested several scenarios and all worked as expected. I even added the HTML button option to the toolbar to verify the exclusive replacements were actually replacing each other. This PR is approved. :shipit:

PR #972, opened 2022-03-14
Point to trunk instead of develop

### LINKED ISSUE #971: Readme preview examples are broken
### Expected

To display the previews correctly in the readme.

### Observed

image

### Reproduced
1. Go to this URL: https://github.com/wordpress-mobile/AztecEditor-Android
2. Scroll down to these words, **Supports Android 4.1+ (API 16 - Jelly Bean)** 

#### Tested
N/A

hypest:
### Fix #971 

Updates the README so the links towads `develop` now point to `trunk`, since `develop` has been removed in favor of `trunk`.

### Test
Check that the README now shows the 2 images correctly.

👋 @planarvoid , added you as a reviewer since I know your are active on this repo these days, thanks!


planarvoid (changes requested):
I've tested the links and all work except the one pointing to the iOS repo (which doesn't seem to be moved to `trunk` yet (link - https://github.com/wordpress-mobile/AztecEditor-iOS/raw/trunk/RepoAssets/aztec.png)

hypest:
Oh, good catch @planarvoid ! I went ahead and added the asset to this repo too, and adjusted the pointer as well. Let me know what you think.

The new pointer won't work yet in this PR, but it should after merging it to trunk.

planarvoid (approved):
looks good! thanks for the changes 🙇

PR #973, opened 2022-03-29
Fix task list click issues

planarvoid:
### Fix
The task list is facing some issues when the checkbox is clicked. In this PR these should be fixed. 
The issues are when you add an item in the middle of the task list or when you try to click on an empty item at the end of the list. Please check whether the task list works as expected (if you can think of other test scenarios). 

### Test 1
1. Create a task list with multiple items
2. Add an item in the middle of the list
3. Try checking/unchecking the item and all the items around it
4. Confirm that the items are checked correctly

### Test 2
1. Create a task list with multiple items at the end of the document
2. Add an empty item at the end of the list
3. Move cursor in one of the non-empty items
4. Click on the empty item checkbox
5. Notice it gets checked correctly

### Review
@AmandaRiu 


AmandaRiu (approved):
Works great @planarvoid! :shipit:

PR #974, opened 2022-03-30
Make sure media and horizontal rule are only added after the currently selected block

planarvoid:
### Fix
The media and the horizontal rule are currently added inline. For example you can add an image as a list item. In this PR I'm adding a setting where you can disable this behaviour. The new functionality adds a media item or a HR after the currently selected block. In addition to this when deleting text before an image (with backspace), the selection is moved so that you don't end up moving the image to become a part of the previous block.

### Test 1
1. Make sure the functionality is enabled - call `aztec.visualEditor.addMediaAfterBlocks()` in the `MainActivity`
2. Click in a block (list, quote, code, heading) 
3. Insert an image/video/horizontal rule
4. Notice it gets added after the currently selected block

### Test 2
1. Make sure the functionality is disabled
2. Click in a block (list, quote, code, heading) 
3. Insert an image/video/horizontal rule
4. Notice it gets in the block as before

### Test 3
1. Enable the functionality and add a block in `Test 1`
2. Move cursor right before the item you added (with arrows)
3. Click on "backspace"
4. Notice the cursor is moved at the end of previous block and you can delete the previous block now

### Test 4
1. Enable the functionality and add a block in `Test 1`
2. Move cursor right after the item you added (with arrows)
3. Click on "backspace"
4. Notice the item is deleted correctly

### Review
@AmandaRiu 


aztec/src/main/kotlin/org/wordpress/aztec/formatting/MediaFormatter.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.wordpress.aztec.spans.AztecVideoSpan
+import org.wordpress.aztec.spans.IAztecBlockSpan
+import org.wordpress.aztec.spans.IAztecNestable
+import org.wordpress.aztec.watchers.EndOfBufferMarkerAdder
+import org.xml.sax.Attributes
+
+/**
+ * This class contains methods to add media and horizontal rule to the editor.
+ */
+class MediaFormatter(editor: AztecText) : AztecFormatter(editor) {
```
AmandaRiu: ❓ Curious why you moved all these methods out of `LineBlockFormatter`? It kind of makes `LineBlockFormatter` a class that doesn't actually format anything which may be confusing. It's also where developers would look for logic that deals with working with a single line within a block which is essentially what is happening inside `MediaFormatter`. I'm fine with it either way, just wondering what the thoughts were behind it for documentation purposes.

AmandaRiu (commented):
Nice work @planarvoid! I left some comments in the code for your review, and here are the results of my testing:

**HR Rule: Issue 1** The new functionality is a bit odd in this scenario when compared to the old functionality: If I place the cursor at the end of a **heading** and then tap to add a **horizontal rule** and then switch to code editor:

**Before** : The horizontal rule gets added after the heading block (though oddly in the visual editor the cursor jumps to an existing horizontal rule later in the document. This wouldn’t happen in a new document though):
 
```
<h1>Heading 1</h1>
<hr>
<h2>Heading 2</h2>
```


https://user-images.githubusercontent.com/5810477/160879972-9496d006-a0bd-486d-87c9-3a48d24febb5.mp4



**After**: The horizontal rule gets added as a child of an `<h2>` block (like it’s copying the block below it). The cursor stays where it should but this is likely because of the way it’s embedded inside another block:

```
<h1>Heading 1</h1>
<h2>
   <hr></h2>
<h2>Heading 2</h2>
```


https://user-images.githubusercontent.com/5810477/160880033-b3a7a8b6-6d61-4f0a-8348-9bb353541171.mp4



*Oddly, this does not happen when performing the same test with inserting images.

**HR Rule: Issue 2** Adding an HR Rule in the middle of a list works as expected, the HR rule is added to the end of the list block (but still has the nesting issues from above). However, using the arrow to move to the beginning of the HR Rule and then tapping backspace adds the HR Rule to the list while also deleting the last letter of the last item of the list:


https://user-images.githubusercontent.com/5810477/160880105-c4101386-f8ec-4130-b9c2-c1e6369d2196.mp4



I don’t really feel like this is necessarily an issue with the new code however since the existing code does the same thing when deleting a horizontal rule after a list. FYI.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/MediaFormatter.kt
planarvoid: I originally created new methods in `BlockFormatter` because this change makes the line blocks behave more like regular blocks. After that I thought that the logic would be better in the same file so I created a new one. I agree that I can use the `LineBlockFormatter`. I'll move the code there 👍

planarvoid:
thanks for the review @AmandaRiu ! the issue was caused by not adding `\n` before the `<hr>` when it was added as a block. This is necessary because otherwise the `hr` is part either of the previous or the following block item. I'm still not 100% sure why the new `h2` was created but adding the `\n` fixes the issue (I've done that with the images before).

AmandaRiu (approved):
Works beautifully @planarvoid and is ready to merge but I did notice that in my code comments you had said you would move the code back into `LineBlockFormatter` but haven't yet so leaving this open. Feel free to merge it when ready 👍🏻

planarvoid:
I forgot to push it @AmandaRiu 🤦

PR #976, opened 2022-04-06
Fix issues with block media at the end of editable text

planarvoid:
### Fix
There was an issue I discovered when I tried to add an image after the last block of the content. This is now fixed in this PR. I've added a lot of tests covering everything (and found a few minor issues during the process). 

### Test
1. Enable the feature by calling `aztec.visualEditor.addMediaAfterBlocks()`
2. Load a content ending with a list
3. Click in the middle of the list
4. Add an image (or HR)
5. Notice it gets added correctly after the list (previously it was added as a new list item)

### Review
@AmandaRiu 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
@@ -168,7 +168,7 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
     }
 
     private fun insertSpanAfterBlock(ssb: SpannableStringBuilder) {
-        var position = getEndOfBlock()
+        val position = getEndOfBlock()
         // We need to be sure the cursor is placed correctly after media insertion
         // Note that media has '\n' around them when needed
         val isLastItem = position == EndOfBufferMarkerAdder.safeLength(editor)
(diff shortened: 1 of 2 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
```diff
@@ -163,33 +163,48 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
...
     private fun insertMediaAfterBlock(span: AztecMediaSpan) {
         val ssb = SpannableStringBuilder(Constants.IMG_STRING)
-        ssb.append("\n")
         buildClickableMediaSpan(ssb, span)
         insertSpanAfterBlock(ssb)
     }
 
     private fun insertSpanAfterBlock(ssb: SpannableStringBuilder) {
-        val position = getEndOfBlock()
+        var position = getEndOfBlock()
```
AmandaRiu: np: this value is never modified so it can be changed back to `val`

AmandaRiu (approved):
Tested and works beautifully! 

Screen Shot 2022-04-07 at 2 04 04 PM

Also ran the new tests and all passed. Left a small low-priority code comment for your review @planarvoid . This PR is approved and is ready to merge when you are! :shipit:

PR #977, opened 2022-04-08
Introduce placeholders for android views

planarvoid:
This PR is a draft for something I was working on recently. Our spannable approach doesn't allow showing of more complex views. For example it's impossible to create a video span that would play the video inline. Adding any type of a complex span is also very complicated. With this solution it is possible to add android views over a placeholder span. The Android view reacts to being deleted/moved around and works quite smoothly. The two placeholder drawers are currently just a POC to show it works. Only the code in the `aztec` repo should be +- final. Whatever is in the app module will be changed.

Questions raised from this PR are: 
- does the architecture look ok? I think I'll polish it a bit but a generally it should remain as it is
- does it work as expected? or is there something you'd polish?
- where should it live? it's currently in the `app` repo. I can either move it to a module (like the `glide-loader` module) or we can move it to our own repo and don't have it in Aztec. 

### Test
1. You'll notice when you start the app that there is already a test placeholder present
2. Try to delete it with backspace
3. Notice it works
4. You can add new placeholders by adding images or videos from your phone (I've replaced the functionality for testing)
    1. If you add a video, it should be shown and auto play (for testing purposes)
    2. If you add an image, it adds a test android layout which is clickable (shows a toast on click)
   

### Review
@AmandaRiu 


AmandaRiu (commented):
@planarvoid I still have more reviewing to do on the PR, but the premise is very interesting and I think this is heading in the right direction. I'm not sure where to put the new code yet but I don't think it should be in the app module (which would mean we'd implement it directly in our app if we want to use it). It feels like something that should either be a new optional plugin module that can be loaded if desired, but I'll give it some more thought.

I did some initial testing and this is what I found so far:
- Image placeholder doesn't load an image:

Screen Shot 2022-05-05 at 6 48 36 PM

- The cursor doesn't properly line up with placeholders once additional new lines are added to the editor.
- Unable to move cursor past the first placeholder once an additional placeholder is added to the editor.

https://user-images.githubusercontent.com/5810477/167041110-bb14254d-38ec-452c-aed6-96e8e12dfc48.mov


- If I just start typing and then press enter in the scenario above, then the text moves behind the place holder. Also can't delete second placeholder.



https://user-images.githubusercontent.com/5810477/167041139-e874551c-58d0-4408-b6a3-917412a65bad.mov

I'll continue my review when I return next week.

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherAPI25AndHigher.kt
```diff
@@ -13,11 +13,11 @@ class DeleteMediaElementWatcherAPI25AndHigher(aztecText: AztecText) : TextWatche
     private var deletedSpans = ArrayList<AztecMediaSpan>()
 
     override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) {
-        if (aztecTextRef.get()?.isTextChangedListenerDisabled() ?: true) {
+        if (aztecTextRef.get()?.isTextChangedListenerDisabled() != false) {
```
AmandaRiu, line 16: This changes the logic of the original if condition. Was this intentional? The equivalent to what is happening originally would be:

```
if (aztecTextRef.get() == null || aztecTextRef.get()?.isTextChangedListenerDisabled() != false) {
```

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherAPI25AndHigher.kt
```diff
@@ -13,11 +13,11 @@ class DeleteMediaElementWatcherAPI25AndHigher(aztecText: AztecText) : TextWatche
...
             return
         }
 
-        if (aztecTextRef.get()?.isMediaDeletedListenerDisabled() ?: true) {
+        if (aztecTextRef.get()?.isMediaDeletedListenerDisabled() != false) {
```
AmandaRiu, line 20: Same comment/question as above

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherPreAPI25.kt
```diff
@@ -10,17 +10,18 @@ class DeleteMediaElementWatcherPreAPI25(aztecText: AztecText) : TextWatcher {
     private val aztecTextRef: WeakReference<AztecText?> = WeakReference(aztecText)
 
     override fun beforeTextChanged(text: CharSequence, start: Int, count: Int, after: Int) {
-        if (aztecTextRef.get()?.isTextChangedListenerDisabled() ?: true) {
+        if (aztecTextRef.get()?.isTextChangedListenerDisabled() != false) {
```
AmandaRiu, line 13: This changes the logic of the original if condition. Was this intentional? The equivalent to what is happening originally would be:

```
if (aztecTextRef.get() == null || aztecTextRef.get()?.isTextChangedListenerDisabled() != false) {
```

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherPreAPI25.kt
```diff
@@ -10,17 +10,18 @@ class DeleteMediaElementWatcherPreAPI25(aztecText: AztecText) : TextWatcher {
...
             return
         }
 
-        if (aztecTextRef.get()?.isMediaDeletedListenerDisabled() ?: true) {
+        if (aztecTextRef.get()?.isMediaDeletedListenerDisabled() != false) {
```
AmandaRiu, line 17: Same as above

app/src/main/kotlin/org/wordpress/aztec/demo/LayoutDrawer.kt
```diff
@@ -0,0 +1,20 @@
+package org.wordpress.aztec.demo
+
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.View
+import android.widget.TextView
+import org.wordpress.android.util.ToastUtils
+import org.wordpress.aztec.demo.PlaceholderManager.PlaceholderDrawer.PlaceholderHeight
+
+class LayoutDrawer(override val placeholderHeight: PlaceholderHeight = PlaceholderHeight.Ratio(0.5f)) : PlaceholderManager.PlaceholderDrawer {
```
AmandaRiu: I kept reading this as "drawer" as in the [Navigation Drawer](url) 🤣 Maybe renaming it to something like `PlaceholderLayout` or `PlaceholderView` would clear up the intention.

app/src/main/kotlin/org/wordpress/aztec/demo/VideoDrawer.kt
```diff
@@ -0,0 +1,40 @@
+package org.wordpress.aztec.demo
+
+import android.content.Context
+import android.net.Uri
+import android.view.View
+import android.widget.MediaController
+import android.widget.VideoView
+import org.wordpress.aztec.demo.PlaceholderManager.PlaceholderDrawer.PlaceholderHeight
+
+class VideoDrawer(private val context: Context, override val placeholderHeight: PlaceholderHeight = PlaceholderHeight.Ratio(0.5f)) : PlaceholderManager.PlaceholderDrawer {
```
AmandaRiu: Just like with `LayoutDrawer`, I kept reading this as "drawer" as in the [Navigation Drawer](url) 🤣 Maybe renaming it to something like `PlaceholderVideoView` would work?  I think having `Placeholder` in the name would make it clear this component works with the Placeholder element. WDYT?

AmandaRiu (commented):
@planarvoid I finished my review! 🎉 Like I said in my last review I really like where this is heading. I didn't find anything that really stood out to me in the code, but left a couple comments. While testing I ran into another issue. I know this is still a work in progress but I figure I'll report anything I find so it can be taken into consideration for future updates if we go this route for production. 

If you switch to view the source code the placeholders are still visible. Once you switch back something strange happens to the video placeholder:

https://user-images.githubusercontent.com/5810477/169184947-13420830-74ff-4ab8-9caa-a2accbf32de7.mov

app/src/main/kotlin/org/wordpress/aztec/demo/LayoutDrawer.kt
planarvoid: I wanted to make it clear this class draws the views. It's not the view itself. Maybe an `adapter` would work?

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherPreAPI25.kt
planarvoid: This was offered by the Android Studio and I don't think it changes the logic. Both the conditions are true when the `aztecTextRef.get()?.isTextChangedListenerDisabled() == null` because `null != false` is true. The only case when both are false are if the `aztecTextRef != null && isTextChangedListenerDisabled == false`

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherPreAPI25.kt
planarvoid: Screenshot 2022-05-24 at 11 14 21

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherAPI25AndHigher.kt
planarvoid: I went bottom to the top so I've answered this already in another place 👍

aztec/src/main/kotlin/org/wordpress/aztec/watchers/DeleteMediaElementWatcherAPI25AndHigher.kt
planarvoid: the same goes for this one

planarvoid:
thanks for the review @AmandaRiu ! What I've done is: 
- fixed the behaviour. Somehow it got broken and I haven't noticed. I think it works really nicely when the placeholder is not the only thing in the document. It works ok even when it's the only thing there but sometimes it's a little bit yanky (I have no idea why, there seems to be an issue with the edit text). Let me know if you face any issues.
- moved the whole logic to a module so it can be imported independently
- renamed `drawer` to `adapter`
- created a new `adapter` to draw an image with caption and deleted the sample ones

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -180,39 +180,11 @@ open class MainActivity : AppCompatActivity(),
...
-                        VIDEOPRESS_2 +
-                        AUDIO +
-                        GUTENBERG_CODE_BLOCK +
-                        QUOTE_RTL +
-                        MARK
+                """
+                    <br>
+                    <placeholder id="123" type="image_with_caption" src="https://filesamples.com/samples/image/jpg/sample_640%C3%97426.jpg" caption="caption of image"/>
+                    <br>
+                """.trimIndent()
```
AmandaRiu: np: Probably don't want to commit this change since it removes the example blocks. Maybe instead add the placeholder as a sample block?

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ImageWithCaptionAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+import android.widget.ImageView
+import android.widget.LinearLayout
+import android.widget.TextView
+import com.bumptech.glide.Glide
+import org.wordpress.aztec.AztecAttributes
+import org.wordpress.aztec.placeholders.PlaceholderManager.PlaceholderAdapter.PlaceholderHeight
+
+/**
+ * A sample drawer which draws a custom layout over the placeholder. Treat this as an example of what can be done.
+ * This drawer draws an image with a caption under it
```
AmandaRiu: np: update comment to use "adapter" instead of "drawer"

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ImageWithCaptionAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    override fun onPlaceholderDeleted(placeholderId: String) {
+        media.remove(placeholderId)
+    }
+
+    data class ImageWithCaptionObject(val id: String, val src: String, val layoutId: Int)
+
+    companion object {
+        private const val DRAWER_TYPE = "image_with_caption"
```
AmandaRiu: np: Recommend renaming to `ADAPTER_TYPE` for consistency

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,338 @@
...
+        private val aztecText: AztecText,
+        private val container: FrameLayout
+) : AztecContentChangeWatcher.AztecTextChangeObserver,
+        IHtmlTagHandler,
+        AztecText.OnMediaDeletedListener,
+        CoroutineScope {
+    private val job = Job()
+    override val coroutineContext: CoroutineContext
+        get() = Dispatchers.Main + job
+    private val drawers = mutableMapOf<String, PlaceholderAdapter>()
```
AmandaRiu: np: recommend renaming to "adapters" for consistency

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,338 @@
...
+        job.cancel()
+        drawers.clear()
+    }
+
+    /**
+     * Register a custom drawer to draw a custom view over a placeholder.
+     */
+    fun registerDrawer(placeholderAdapter: PlaceholderAdapter) {
+        drawers[placeholderAdapter.type] = placeholderAdapter
+    }
```
AmandaRiu: np: Recommend renaming to "adapters" for consistency

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,338 @@
...
+     * Call this method to manually insert a new item into the aztec text. There has to be a drawer associated with the
+     * item type.
+     * @param id placeholder ID
+     * @param type placeholder type
+     * @param attributes other attributes passed to the view. For example a `src` for an image.
+     */
+    fun insertItem(id: String, type: String, vararg attributes: Pair<String, String>) {
+        if (drawers[type] == null) throw IllegalArgumentException("Drawer for inserted type not found. Register it with `registerDrawer` method")
+        val attrs = getAttributesForMedia(id, type, attributes)
+        val drawer = drawers[type] ?: return
```
AmandaRiu: Recommend simplifying:

```suggestion
        val drawer = drawers[type] ?: throw IllegalArgumentException("Drawer for inserted type not found. Register it with `registerDrawer` method")
        val attrs = getAttributesForMedia(id, type, attributes)
```

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,338 @@
...
+                return match
+            }
+        }
+        val targetPosition = aztecText.getElementPosition(predicate) ?: return
+
+        insertInPosition(aztecAttributes ?: return, targetPosition, currentPosition)
+    }
+
+    private fun insertInPosition(attrs: AztecAttributes, targetPosition: Int, currentPosition: Int? = null) {
+        validateAttributes(attrs)
```
AmandaRiu: This check doesn't currently do anything if it fails. Maybe do something like 
```suggestion
        if (!validateAttributes(attrs)) {
             // throw something?
        }
```

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ImageWithCaptionAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    override fun onViewCreated(view: View, placeholderId: String) {
+        val image = media[placeholderId]!!
+        val width = view.width
+        val imageView = view.findViewById<ImageView>(image.layoutId)
+        val height = getHeight(width)
+//        imageView.layoutParams = ViewGroup.LayoutParams(width, height)
+
+        Glide.with(view).load(image.src).into(imageView)
```
AmandaRiu, line 65: I experienced a crash at this line due to `imageView` being null. I'll have more details in the main PR Review but wanted to place a note here as a reminder.

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,338 @@
...
+            index += 1
+        }
+        return index
+    }
+
+    private fun validateAttributes(attributes: AztecAttributes): Boolean {
+        if (!attributes.hasAttribute(UUID_ATTRIBUTE)) return false
+        if (!attributes.hasAttribute(TYPE_ATTRIBUTE)) return false
+        val type = attributes.getValue(TYPE_ATTRIBUTE)
+        return drawers[type] != null
```
AmandaRiu: This could be simplified, but maybe it's more readable the way it is. I just wanted to play with it:

```suggestion
        return when {
            !attributes.hasAttribute(UUID_ATTRIBUTE) || !attributes.hasAttribute(TYPE_ATTRIBUTE) -> false
            else -> drawers[attributes.getValue(TYPE_ATTRIBUTE)] != null
        }
```

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,338 @@
...
+                it.visibility = View.GONE
+                container.removeView(it)
+            }
+        }
+    }
+
+    /**
+     * Called before media is deleted. There is a delay between user deleting a media and when the media is actually is
+     * confirmed. That's why we first hide the media and we delete it when `onMediaDeleted` is actually called.
+     */
```
AmandaRiu, line 214: Brilliant ⚡ 👍🏻

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,338 @@
...
+            launch {
+                delay(100)
+                // At this point we know the editor width so we need to update the drawable.
+                val editorWidth = aztecText.width
+                span.drawable?.setBounds(0, 0, editorWidth, drawer.getHeight(editorWidth))
+                aztecText.refreshText(false)
+                delay(100)
+                // Once the drawable and the placeholder are redrawn, we can finally insert the custom view over it.
+                insertInPosition(aztecAttributes, position)
+            }
```
AmandaRiu: I think this can be made a bit more reliable using a `viewTreeObserver`. When I use this `EXAMPLE` editor text:

```
private val EXAMPLE =
        """
            <br>
            <placeholder id="126" type="image_with_caption" src="https://examplebloge.files.wordpress.com/2017/02/3def4804-d9b5-11e6-88e6-d7d8864392e0.png" caption="Happy times"/>
            <br>
            <placeholder id="0" type="image_with_caption" src="https://filesamples.com/samples/image/jpg/sample_640%C3%97426.jpg" caption="Test caption" />
            <br>
            <h1>Amanda was here</h1>
            <br>
            <placeholder id="1" type="image_with_caption" src="https://images.unsplash.com/photo-1608744882201-52a7f7f3dd60?crop=entropy&cs=tinysrgb&fm=jpg&ixlib=rb-1.2.1&q=80&raw_url=true&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=736" caption="Test caption" />                 
            <br>
            <placeholder id="2" type="image_with_caption" src="https://examplebloge.files.wordpress.com/2017/02/3def4804-d9b5-11e6-88e6-d7d8864392e0.png" caption="Test caption" />
            <br>
        """.trimIndent()
```

 with the original code this is the result:

original

but using the updated code below I get:

view-tree-observer

```suggestion
            aztecText.viewTreeObserver.addOnGlobalLayoutListener(object : ViewTreeObserver.OnGlobalLayoutListener {
                override fun onGlobalLayout() {
                    aztecText.viewTreeObserver.removeOnGlobalLayoutListener(this)
                    launch {
                        val editorWidth = aztecText.width
                        span.drawable?.setBounds(0, 0, editorWidth, drawer.getHeight(editorWidth))
                        aztecText.refreshText(false)
                        insertInPosition(aztecAttributes, position)
                    }
                }
            })
```

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ImageWithCaptionAdapter.kt
```diff
@@ -0,0 +1,86 @@
...
+import android.widget.TextView
+import com.bumptech.glide.Glide
+import org.wordpress.aztec.AztecAttributes
+import org.wordpress.aztec.placeholders.PlaceholderManager.PlaceholderAdapter.PlaceholderHeight
+
+/**
+ * A sample drawer which draws a custom layout over the placeholder. Treat this as an example of what can be done.
+ * This drawer draws an image with a caption under it
+ */
+class ImageWithCaptionAdapter(override val placeholderHeight: PlaceholderHeight = PlaceholderHeight.Ratio(0.5f), override val type: String = "image_with_caption") : PlaceholderManager.PlaceholderAdapter {
```
AmandaRiu: np: code formatting could be cleaner here to make it easier to read

AmandaRiu (changes requested):
@planarvoid Definitely getting closer! What happened to the video component? Will that be added back in a different PR? I left some code comments, plus have the following notes from testing:

- Love the name change to `Adapter` - makes a lot more sense. Thanks for doing that. I started leaving code comments to suggest updating comments and variables to use “adapter” instead of “drawer for consistency, but I stopped because I didn’t want to flood the review with a bunch of nitpick comments 😃 I recommend just doing a search and updating occurrences.

- When switching to view the source code the placeholder is still visible





- When attempting to add text after the first freshly placed placeholder image, the text goes behind the image, but then corrects when `ENTER` is pressed. However, adding another placeholder image will push the text back behind the first placeholder.

https://user-images.githubusercontent.com/5810477/170144550-ddd8a91f-8bef-4df3-a32e-5d46c34f3c31.mov



- Using Undo with a placeholder present will cause the app to crash:

https://user-images.githubusercontent.com/5810477/170144571-29900185-d9bd-43f7-bb42-d65a43da7e31.mov



```
2022-05-24 15:28:16.303 11149-11149/org.wordpress.aztec E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wordpress.aztec, PID: 11149
    java.lang.NullPointerException: Argument must not be null
        at com.bumptech.glide.util.Preconditions.checkNotNull(Preconditions.java:29)
        at com.bumptech.glide.util.Preconditions.checkNotNull(Preconditions.java:23)
        at com.bumptech.glide.RequestBuilder.into(RequestBuilder.java:669)
        at org.wordpress.aztec.placeholders.ImageWithCaptionAdapter.onViewCreated(ImageWithCaptionAdapter.kt:66)
        at org.wordpress.aztec.placeholders.PlaceholderManager.insertInPosition(PlaceholderManager.kt:156)
        at org.wordpress.aztec.placeholders.PlaceholderManager.insertInPosition$default(PlaceholderManager.kt:112)
        at org.wordpress.aztec.placeholders.PlaceholderManager$handleTag$1.invokeSuspend(PlaceholderManager.kt:267)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.ResumeModeKt.resumeMode(ResumeMode.kt:67)
        at kotlinx.coroutines.DispatchedKt.resume(Dispatched.kt:272)
        at kotlinx.coroutines.DispatchedKt.dispatch(Dispatched.kt:261)
        at kotlinx.coroutines.CancellableContinuationImpl.dispatchResume(CancellableContinuationImpl.kt:218)
        at kotlinx.coroutines.CancellableContinuationImpl.resumeImpl(CancellableContinuationImpl.kt:227)
        at kotlinx.coroutines.CancellableContinuationImpl.resumeUndispatched(CancellableContinuationImpl.kt:299)
        at kotlinx.coroutines.android.HandlerContext$scheduleResumeAfterDelay$$inlined$Runnable$1.run(Runnable.kt:19)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:201)
        at android.os.Looper.loop(Looper.java:288)
        at android.app.ActivityThread.main(ActivityThread.java:7839)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
2022-05-24 15:28:16.305 11149-11149/org.wordpress.aztec E/WordPress-EDITOR: HTML content of Aztec Editor before the crash:
2022-05-24 15:28:16.307 11149-11149/org.wordpress.aztec E/WordPress-EDITOR: <placeholder id="0" uuid="9e4b3620-8c13-45d0-9fa2-9aa9c946d8e0" type="image_with_caption" src="content://com.android.providers.downloads.documents/document/msf%3A69" caption="Test caption" /><br><placeholder id="123" type="image_with_caption" src="https://filesamples.com/samples/image/jpg/sample_640%C3%97426.jpg" caption="caption of image" uuid="0ab4bb4b-6298-44f1-b9c0-d03889820da9" /><placeholder id="1" uuid="dffcb7dd-e0da-4ff2-bafc-079e225f7ca5" type="image_with_caption" src="content://com.android.providers.downloads.documents/document/msf%3A70" caption="Test caption" /><br><br><placeholder id="2" uuid="17961e36-bb5d-46c5-9750-1b26feaa03b0" type="image_with_caption" src="content://com.android.providers.downloads.documents/document/msf%3A65" caption="Test caption" /><br>This is some test.<br><strong>Bolded text</strong><br><br><blockquote>I lost my doggie to cupcakes</blockquote><br>Hi<br>
2022-05-24 15:28:16.307 11149-11149/org.wordpress.aztec D/WordPress-EDITOR: Below are the details of the content in the editor:
2022-05-24 15:28:16.315 11149-11149/org.wordpress.aztec D/WordPress-EDITOR: ￼¶¶￼¶￼¶¶¶￼¶¶This is some test.¶Bolded text¶¶I lost my doggie to cupcakes¶¶Hi¶¬  length = 78
    >-----------------------------------------------------------------------------<       000 -> 078 : ChangeWatcher
    >-----------------------------------------------------------------------------<       000 -> 078 : ChangeWatcher
    <>                                                                                    000 -> 001 : AztecPlaceholderSpan
    <>                                                                                    000 -> 001 : AztecMediaClickableSpan
       <>                                                                                 003 -> 004 : AztecPlaceholderSpan
       <>                                                                                 003 -> 004 : AztecMediaClickableSpan
         <>                                                                               005 -> 006 : AztecPlaceholderSpan
         <>                                                                               005 -> 006 : AztecMediaClickableSpan
             <>                                                                           009 -> 010 : AztecPlaceholderSpan
             <>                                                                           009 -> 010 : AztecMediaClickableSpan
                                   <---------->                                           031 -> 042 : AztecStyleStrongSpan
                                                >---------------------------->            044 -> 073 : AztecQuoteSpanAligned
        >                                                                                  004 -> 004 : AztecVisualLinebreak
     >                                                                                     001 -> 001 : AztecVisualLinebreak
          >                                                                                006 -> 006 : AztecVisualLinebreak
              >                                                                            010 -> 010 : AztecVisualLinebreak
                                                                            >              072 -> 072 : AztecVisualLinebreak
    >-----------------------------------------------------------------------------<       000 -> 078 : TextKeyListener
    >-----------------------------------------------------------------------------<       000 -> 078 : SpanController
                                                                                 <         077 -> 077 : START
                                                                                 <         077 -> 077 : END
2022-05-24 15:28:16.323 11149-11149/org.wordpress.aztec I/Process: Sending signal. PID: 11149 SIG: 9
```

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
planarvoid: I'll revert the `MainActivity` changes 👍

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
planarvoid: Good point 👍, I've simplified it into one line

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
planarvoid: good point, that's definitely more reliable 👍

planarvoid:
thanks for the detailed review @AmandaRiu ! 

I've fixed all the issues. The only thing I couldn't reproduce is that text behind a newly added placeholder. Could you please check if it's still happening? 

I've removed changes to the `MainActivity` so if you want to test the changes, you have to register `PlaceholderManager` to the `AztecText` and ideally turn on `AztecText.addMediaAfterBlocks`

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
```diff
@@ -0,0 +1,357 @@
...
+
+    /**
+     * Call this method to manually insert a new item into the aztec text. There has to be a adapter associated with the
+     * item type.
+     * @param type placeholder type
+     * @param attributes other attributes passed to the view. For example a `src` for an image.
+     */
+    fun insertItem(type: String, vararg attributes: Pair<String, String>) {
+        val adapter = adapters[type]
+                ?: throw IllegalArgumentException("Adapter for inserted type not found. Register it with `registerDrawer` method")
```
AmandaRiu: np: error message should read `registerAdapter`

AmandaRiu (commented):
Hi @planarvoid ! This is definitely a huge improvement and I feel like each round we are getting closer! The placeholder no longer shows up in source viewer so that’s awesome! Also everything seems to be loading much nicer - though I do have issues where the image inside a placeholder is cut off. I turned on the **Show Layout Bounds** option in Developer Settings when this happened so you can see what it looks like:

Normal | Cut Off
-- | --
Screen Shot 2022-05-25 at 8 10 00 PM|Screen Shot 2022-05-25 at 8 08 38 PM

Once I had **Show Layout Bounds** enabled I noticed a new issue (well not new, but new for me). I have noticed that it’s difficult to get the cursor to appear in the editor when a placeholder is present. This most noticeable when you start with an empty editor and add a new image, then try to add text. Normally just tapping in the editor is enough, but with the placeholder present you have to tap in very specific places before or after the image in order to get a cursor - this is definitely a problem. With the layout bounds visible it’s clear why this is, the placeholder container is huge and each new placeholder adds another layer to the UI which you see because the pink gets darker. This can be bad for performance and may be the reason why images eventually get cut off. You can see it in this recording:


https://user-images.githubusercontent.com/5810477/170393145-ce112186-f1ad-4c0b-aefd-e0cc7aafb000.mov



To compare, here’s a recording of the regular example text with images and everything - there are no UI elements layering over each other:


https://user-images.githubusercontent.com/5810477/170393166-c6505744-9dec-4302-b738-8552d996a83b.mov



I’m not sure what’s going on here. I tried to pick it apart with the layout inspector but didn’t have much luck. In fact it seemed like the placeholder containers were just the right size. I did a session where I just kept adding images and you can see the layers just getting thicker and thicker. I’m not 100% certain this is a problem - it could just be due to the fact that you’re adding `LinearLayouts` over the original editor container, but I do find it interesting that with every new placeholder the scroll view expands. You can also see in this video how weird it is to add text after adding the first placeholder:



https://user-images.githubusercontent.com/5810477/170393395-0ef5b1ec-51af-479d-a4fc-edd157c9c4c6.mov




Everything else seems to be working nicely. I didn’t run into any issues with text popping up behind the placeholder like I did before. I did notice that there were two review comments left from the last round that maybe you missed - I resolved the rest:

- [This comment](https://github.com/wordpress-mobile/AztecEditor-Android/pull/977#discussion_r880873028) from the last PR Review still requires attention.
- [This comment](https://github.com/wordpress-mobile/AztecEditor-Android/pull/977#discussion_r880880560) from the last PR Review still requires attention. 

Also left some minor code comment. I think this PR is really close. Really nice work!

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ImageWithCaptionAdapter.kt
planarvoid: this one should be already fixed. The issue was that the create view method was called multiple times (because of undo) and I wasn't correctly clearing the views and readding them. On `undo` the whole HTML is redrawn. It should be ok now. Let me know if you experience the crash again.

planarvoid:
thanks for testing and review @AmandaRiu 

I've added an `onTouchListener` to the placeholder view which returns `false` by default. This passes the click to the editor and correctly places the cursor. The `adapter` can add `onClickListener` if it wants to handle the click instead. 

I've looked into the layout bounds and I think that the pink background is showing the margins. I can't think of anything else tbh. The `LinearLayout` has a correct size and there doesn't seem to be anything else above the views. However, if we ever experience performance issues, I have a possible solution (which seems like an overkill now). We can have `onScrollListener` on the scroll view and take that + the height of the view and only draw views which are visible to the user. The other ones could be hidden. We should keep performance in mind when we're working on the other views.

I've also made some additional changes to the view proportions. It didn't work in 100% of cases.

AmandaRiu (approved):
Wow, this is working BEAUTIFULLY! Fantastic work on this - I'm so excited! :shipit: I'll leave it to you to merge since you mentioned something in slack about the majority of the code being in our Day One repo - not sure if you meant some of this code or just the usage of it.

PR #978, opened 2022-04-11
Fix issue when adding image when no blocks are present

planarvoid:
### Fix
There is an issue that when there are no blocks in the selection and the user wants to add an image. In that case the selection is set to 0 and the image is added to the beginning of the editor. 

I've added unit tests to cover this functionality as well

### Test
1. Turn on `aztec.visualEditor.addMediaAfterBlocks()` in the `MainActivity`
2. Set `EXAMPLE` to something like `<h1>Test 1</h1>\nBody`
3. Run the app
4. Click at the end of `Body`
5. Add an image
6. Check it gets added after the `Body` and not at the beginning of the editor

### Review
@AmandaRiu 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
@@ -182,7 +182,6 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
             }
         } else {
             ssb.append("\n")
-
             val ssbLength = ssb.length
             editableText.getSpans(position, position + ssbLength, IAztecBlockSpan::class.java).filter {
                 it !is AztecMediaSpan && editableText.getSpanStart(it) == position
@@ -238,6 +237,12 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
             if (spanEnd > position) {
                 position = spanEnd
             }
+        }
+        if (position <= 0 && selectionEnd != 0) {
+            position = editableText.indexOf("\n", selectionEnd)
+        }
+        if (position <= 0 && selectionEnd != 0) {
+            position = editableText.length
         }
         return position
     }
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
@@ -231,6 +231,9 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
     }
 
     private fun getEndOfBlock(): Int {
+        if (selectionStart == 0 && selectionEnd == 0) {
+            return 0
+        }
         var position = 0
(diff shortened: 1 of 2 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
```diff
@@ -239,6 +238,12 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
                 position = spanEnd
             }
         }
+        if (position <= 0 && selectionEnd != 0) {
+            position = editableText.indexOf("\n", selectionEnd)
+        }
+        if (position <= 0 && selectionEnd != 0) {
+            position = editableText.length
+        }
```
AmandaRiu: At first I thought the duplicate if-statements were a bug, but after some testing I realized really quickly why this is needed. I think we can simplify this a bit and also make it clear that this behavior is intentional:

```suggestion
        if (position <= 0 && selectionEnd != 0) {
            // If the text contains "\n" return that as the position, else set the position to the end of the text
            position = editableText.indexOf("\n", selectionEnd).takeIf { it >= 0 } ?: editableText.length
        }
```

AmandaRiu (approved):
@planarvoid One thing I did notice is that if nothing is selected then the image is added to the end of the last block span detected which seems fine until you have regular text following the block. For example, if I set the `EXAMPLE` text to `"<h1>Test 1</h1>\nMy Test"`, then run the app and just tap to add a photo  (without tapping inside the editor so `selectionEnd` is 0), then the image is added between the Heading and the body. This is an edge case I realize and I'm not sure what the fix would be in that case other than either defaulting the `position` to either 0 to add the image to the top of the document, or to `editableText.length` to return the position at the end of the document. I think the second option would keep everything in line with expected functionality for Aztec, but either way I'm not sure it's a big enough problem to address in this PR. 

I also added a code comment for your review. Everything works great so approved, but will wait to merge in case you are still online to look at this real quick.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
planarvoid: looks good, thanks for the suggestion 👍

planarvoid:
thanks for the review @AmandaRiu 

> One thing I did notice is that if nothing is selected then the image is added to the end of the last block span detected which seems fine until you have regular text following the block. For example, if I set the EXAMPLE text to "<h1>Test 1</h1>\nMy Test", then run the app and just tap to add a photo (without tapping inside the editor so selectionEnd is 0), then the image is added between the Heading and the body. This is an edge case I realize and I'm not sure what the fix would be in that case other than either defaulting the position to either 0 to add the image to the top of the document, or to editableText.length to return the position at the end of the document. I think the second option would keep everything in line with expected functionality for Aztec, but either way I'm not sure it's a big enough problem to address in this PR.

Good catch 👍 , I've added an edgecase solution just for this usecase (if selection start and end are 0, insert at the beginning of the entry).

AmandaRiu:
Changes look good! I'll merge as soon as CI has passed 👍🏻

PR #979, opened 2022-04-15
Improve nested lists behaviour

planarvoid:
### Fix
The nested lists were not previously working properly. In this PR I'm fixing the behaviour. 
- the clicked nested list is now correctly highlighted in the toolbar
- you can change type of the selected nested list (without changing the parent list type)
- you can change type only of the parent list type, not of the child
- you can change type of both if you select both

### Test
1. Call `toolbar.setToolbarItems(ToolbarItems.BasicLayout(ToolbarAction.LIST, ToolbarAction.INDENT, ToolbarAction.OUTDENT))` in the `MainActivity` to enable indents and list operations
2. Run the app
3. Add a list
4. Try toggling its type
5. Add nested lists by `indent` 
6. Try toggling type of the nested list
7. Try toggling type of the parent list
8. Click around and notice the toolbar `list` icon is highlighted according to the selected list

### Review
@AmandaRiu 

https://user-images.githubusercontent.com/1079756/163550607-316bfc3b-5595-4f39-bc3d-3a04b19f38b4.mp4


AmandaRiu (changes requested):
@planarvoid Changes look good and list manipulation definitely seems better, but I did run into funkiness with the enabling/disabling of the indent/outdent buttons:

https://user-images.githubusercontent.com/5810477/163886633-ae22fa6e-b3cb-4b36-aa60-05f026e7cd4e.mp4

As well as with highlighting list items across indentations and attempting to indent further:


https://user-images.githubusercontent.com/5810477/163886711-a0ca489c-fdbd-4c82-8aeb-72baadfcb149.mp4

And some items losing their list status:


https://user-images.githubusercontent.com/5810477/163886838-6025dfbb-bdc0-478b-ae61-e93211474a40.mp4

planarvoid:
Thanks for the review @AmandaRiu, all items were great catches! I have fixed all of them in the last 3 commits and I think everything works well

AmandaRiu (commented):
What an incredible improvement @planarvoid ! I tested a bunch and only found one issue, a scenario that leads to an ANR. I was able to replicate the ANR with the following steps:

1. Create an ordered list that looks like this:

Screen Shot 2022-05-10 at 10 19 40 PM

2. place cursor in "Four" and tap the button to outdent that item so the three middle list items are all at the same indentation.
3. Select items "three", "four" and "five" and tap the button to outdent them so that all list items will be in the same level. This causes the app to freeze and eventually an ANR will display.

Here's a video:

https://user-images.githubusercontent.com/5810477/167756095-c21eaf85-6b7e-431e-98c2-c60b43e3f7ce.mp4

planarvoid:
Good catch @AmandaRiu ! It was caused by a solution I've used previously to determine the following list item (to find out the desired behaviour of the outdent). I've fixed it in [8f1f87c](https://github.com/wordpress-mobile/AztecEditor-Android/pull/979/commits/8f1f87cfbf557b591b0f77425465b4d4b3dcfade)

AmandaRiu (approved):
Works beautifully!!! Very nice work @planarvoid !!! 🥳 

:shipit:

PR #984, opened 2022-05-30
Enable media within paragraph blocks

planarvoid:
Previously we've added the option to `addMediaAfterBlocks`. This works well with almost all the blocks but it doesn't work nicely with paragraphs. We want to be able to add images within paragraphs and not only after the currently selected paragraph. This is important for paragraphs with linebreaks (where the image should be logically positioned at the linebreak, not after the paragraph). 

You can test this by setting `EXAMPLE` to 
```
<p>test<br>
<br>
test</p>
```

### Test
1. Turn on `addMediaAfterBlocks` on the visual editor
2. Run the app with the `EXAMPLE` set from the code above
3. Click between the lines
4. Add an image
5. Notice it gets added between the lines, not at the end of the document

### Review
@danilo04 


danilo04 (approved):
Hi @planarvoid. The changes are working as expected. Thanks!!

PR #985, opened 2022-06-01
Insert text when current text is empty to preserve formatting when pasting to empty edit text

planarvoid:
### Fix
In this PR I'm replacing the functionality when pasting text into the empty aztec editor. Currently this triggers `setText` which removes all the formatting spans. However, I think it's a valid use case to turn on formatting and paste text and expect the pasted text to be formatted. Otherwise the functionality remains unchanged

### Test
1. Set `EXAMPLE` field in the `MainActivity` to `<h1></h1>`
2. Run the app
3. Paste text to the editor
4. Notice it gets formatted with `h1` (previously it didn't keep the formatting)

### Review
@AmandaRiu 


AmandaRiu (approved):
Works like a charm 👍🏻 

:shipit:

PR #986, opened 2022-06-08
Add option to set placeholder size based on the item shown

planarvoid:
### Fix
This PR adds an option to the placeholder sizing where you can define placeholder size based on the item (and the AztecAttributes). This is good for things like images and videos which might not always be in landscape/portrait and showing them differently might result in a non-perfect experience 

### Test
1. Set up `MainActivity` with the placeholder manager and the `ImageWithCaptionAdapter`. The steps are:
  - Create `PlaceholderManager` in the `onCreate` method in `MainActivity` passing in the `visualEditor`
  - Call `placeholderManager.registerAdapter(ImageWithCaptionAdapter())` to register the `ImageWithCaptionAdapter`
  - Call `aztec.addPlugin(placeholderManager)` 
  -  Call `aztec.addOnMediaDeletedListener(placeholderManager)`
2. Modify the `getWidth` and `getHeight` methods in order to change them based on the item shown. For example:
  - Set `EXAMPLE` to ```<placeholder type=\"image_with_caption\" src=\"https://file-examples.com/storage/febc474733629f43d9f078c/2017/10/file_example_JPG_100kB.jpg\" caption=\"1\"><br><placeholder type=\"image_with_caption\" src=\"https://file-examples.com/storage/febc474733629f43d9f078c/2017/10/file_example_JPG_100kB.jpg\" caption=\"2\">```
  - Modify `getHeight` method in `ImageWithCaptionAdapter` so that it returns different value based on the `caption` attribute`
3. Run the app
4. Notice the images are drawn differently

### Review
@danilo04 

### DIFF
--- a/media-placeholders/src/main/java/org/wordpress/aztec/placeholders/AztecPlaceholderSpan.kt
+++ b/media-placeholders/src/main/java/org/wordpress/aztec/placeholders/AztecPlaceholderSpan.kt
@@ -15,14 +15,14 @@ class AztecPlaceholderSpan(
         attributes: AztecAttributes = AztecAttributes(),
         onMediaDeletedListener: AztecText.OnMediaDeletedListener? = null,
         editor: AztecText? = null,
-        private val adapter: PlaceholderManager.PlaceholderAdapter) :
+        private val adapter: PlaceholderManager.PlaceholderAdapter,
+        override val TAG: String) :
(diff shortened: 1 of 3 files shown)


danilo04 (approved):
Hi @planarvoid. this works as expected. Thank you for making these changes!!!

PR #988, opened 2022-06-13
Stop using custom getLineForOffset

planarvoid:
### Fix
There is an issue where there is a multiline text element before a placeholder. The custom method that was calculating line for offset doesn't take that into account. I've replaced it with the default `getLineForOffset` and it seems to work well now.

In addition to this I've also made some of the functions in the `PlaceholderManager` suspended to improve performance and allow async computation

### Test
1. Enable placeholders and add a multiline text followed by a placeholder to the `Example` field
2. Run the app
3. Notice the placeholder is drawn correctly

### Review
@danilo04 


danilo04 (approved):
This works as expected @planarvoid. Thank you for fixing the bug and also adding support to prefetch the views concurrently. I was thinking about how to do this too but were a step ahead 🎉 !!

PR #989, opened 2022-06-15
Fix media removal and improve coroutine handling

planarvoid:
### Fix
The media removal is not currently added to history. That's fixed in this PR. Also the logic to remove media span is moved to the `AztecText` so that it's reusable. I've also fixed the coroutine usage and made the `PlaceholderManager` a coroutine scope.

### Test
1. Set up `PlaceholderManager` in the main activity with the `ImageWithCaptionAdapter`
2. Make the `EXAMPLE` contain a `placeholder` that you can identify
3. Add `removeItem` action to something you can control (for example the camera button)
4. In the `removeItem` identify the placeholder you want to remove
5. Run the app
6. Remove the placeholder
7. Notice it gets removed correctly

### Review
@danilo04 


danilo04 (approved):
Nice work @planarvoid! Tested the changes and they are working properly :shipit: !

PR #990, opened 2022-06-16
Release v1.5.9

planarvoid:
This PR will be used to test the changes in the new release in all the repositories that use `Aztec` editor. Do not merge it until all the related PRs are approved.

To target this release use `990-d5be2428ed6f8db33a3e6ec3c9df5d09ea26fc82` as the aztec version.

### Review
@mchowning 
@danilo04 

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,42 @@ When switched, this editor will always add media and horizontal rule after the c
 aztecText.addMediaAfterBlocks()
 ```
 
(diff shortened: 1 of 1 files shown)


derekblank (approved):
I successfully tested this Aztec reference on WPAndroid with a local Gutenberg reference to Aztec -- looks good to me. 👍

Issue #931, opened 2020-11-20
Incorrect rendering / overlapping text / "stuck" lines when inserting list/heading above text


tipa:
### Expected
I can make lists, headers etc. and the text re-layouts itself without issues

### Observed
In certain situation, lines below a header, list or bold text are not rendered correctly but on the incorrect vertical position.

### Reproduced
1. Start demo app
2. Type a few plain text lines on the top
3. Make first line to list
4. (optional) type text to force line break on line or press enter button to generate more list items

### Comments
- I tried to find what causes this problem for multiple hours, whiteout success.
- The issue can also be observed when changing text to a heading (instead of a list) or when converting text to bold text - so anything that changes the line height...?
- It seems like the problematic lines are "stuck" on a certain vertical position
- The problem can be "resolved" by calling "refreshText()" on the visual editor or by placing the cursor on the line that is incorrectly rendered 

#### Tested
OnePlus 5 on Android 10 with v1.0
Pixel 5 on Android 11 with v1.0
Android Simulator on Android 10 & 6 with v1.0

Video: [recordedVideo2020-11-20-135142.zip](https://github.com/wordpress-mobile/AztecEditor-Android/files/5573549/recordedVideo2020-11-20-135142.zip)

![Screenshot_20201120-133637](https://user-images.githubusercontent.com/2863198/99801172-e2026180-2b35-11eb-8001-104a265c411a.jpg)
![image](https://user-images.githubusercontent.com/2863198/99801900-0d398080-2b37-11eb-8900-19deba7fbc5e.png)

imthiaz:
Hello

Even we have used aztec library in our application and we are also facing the same issue when user applies ordered list option for one of the middle lines.

Not sure if they will fix this issue. But if you have found a fix let me know please as it would help us improve the performance of our application

PR #992, opened 2022-07-08
Advanced styling of headings

planarvoid:
### Fix
In this PR I'm introducing advanced options for styling some elements. For example headings now can be sized and changed each one differently. Headings now support size and color. 

### Test
1. Modify the AztecText style in MainActivity to use `headingNFontSize` or `headingNFontColor` where `N` stands for heading types between `one` and `six`
2. Run the app
3. Notice the heading styling

### Review
@[USER_NAME]

### DIFF
--- a/app/src/main/res/layout/activity_main.xml
+++ b/app/src/main/res/layout/activity_main.xml
@@ -39,11 +39,7 @@
                 android:scrollbars="vertical"
                 android:imeOptions="flagNoExtractUi"
                 aztec:historyEnable="true"
-                aztec:historySize="10"
-                aztec:headingOneFontSize="24sp"
-                aztec:headingOneFontColor="@color/blue_light"
-                aztec:headingThreeFontSize="16sp"
-                aztec:headingThreeFontColor="#F09389"/>
+                aztec:historySize="10"/>
 
             <org.wordpress.aztec.source.SourceViewEditText
                 android:id="@+id/source"
--- a/app/src/main/res/layout/activity_main.xml
+++ b/app/src/main/res/layout/activity_main.xml
@@ -39,7 +39,11 @@
                 android:scrollbars="vertical"
                 android:imeOptions="flagNoExtractUi"
                 aztec:historyEnable="true"
-                aztec:historySize="10"/>
+                aztec:historySize="10"
+                aztec:headingOneFontSize="24sp"
+                aztec:headingOneFontColor="@color/blue_light"
+                aztec:headingThreeFontSize="16sp"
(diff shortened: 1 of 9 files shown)


app/src/main/res/layout/activity_main.xml
```diff
@@ -39,7 +39,11 @@
                 android:scrollbars="vertical"
                 android:imeOptions="flagNoExtractUi"
                 aztec:historyEnable="true"
-                aztec:historySize="10"/>
+                aztec:historySize="10"
+                aztec:headingOneFontSize="24sp"
```
khaykov: 24sp is visually very similar (if not the same) to H2. Maybe we can make it a bit bigger to differentiate better in demo app?

khaykov (commented):
I haven't touched the Aztec for a while, but from what I can tell the size and color of Heading seems to work as expected. But it looks like something funky is happening with paddings? If you add a text after H1, and convert it to, say, H3, the paddings below H3 shift in weird way. This issue is present to some degree in trunk, with bigger headings, but behavior on the gif bellow is recent development:

[Image from Gyazo](https://gyazo.com/e27607eab74065490120fa533ebee6af)

Comparison using same text sizes:
[Image from Gyazo](https://gyazo.com/f4e0e89a974050cad90d740a0032290a)

Also, it looks like there are some changes around paragraph margin/padding, but I'm not sure what they affect here?

planarvoid:
you're right @khaykov . However, I've tested `trunk` and it's happening there as well so I don't think it's related to these changes. 

I have no idea why the edit text isn't redrawn when the heading is added/removed. I've tried a few approaches and nothing redraws it so I've added a hacky fix which adds and removes a new line when a heading is applied/removed. It works well so let me know what you think!

app/src/main/res/layout/activity_main.xml
planarvoid: This was left here by accident, I've removed it 👍

khaykov:
@planarvoid Got it! While fix definitely helps avoid some crazy situations, I think it is a bit too hacky, and I can't really be sure it will not lead to some weird behavior down the line :) I think it would be a good idea to have a dedicated effort on this one, since it looks like other elements apart from heading are also affected by it, as described in #931.

planarvoid:
yeah, it just ignores the updated line height for all the elements. It's just much more prominent for the headings because the line height there is much bigger. I'll revert the change and we can leave it for a separate fix 👍

khaykov (approved):
Thanks for the changes, @planarvoid ! Looks good to me 👍

PR #993, opened 2022-07-19
Release 1.6.0

planarvoid:
This PR purpose is to test the new version with the other projects using this library. 

In order to test it please change the aztec version in your project to `993-80d61644d68813b3f68c6bcc11075cfcc08829a5`

### Review
@derekblank 


khaykov (approved):
Look good 👍

PR #994, opened 2022-07-19
Simplify containsPreformat method

planarvoid:
### Fix
The `containsPreformat` method doesn't work well when the preformat doesn't contain any code. In this PR I'm simplifying it so that it works in all cases. I don't know why it was originally so complicated. I think the original behaviour doesn't have any benefits.

### Test
Enable preformat action by adding it to toolbar actions
```
        val defaultBasicLayout = BasicLayout(
                ToolbarAction.PREFORMAT,
...
```
to make it easier set the `EXAMPLE` in the `MainActivity` to something like this: 
```
"""
<pre></pre>
<p>paragraph 1</p>
""".trimIndent()
```

1. Run the app
2. Click into the preformat line
3. Trigger the toolbar preformat action
4. Notice the `preformat` disappears and is not duplicated

### Review
@khaykov 


khaykov:
@planarvoid At this point I don't remember all the use cases for the complicated method, but I think it's related to line breaks (and maybe something else.

On the left is this branch, on the right is trunk. As you can see the cursor is positioned below the pre block on a new line. In trunk it's correctly detecting that it's not a pre (icon is not highlighted in toolbar) and in this branch it detects pre incorrectly (icon is highlighted).
[Image from Gyazo](https://gyazo.com/74565c397e0f634032af59d9b859cee2)

planarvoid:
Thanks for the review @khaykov . I've returned the original method but it was broken as well so I've fixed it. The previous The previous solution didn't support something like this: 
```
<p>First line</p>
<pre>Pre line 1<br></pre>
<p></p>
<p>Last line</p>
```
If you had an empty line at the end of PRE block, the `PRE` button wasn't highlighted when you clicked into it. I think it should be fixed now.

I've also uncommented and fixed all the preformat tests.

khaykov (approved):
Thanks for the fixes! Looks good from what I can tell, but looks like there are some broken tests?

PR #995, opened 2022-07-21
Clear references in AztecExceptionHandler.kt

planarvoid:
### Fix
`AztecExceptionHandler` is leaking `AztecText`. In this PR I'm replacing the references with `WeakReference` so that the reference to the `AztecText` is not held by the exception handler longer than necessary

### Review
@khaykov 


khaykov (approved):
Smoke tested this one, and no issues so far 👍

PR #998, opened 2022-08-02
Use adapter height when drawing a placeholder

planarvoid:
### Fix
This PR fixes an issue when the placeholder size was different (by a little bit) than what was defined in the adapter. This was mostly visible on smaller placeholders. I think only the last placeholder of the document was affected. 

In addition it fixes the width so that the cursor doesn't overlap the placeholder

### Test (it's better to test this with the PR where it's used)
1. Setup placeholders
2. Add 2 placeholders to the `EXAMPLE` 
3. Add logging of actual height to the code
4. Check the height is fixed

### Review
@khaykov 

### DIFF
--- a/media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
+++ b/media-placeholders/src/main/java/org/wordpress/aztec/placeholders/PlaceholderManager.kt
@@ -160,7 +160,6 @@ class PlaceholderManager(
         aztecText.getLocationOnScreen(parentTextViewLocation)
         val parentTextViewTopAndBottomOffset = aztecText.scrollY + aztecText.compoundPaddingTop
 
-
         val adapter = adapters[type]!!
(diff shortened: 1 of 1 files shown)


Teelair (approved):
I tested this using Day One Android on the `fix/too-big-audio-player` branch. 

To test, I made four entries each containing photos, audio files, videos, and a PDF. I took screenshots of the entries looked in both read and edit mode before applying this PR's changes, and then after. 

The most noticeable issue where audio placeholders were visibly larger than others has been fixed, but also other placeholders have had their size slightly altered. It is not by much so I feel like this is not a big deal within the context of Day One, but because I am not sure of what all is using Aztec, I'll leave an approval and leave it up to you to merge this if that sounds okay to you @planarvoid!

:shipit:

khaykov (approved):
Approving since @Teelair gave a 🟢

PR #1001, opened 2022-08-09
Add UrlPastePlugin.kt to handle pasted links

planarvoid:
In this PR I'm adding a plugin option to handle pasted text. This could be useful to override the default behaviour of handling any links. I'm also adding `UrlPastePlugin` which makes sure that when you paste an URL over selected text, it makes the whole thing into a link instead of replacing the selected text with a link.

Before you start make sure you add the following code to `aztec` in your `MainActivity` 
```
aztec.addPlugin(UrlPastePlugin())
```

### Test
1. Run the app
2. Copy link from your browser
3. Long press within the text
4. Paste content of your clipboard
5. Notice the pasted text is a link
6. Click on the link
7. Notice it's recognized by the toolbar as a link

### Test
1. Run the app
2. Copy link from your browser
3. Select a word in the text
4. Long press on it
5. Paste content of your clipboard
6. Notice the previously selected text is now a link
7. Click on the link
8. Notice it's recognized by the toolbar as a link 

### Test
1. Run the app
2. Copy something that's not a link
3. Long press within the text
4. Paste content of your clipboard
5. Notice the pasted text is not a link

### Test
1. Run the app
2. Copy something that's not a link
3. Select a word in the text
4. Long press on it
5. Paste content of your clipboard
5. Notice the selected text is replaced by the pasted text and is not a link

### Review
@khaykov 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -74,6 +74,7 @@ import org.wordpress.aztec.handlers.ListItemHandler
 import org.wordpress.aztec.handlers.PreformatHandler
 import org.wordpress.aztec.handlers.QuoteHandler
 import org.wordpress.aztec.plugins.IAztecPlugin
+import org.wordpress.aztec.plugins.ITextPastePlugin
 import org.wordpress.aztec.plugins.IToolbarButton
 import org.wordpress.aztec.source.Format
 import org.wordpress.aztec.source.SourceViewEditText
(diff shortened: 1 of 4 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -1875,7 +1878,14 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
                 else clip.getItemAt(0).coerceToHtmlText(AztecParser(alignmentRendering, plugins))
 
                 val oldHtml = toPlainHtml().replace("<aztec_cursor>", "")
-                val newHtml = oldHtml.replace(Constants.REPLACEMENT_MARKER_STRING, textToPaste + "<" + AztecCursorSpan.AZTEC_CURSOR_TAG + ">")
+                val pastedHtmlText = plugins.filterIsInstance<ITextPastePlugin>().fold(textToPaste) { acc, plugin ->
```
khaykov, line 1881: What is `acc` in this context?

khaykov (commented):
Thanks for the awesome feature, @planarvoid ! It works surprisingly well! One thing that is a bit weird is pasting URL over multiline text:

[Image from Gyazo](https://gyazo.com/d4d1355e0e2008a71b7f61f57c7ffe48)

It removes linebreaks. Marking same text as URL using toolbar button works as expected.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
planarvoid: accumulator. I can use a better name. This is the default one 👍

khaykov (approved):
Approving it 👍 As we discussed on Slack, pasting over multiple lines is rare case, and will be addressed separately.

PR #1002, opened 2022-09-20
Updated span removal logic when using exclusiveBlocks

khaykov:
This PR fixes issue with adding new block spans right next to the existing ones, when `exclusiveBlocks` is set to `true`.

To test this PR and the original issue you need to add `aztec:exclusiveBlocks="true"` to `AztecText` [layout](https://github.com/wordpress-mobile/AztecEditor-Android/blob/trunk/app/src/main/res/layout/activity_main.xml#L28).

When `exclusiveBlocks` is set to `true` the block are replaced when other block style is applied to them (normal behavior is to allow multiple block style to be applied).

Issue was caused because the range we used to detect spans we should remove included the span on the line before the cursor. I used the range logic from the toolbar highlight code, which I believe is more accurate.

Steps to reproduce original issue:

1. Create a list with one element in it.
2. Press enter to add a second element element, and press enter again to remove it a start a new line.
3. Add a quote.
4. Notice that the list styling has disappeared.

[Image from Gyazo](https://gyazo.com/1931a4f46ff0decf0c688fe5aba59525)

### Test
1. Create a list with one element in it.
2. Press enter to add a second element element, and press enter again to remove it a start a new line.
3. Add a quote.
4. Notice that the quote is added correctly and list style remains.

This behavior affects block styles, like Lists, Quotes and Headings.
Try testing it in various ways - applying block styles in the middle of multiline block element, at the beginning and at the end.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/BlockFormatter.kt
@@ -245,8 +245,19 @@ class BlockFormatter(editor: AztecText,
         }
         val selectionStart = editor.selectionStart
 
+        var newSelStart = if (selectionStart > selectionEnd) selectionEnd else selectionStart
+        var newSelEnd = selectionEnd
+
+        if (newSelStart == 0 && newSelEnd == 0) {
+            newSelEnd++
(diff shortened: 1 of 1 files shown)


AmandaRiu (approved):
Tested a bunch of different scenarios and all seem to be working as expected. Nice work! :shipit:

PR #1005, opened 2022-09-30
Add onDraw plugin and header leading margin option

khaykov:
This PR adds `IOnDrawPlugin` that is used to plugin into drawing cycle of AztecText.

I also added a leading margin option for Header spans.

`IOnDrawPlugin` are abstract, so there is nothing particular to test - just confirm that it looks ok.

### Test
1. Change leading margin value of any heading using `headingXYZLeadingMargin` style and make sure it is working.
2. Make sure leading margin is not visible when it is not set.


AmandaRiu (approved):
Works great! :shipit:

PR #1006, opened 2022-10-03
Updates: target sdk version to 31

### LINKED ISSUE #1003: Update `targetSdkVersion` to 31 for Android 12 Migration
Parent #https://github.com/wordpress-mobile/WordPress-Android/issues/16062

As part of the migration of WP Android to Android 12, the `targetSdkVersion` on the dependent libraries can be updated to 31. The [Wordpress Android](https://github.com/wordpress-mobile/WordPress-Android) had a `targetSdkVersion` of 30 which now has been updated to 31 in this PR - https://github.com/wordpress-mobile/WordPress-Android/pull/17153. So in order to keep the `targetSdkVersion` to be consistent with the apps, this change needs to be taken up.

AjeshRPai:
Closes #1003 

This PR updates the `targetSDKversion` to 31 for the Android 12 migration

Test Instructions
- Install the app from the Parent https://github.com/wordpress-mobile/WordPress-Android/pull/17153
- Make sure the functionalities provided by the library work as expected

[x] If there are new strings that have to be translated, I have added them to the client's `strings.xml` as a part of the integration PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,7 @@ ext {
     jSoupVersion = '1.11.3'
     wordpressUtilsVersion = 'trunk-1ed207c03d2242b6fc3d74f9e388e9163cbc82a6'
     espressoVersion = '3.0.1'
-    commonTargetSdkVersion = 30
+    commonTargetSdkVersion = 31
     commonMinSdkVersion = 21
 
     aztecProjectDependency = project.hasProperty("aztecVersion") ? "org.wordpress:aztec:${project.getProperty("aztecVersion")}" : project(":aztec")
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -17,6 +17,7 @@
 
         <activity
             android:name=".MainActivity"
+            android:exported="true"
             android:windowSoftInputMode="adjustResize|stateHidden"
             android:theme="@style/MainActivityTheme" >
 
--- a/aztec/build.gradle
+++ b/aztec/build.gradle
@@ -60,9 +60,9 @@ dependencies {
 
     testImplementation "junit:junit:$jUnitVersion"
     testImplementation "org.robolectric:robolectric:$robolectricVersion"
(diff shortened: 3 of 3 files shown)


fluiddot (approved):
I've tested the editor both as a standalone app and using https://github.com/wordpress-mobile/WordPress-Android/pull/17153 branch in WP-Android, I haven't encountered any issue so looks like these changes are good to go 🎊 .

build.gradle
```diff
@@ -66,7 +66,7 @@ ext {
     jSoupVersion = '1.11.3'
     wordpressUtilsVersion = 'trunk-1ed207c03d2242b6fc3d74f9e388e9163cbc82a6'
     espressoVersion = '3.0.1'
-    commonTargetSdkVersion = 30
+    commonTargetSdkVersion = 31
```
ParaskP7, line 69: 👋 @AjeshRPai !

Maybe we can take this opportunity, in this PR, or, most probably, in a separate PR afterwards (as to not block this one) to do the following as well:

- Rename the `commonTargetSdkVersion` and `commonMinSdkVersion` to `targetSdkVersion` and `minSdkVersion` respectfully to much our usual such naming convension. PS: The 
[about-automattic-android](https://github.com/Automattic/about-automattic-android) also has it's own such naming convention too, which utilizes underscores instead of a camel-case naming pattern, maybe we should consider updating that at some point too.
- Try to extract and then update the `compileSdkVersion` to `31` on all modules (libs and app included), I am seeing it being `28` and I wonder if we can easily update it to `31`, which will give us another additional safety net when testing such upgrades in the future.

Wdyt? 🤔

AjeshRPai: @ParaskP7 
I agree. I wish I had catched that change earlier. I will add it to my to-do and take it up in a separate PR since I don't want to block this PR. WDYT?. 🤔 
Thanks for pointing it out.

ParaskP7: Awesome, I agree, thank you so much Ajesh! 🙇

PR #1007, opened 2022-10-06
Feature/add styling options to task list

khaykov:
This PR adds option to specify checked task list item color and if it should use strikethrough.

I tried couple of approaches, but the easiest one was to add styling to the `AztecListItemSpan`. This requires wiring the styling data to a lot of places, but it also opens up potential for styling list items for other purposes.

### Test
You can add the task list button to the toolbar by pasting following code somewhere [here](https://github.com/wordpress-mobile/AztecEditor-Android/blob/trunk/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L394):

```
toolbar.setToolbarItems(ToolbarItems.BasicLayout(
        ToolbarAction.TASK_LIST
))
toolbar.enableTaskList()
```

[Image from Gyazo](https://gyazo.com/45effa99808119eab242a5bc4844858e)

1. Confirm that task list in HTML rendered from example has a task list, it works as expected and selected items use grey strikethrough text.
1. Confirm that task list added from toolbar works as expected and selected items use grey strikethrough text.


AmandaRiu (approved):
@khaykov LGTM! Works as advertised. :shipit:

PR #1013, opened 2022-11-09
Improve behaviour of pasting of other types of objects

planarvoid:
### Fix
The current solution is ignoring the other types of pasted objects than text - URL and Intent. In this solution I've created 3 independent interfaces for plugins to override this behaviour for all types of pasted items.

### Test 1 - paste without selected text
1. Run the app without changes
2. Copy an URL from outside of the app
3. Click into an empty line
4. Long press + paste
5. Notice the URL is pasted and is not clickable

### Test 2 - paste over selected text
1. Run the app without changes
2. Copy an URL from outside of the app
3. Long press on a word
4. Paste
5. Notice the URL has replaced the selected text and is not clickable

### Test 3 - paste URL with plugin enabled
1. In the MainActivity add `.addPlugin(UrlPastePlugin())` to the Aztec object
2. Run the app
3. Copy an URL from outside of the app
4. Click on an empty line
5. Paste
6. Notice the URL is pasted but it's a link redirecting to the URL

### Test 3 - paste URL over selected text with plugin enabled
1. In the MainActivity add `.addPlugin(UrlPastePlugin())` to the Aztec object
2. Run the app
3. Copy an URL from outside of the app
4. Long press on a word
5. Paste
6. Notice the selected word is now a link and it's a link redirecting to the URL

### Review
@khaykov 


khaykov (approved):
Works as expected 👍

PR #982, opened 2022-05-05
Update kotlin and other libraries and fix the code

planarvoid:
In this PR I'm updating the core libraries and fixing code after the changes. I don't think there are any breaking changes except of what I fixed. I've run all the tests and tested the app as well and I think it works well. Let me know if You'd like me to split the PR into multiple PRs.

### Test
1. Make sure the app works as expected. 

### Review
@danilo04 

Make sure strings will be translated:

- [x] If there are new strings that have to be translated, I have added them to the client's `strings.xml` as a part of the integration PR.

PR #1015, opened 2022-11-17
Update Kotlin to 1.6.10

ParaskP7:
This PR upgrades Kotlin to `1.6.10`.

FYI: This is primarily done to unblock development with JDK17.

-----

It also includes the following:

- [Remove max perm size option gradle jvmargs from gradle properties.](https://github.com/wordpress-mobile/AztecEditor-Android/commit/ba86630a6c7ccdd0ee4e0258c796969c8fa1466e) This is done because otherwise the Gradle daemon fails to start for JDK17.
- [Update robolectric to 4.9.](https://github.com/wordpress-mobile/AztecEditor-Android/commit/310d583c619c8ed48c1486ce9f165c7b553445d4) This is done because otherwise Unit Tests are failing to run.
- [Remove unnecessary kotlin stdlib dependency from all modules.](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1015/commits/47a77e856f7df95fc41f24d8ac8e0999967c6b14) This is a clean-up commit.

-----

### Test
1. Smoke test the app.
2. Test for build time increases (local & CI).

-----

### Review

@planarvoid and @danilo04 as you were also involved in #982.

-----

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -41,7 +41,6 @@ dependencies {
     implementation project(':wordpress-shortcodes')
     implementation project(':media-placeholders')
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$gradle.ext.kotlinVersion"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinCoroutinesVersion"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlinCoroutinesVersion"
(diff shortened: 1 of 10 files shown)


planarvoid (approved):
LGTM :shipit:

ParaskP7:
Thank you so much for the review and testing @planarvoid , awesome! 🚀 ❤️ 🥇

PR #1016, opened 2022-11-17
[Compile Warnings As Errors] Resolve Warnings & Enable All Warnings as Errors

ParaskP7:
This PR resolves/suppresses all warnings for all modules and then enables all warnings as errors as the default for this project (a27076dad1c437d27e2dbfd5eef68330ea319c26).

-----

Warnings Resolution List:

1. [Resolve kotlin stdlib's to lower case deprecated warnings](https://github.com/wordpress-mobile/AztecEditor-Android/commit/6ba043cf9fdaeb3e61e400fb7c31e1bf3cac5bd1)
2. [Resolve parameter is never and could be renamed to _ warnings](https://github.com/wordpress-mobile/AztecEditor-Android/commit/0e8e308042580f24d9d751cdbbf97e88d3aeef1d)
3. [Resolve variable is never used warning](https://github.com/wordpress-mobile/AztecEditor-Android/commit/811ca5142a5c2fecd8f608f664b89b93f907ee2f)
4. [Resolve type mismatch type warnings](https://github.com/wordpress-mobile/AztecEditor-Android/commit/f0f47f0cc42506e39fa9644584affa2e9362efd8)
5. [Resolve unsafe use of a nullable receiver type warning](https://github.com/wordpress-mobile/AztecEditor-Android/commit/8ce75c252e344801b71fc969251d6fc590c13a4d)
6. [Resolve kotlin stdlib's sum by deprecated warnings](https://github.com/wordpress-mobile/AztecEditor-Android/commit/1e7f4a91ea17344a85c99ca8fb931c8851d407a5)
7. [Resolve name shadowed warning](https://github.com/wordpress-mobile/AztecEditor-Android/commit/96c966864dd93a9bad2dc76e78598853de67dc88)
8. [Resolve parameter supertype naming type warnings](https://github.com/wordpress-mobile/AztecEditor-Android/commit/56892fd9a4bf176dfceaabe76f9c6ee465d3b788)
9. [Resolve non exhaustive when statements when warning](https://github.com/wordpress-mobile/AztecEditor-Android/commit/3c54be7918739a8028203715d271cf3f5b52e30b)
10. [Resolve open has no effect in an object warnings](https://github.com/wordpress-mobile/AztecEditor-Android/commit/6c5296d8d49412d77a99a2143b87ecf41fa09d07)
11. [Resolve type mismatch type warning for app module](https://github.com/wordpress-mobile/AztecEditor-Android/commit/d09302b48be78bb175f813b05bb7dfb2e7c918f4)
12. [Resolve unnecessary safe call type warning for app module](https://github.com/wordpress-mobile/AztecEditor-Android/commit/fcbb63cd4e49e790e1fce45b6ad3c2a5f8fc589f)
13. [Resolve unnecessary safe call type warning for ui tests](https://github.com/wordpress-mobile/AztecEditor-Android/commit/95df07ba93f9a746729492a8820f36a5ad3c8f09)

Warnings Suppression List:

1. [Suppress get color deprecated warning](https://github.com/wordpress-mobile/AztecEditor-Android/commit/c32e3203674c75195a8ae616804068ff379aaf7f)
2. [Suppress variable initializer is redundant warning](https://github.com/wordpress-mobile/AztecEditor-Android/commit/1315ae8ffa784a47ed5749c0aba626b643968d1b)
3. [Suppress android junit4 deprecated warnings for ui tests](https://github.com/wordpress-mobile/AztecEditor-Android/commit/e190bc1f3e850c9f2b08d950eb70b6b1fe04755f)

-----

### Test
1. There is nothing much to test here.
2. Verifying that all the CI checks are successful should be enough.
3. However, if you want to be thorough about reviewing this change, you could quickly smoke test the example `app` and see if it is working as expected.

-----

### Review
@planarvoid and @danilo04 as you were also involved in #982.

-----

### DIFF
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/BaseTest.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/BaseTest.kt
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION")
+
 package org.wordpress.aztec.demo
 
 import android.util.Log
--- a/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/GutenbergCompatTests.kt
+++ b/app/src/androidTest/kotlin/org/wordpress/aztec/demo/tests/GutenbergCompatTests.kt
@@ -328,7 +328,7 @@ class GutenbergCompatTests : BaseTest() {
         val editorPage = EditorPage()
         val audioShortcodePlugin = AudioShortcodePlugin()
         val aztecText = mActivityIntentsTestRule.activity.findViewById<AztecText>(R.id.aztec)
-        aztecText.plugins?.add(audioShortcodePlugin)
+        aztecText.plugins.add(audioShortcodePlugin)
 
         // let's test the plugin works as expected, i.e. it preserves the Gutenberg block structure
         editorPage
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -525,7 +525,7 @@ open class MainActivity : AppCompatActivity(),
         super.onSaveInstanceState(outState)
 
         if (mediaUploadDialog != null && mediaUploadDialog!!.isShowing) {
-            outState?.putBoolean("isMediaUploadDialogVisible", true)
+            outState.putBoolean("isMediaUploadDialogVisible", true)
         }
     }
 
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -253,15 +253,13 @@ open class MainActivity : AppCompatActivity(),
 
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         if (resultCode == Activity.RESULT_OK) {
-            var bitmap: Bitmap
-
             when (requestCode) {
                 REQUEST_MEDIA_CAMERA_PHOTO -> {
                     // By default, BitmapFactory.decodeFile sets the bitmap's density to the device default so, we need
                     //  to correctly set the input density to 160 ourselves.
                     val options = BitmapFactory.Options()
                     options.inDensity = DisplayMetrics.DENSITY_DEFAULT
-                    bitmap = BitmapFactory.decodeFile(mediaPath, options)
+                    val bitmap = BitmapFactory.decodeFile(mediaPath, options)
                     insertImageAndSimulateUpload(bitmap, mediaPath)
                 }
                 REQUEST_MEDIA_PHOTO -> {
@@ -271,7 +269,7 @@ open class MainActivity : AppCompatActivity(),
                     //  to correctly set the input density to 160 ourselves.
                     val options = BitmapFactory.Options()
                     options.inDensity = DisplayMetrics.DENSITY_DEFAULT
-                    bitmap = BitmapFactory.decodeStream(stream, null, options)
+                    val bitmap = BitmapFactory.decodeStream(stream, null, options)
 
                     insertImageAndSimulateUpload(bitmap, mediaPath)
                 }
@@ -287,7 +285,7 @@ open class MainActivity : AppCompatActivity(),
 
                         override fun onThumbnailLoaded(drawable: Drawable?) {
                             val conf = Bitmap.Config.ARGB_8888 // see other conf types
-                            bitmap = Bitmap.createBitmap(drawable!!.intrinsicWidth, drawable.intrinsicHeight, conf)
+                            val bitmap = Bitmap.createBitmap(drawable!!.intrinsicWidth, drawable.intrinsicHeight, conf)
                             val canvas = Canvas(bitmap)
                             drawable.setBounds(0, 0, canvas.width, canvas.height)
                             drawable.draw(canvas)
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecTagHandler.kt
@@ -53,6 +53,7 @@ import org.wordpress.aztec.spans.createTaskListSpan
 import org.wordpress.aztec.spans.createUnorderedListSpan
 import org.wordpress.aztec.util.getLast
 import org.xml.sax.Attributes
+import java.util.Locale
 
 class AztecTagHandler(val context: Context, val plugins: List<IAztecPlugin> = ArrayList(), private val alignmentRendering: AlignmentRendering
 ) : Html.TagHandler {
@@ -75,7 +76,7 @@ class AztecTagHandler(val context: Context, val plugins: List<IAztecPlugin> = Ar
             return true
         }
 
-        when (tag.toLowerCase()) {
+        when (tag.lowercase(Locale.getDefault())) {
             LIST_LI -> {
                 val span = createListItemSpan(nestingLevel, alignmentRendering, AztecAttributes(attributes))
                 handleElement(output, opening, span)
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -706,7 +706,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             temp
         }
 
-        val emptyEditTextBackspaceDetector = InputFilter { source, start, end, dest, dstart, dend ->
+        val emptyEditTextBackspaceDetector = InputFilter { source, start, _, _, dstart, dend ->
             if (selectionStart == 0 && selectionEnd == 0
                     && start == 0
                     && dstart == 0 && dend == 0
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -2083,8 +2083,8 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
 
         unknownBlockSpanStart = text.getSpanStart(unknownHtmlSpan)
         blockEditorDialog = builder.create()
-        blockEditorDialog!!.window.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE)
-        blockEditorDialog!!.show()
+        blockEditorDialog?.window?.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE)
+        blockEditorDialog?.show()
     }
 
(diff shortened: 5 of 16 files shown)


oguzkocer (approved):
The changes look good to me.

For full disclosure, there are a few cases where some code can be moved to the `?.let` block, such as [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-82bd9a130476b54a32e3b1528c8c1bd59c3d8e0fbf46dd4585ff152fbf341b71R120-R122) or [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-fcf2b9b853e49ed83f59bdd0f5cc372e71de5ccb788cb0590afa031070bebee1R391-R399), but I don't think they are particularly important - especially for this PR. So, I deliberately avoided a line comment, but then I felt uneasy that I am not mentioning it at all 😓 

---

Related to null-ability changes; I wonder if there are cases where it'd be better to ensure that the value we are dealing is not `null` (by throwing an exception) rather than only doing the assignment if it's not `null`. However, I am not familiar enough with the codebase to tell where that'd make sense without investigating each change.

I suggest waiting for another approval from someone who has more experience with the codebase. It might also be a good idea to ask the editor to be beta tested after this is merged for the extra peace of mind.

ParaskP7:
👋 @oguzkocer !

Thank you so much for the review and testing, much appreciated! 🙇 ❤️ 🥇 

> For full disclosure, there are a few cases where some code can be moved to the ?.let block, such as [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-82bd9a130476b54a32e3b1528c8c1bd59c3d8e0fbf46dd4585ff152fbf341b71R120-R122) or [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-fcf2b9b853e49ed83f59bdd0f5cc372e71de5ccb788cb0590afa031070bebee1R391-R399), but I don't think they are particularly important - especially for this PR. So, I deliberately avoided a line comment, but then I felt uneasy that I am not mentioning it at all 😓

True, I agree, especially on those two cases you mentioned. 💯 

PS: I will go ahead and make the change, those seems easy enough to update anyway, thank you for spotting that! 🙇 

> Related to null-ability changes; I wonder if there are cases where it'd be better to ensure that the value we are dealing is not null (by throwing an exception) rather than only doing the assignment if it's not null. However, I am not familiar enough with the codebase to tell where that'd make sense without investigating each change.

Ha, this is a good point and I was thinking the same while applying the code changes myself, as I am not sure about the implication of suppressing the nullability by ignoring it like that vs. throwing an exception instead. The reason that I went ahead by applying changes that are suppressing the nullability is that in this #982 PR @planarvoid created a while back, he followed the same pattern. As such, on every change that I did, I went back to this PR and made sure that the same change were applied here too.

@planarvoid wdyt? 🤔

> I suggest waiting for another approval from someone who has more experience with the codebase.

I agree, maybe @planarvoid can help with that extra approval too. 😊 

> It might also be a good idea to ask the editor to be beta tested after this is merged for the extra peace of mind.

I am shamelessly pinging @fluiddot here in case you can help with this, or at least so that you are aware what's coming in (also, the following PR should be the `minSdkVersion = 24`).

I'll also add it to my list of things to ask for an extra beta test of the editor on WPAndroid, after this PR is merged and when WPAndroid starts using that new version of Aztec that contain the changes.

That's all what you meant Oguz, right? 🤔

fluiddot:
> > It might also be a good idea to ask the editor to be beta tested after this is merged for the extra peace of mind.
> 
> I am shamelessly pinging @fluiddot here in case you can help with this, or at least so that you are aware what's coming in (also, the following PR should be the `minSdkVersion = 24`).
> 
> I'll also add it to my list of things to ask for an extra beta test of the editor on WPAndroid, after this PR is merged and when WPAndroid starts using that new version of Aztec that contain the changes.

Thanks for the ping @ParaskP7 🙇 ! I can perform a quick check in Gutenberg using this PR as a double-check, although I don't expect any issues based on the changes outlined in the PR's description 👍 .

ParaskP7:
> Thanks for the ping @ParaskP7 🙇 ! I can perform a quick check in Gutenberg using this PR as a double-check.

That would be awesome, thank you so much for the extra hand @fluiddot ! 🙇 x 💯 

> ...although I don't expect any issues based on the changes outlined in the PR's description 👍 

Famous last words... 😅 🤞 🤞 🤞

oguzkocer:
> That's all what you meant Oguz, right? 🤔

Yeap, that's all. Thank you @ParaskP7 🙇‍♂️

ParaskP7:
👋 @oguzkocer !

> For full disclosure, there are a few cases where some code can be moved to the ?.let block, such as [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-82bd9a130476b54a32e3b1528c8c1bd59c3d8e0fbf46dd4585ff152fbf341b71R120-R122) or [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-fcf2b9b853e49ed83f59bdd0f5cc372e71de5ccb788cb0590afa031070bebee1R391-R399), but I don't think they are particularly important - especially for this PR. So, I deliberately avoided a line comment, but then I felt uneasy that I am not mentioning it at all 😓
> 
> True, I agree, especially on those two cases you mentioned. 💯
> 
> PS: I will go ahead and make the change, those seems easy enough to update anyway, thank you for spotting that! 🙇

After looking at how to move some of the code to the `?.let { ... }` block, I am now unsure about your suggestion as I can't think of a good way of moving this code and making it better. 🤔

Can you help me understand your suggestion, for example let's focus on this case [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-fcf2b9b853e49ed83f59bdd0f5cc372e71de5ccb788cb0590afa031070bebee1R391-R399) and see if we can improve it anyhow, thank you for your help! 🙏

fluiddot:
> Thanks for the ping @ParaskP7 🙇 ! I can perform a quick check in Gutenberg using this PR as a double-check, although I don't expect any issues based on the changes outlined in the PR's description 👍 .

I've checked this using https://github.com/wordpress-mobile/AztecEditor-Android/pull/1017 and shared [this comment](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1017#pullrequestreview-1186576213). I still need further time to test this as we'd also need to bump the `minSdkVersion` to `24` in Gutenberg. As I discussed internally with @ParaskP7, I won't have the bandwidth to take a look until Nov 25th, 2022. I'll let you know when I have the results, thanks 🙇 !

oguzkocer:
> Can you help me understand your suggestion, for example let's focus on this case [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1016/files#diff-fcf2b9b853e49ed83f59bdd0f5cc372e71de5ccb788cb0590afa031070bebee1R391-R399) and see if we can improve it anyhow, thank you for your help! 🙏

@ParaskP7 As I mentioned in my original review, I don't think this is particularly important for this PR. I just mentioned it because I feel very uneasy when I see something and don't mention it at all. In any case, below is how I would have "improved" this function, but note that, although I believe this is more Kotlin-y, there is still some personal taste involved.

```
override fun onSaveInstanceState(): Parcelable? =
    super.onSaveInstanceState()?.let { superState ->
        SourceViewEditText.SavedState(superState).apply {
            state = Bundle().apply {
                putBoolean("isSourceVisible", sourceEditor?.visibility == View.VISIBLE)
                putBoolean("isMediaMode", isMediaModeEnabled)
                putBoolean("isExpanded", isExpanded)
                putBoolean("isMediaToolbarVisible", isMediaToolbarVisible)
                putByteArray(RETAINED_EDITOR_HTML_PARSED_SHA256_KEY, editorContentParsedSHA256LastSwitch)
                putByteArray(RETAINED_SOURCE_HTML_PARSED_SHA256_KEY, sourceContentParsedSHA256LastSwitch)
            }
        }
    }
```

I hope you find it useful. Let me know what you think!

_P.S: [Here is the documentation](https://kotlinlang.org/docs/scope-functions.html) I consulted to double check that my suggestion is following Kotlin conventions. Hopefully I interpreted it correctly. 🤞_

ParaskP7:
👋 @fluiddot !

> I've checked this using https://github.com/wordpress-mobile/AztecEditor-Android/pull/1017 and shared https://github.com/wordpress-mobile/AztecEditor-Android/pull/1017#pullrequestreview-1186576213. I still need further time to test this as we'd also need to bump the minSdkVersion to 24 in Gutenberg. As I discussed internally with @ParaskP7, I won't have the bandwidth to take a look until Nov 25th, 2022. I'll let you know when I have the results, thanks 🙇 !

As per my response [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1017#issuecomment-1321684108), thank you so much for all your help, let's pick it up again after your return, this is not urgent work! 🙇 ❤️ 🙏

ParaskP7:
👋 @oguzkocer !

> @ParaskP7 As I mentioned in my original review, I don't think this is particularly important for this PR. I just mentioned it because I feel very uneasy when I see something and don't mention it at all. In any case, below is how I would have "improved" this function, but note that, although I believe this is more Kotlin-y, there is still some personal taste involved.

Ah got it, thank you so much for sharing the code snippet, much appreciated, I now understand what you meant! 🙏 

I was trying to figure out how to refactor by still keeping the number of nesting levels the same, naively thinking that this is what you meant, a quick win. But, I now see your recommendation more clearly. The only disadvantage of doing this refactor is the levels of nesting and the fact that using 3 chained scope functions like that might increase the mental power needed to understand what is going on. Thus, I might have stopped at the below alternative to keep the nesting at level 1, but as you said, with those refactoring, there is always personal taste involved.

```
    override fun onSaveInstanceState(): Parcelable? {
        val superState = super.onSaveInstanceState()
        val savedState = superState?.let { SourceViewEditText.SavedState(it) }
        savedState?.state = Bundle().apply {
            putBoolean("isSourceVisible", sourceEditor?.visibility == View.VISIBLE)
            putBoolean("isMediaMode", isMediaModeEnabled)
            putBoolean("isExpanded", isExpanded)
            putBoolean("isMediaToolbarVisible", isMediaToolbarVisible)
            putByteArray(RETAINED_EDITOR_HTML_PARSED_SHA256_KEY, editorContentParsedSHA256LastSwitch)
            putByteArray(RETAINED_SOURCE_HTML_PARSED_SHA256_KEY, sourceContentParsedSHA256LastSwitch)
        }
        return savedState
    }
```

No matter, thanks for bouncing ideas off that and how these functions can be further improved. I recommend keeping it as is for now, without introducing more changes, just because I think it is better to keep this PR connected to this #982 PR @planarvoid created, along with its changes, and chained with this upstream #1017 PR, without including unnecessary* extra changes in between.

(*) By unnecessary I don't mean that they are not valuable and welcome, I just mean that they won't do much to help with this PR's scope.

planarvoid (approved):
Thanks for the PR @ParaskP7 ! I won't say I'm an expert on Aztec. I think I was trying to be overly cautious in my kotlin update PR. However, these changes are looking good! The only changes would be very nitpicky matter of taste so I'm fine with leaving this as is 👍

ParaskP7:
Thank you so much for reviewing, testing and merging this @planarvoid , great stuff! 🚀 ❤️ 🥇

PR #1021, opened 2022-12-07
Make sure placeholders are not drawn twice when activity is destroyed

planarvoid:
### Fix
When the "Don't keep activities" is turned on and the user leaves the editor and comes back (for example to open a camera and take a photo), the placeholders are redrawn because the `toHtml` is called twice and this triggers the `mediaLoadingStarted` to be called twice. In this PR I'm making sure that doesn't happen.

### Test
- Test this in the linked PR

### Review
@khaykov 


khaykov (approved):
I confirmed the issue and that the fix works!

khaykov:
Feel free to merge whenever 👍

PR #1022, opened 2022-12-08
Fix issue when adding an image on a blank line after a paragraph

planarvoid:
### Fix
The issue fixed here is:
- turn the `addMediaAfterBlocks` on the visual editor
- Make sure you have 3 paragraphs (for example `<p>Line 1</p><p>Line 2</p><p>Line 3</p>`
- Click at the end of "Line 1"
- Add a new line ("enter")
- Add an image (or a placeholder)
- Notice it gets actually added after the "Line 2" instead of "Line 1"

The reason for this was that when the cursor is in a paragraph block, it's trying to add the image at the next line break. However, if it's on a line break, it should use that instead of looking for the next one after the current cursor position.
 
### Test (or test it with the linked PR)
1. turn the `addMediaAfterBlocks` on the visual editor
2. Make sure you have 3 paragraphs (for example `<p>Line 1</p><p>Line 2</p><p>Line 3</p>`
3. Click at the end of "Line 1"
4. Add a new line ("enter")
5. Add an image (or a placeholder)
6. Notice it gets added after "Line 1"

### Review
@khaykov 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/LineBlockFormatter.kt
@@ -247,8 +247,12 @@ class LineBlockFormatter(editor: AztecText) : AztecFormatter(editor) {
                     }
                 }
         if (position <= 0 && selectionEnd != 0) {
-            // If the text contains "\n" return that as the position, else set the position to the end of the text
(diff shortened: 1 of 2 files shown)


khaykov (approved):
Looks good 👍 Thanks for the fix! 

For the record, the original issue only happened to me with "Don't keep activities" flag on.

PR #1024, opened 2023-01-09
Fix predictive text issue with Samsung devices on Android 13

### LINKED ISSUE #1023: Predictive text (grammarly) on Samsung devices with Android 13 replaces the images with OBJ
As you can see from the video, the moment predictive text (grammarly) kicks in, by suggesting comma in "Hello my" the image is replaced with OBJ.

https://user-images.githubusercontent.com/728822/211372645-84231319-8c2b-43c4-bd79-24abe46393d2.mp4

This happens because when adding SuggestionSpan Samsung strips or messes up in other ways existing spans in edit text.

Disable predictive text in keyboard or in Aztec fixes the issue. This is not ideal, obviously.

### Reproduced
1. Add an image to editor.
2. Type something that will activate the Grammarly (in my case it's "Hello my name i"
3. Notice image being replaced with OBJ.

#### Tested
Samsung Galaxy S22 Ultra with Android 13

khaykov:
Fixes #1023 

This PR fixes the predictive text (Grammarly) issue present in default Samsung keyboard on devices with Android 13 (api 33).

Problem:
When Grammarly kick-in and marks issues in the text, content of the editor is replaced with malformed content, where some of the existing spans are removed from editor, and cursor sometimes moves to the end of the text. 

Video from the issue to illustrate it:
https://user-images.githubusercontent.com/728822/211878346-1d296b47-8031-4f18-841a-23fb2a190172.mp4

In addition, we can't debug the Android source code, since it is different on Samsung.

Solution:
I tried multiple approaches using `InputFilters` and `TextWatchers` to detect when the issue is happening, and remedy it. While they do work to some extent, they are not very reliable, and there is one problem I was not able to adress - cursor moving to the end of the text (and scrolling the editor as well).

My last resort was trying to implement a custom `InputConnection`, but I realized that I don't necessarily need to do it - I can make a wrapper around Samsung version, and direct/intercept all the calls to it to see what is up. This worked well, and If found out that Samsung is using Extracted Text (one you usually see in EditText when it expands in landscape mode) for Grammarly, and returning null from it (default behavior of BaseInputConnection) disables the Grammarly, but still keeps regular suggestions and spellchecker. 

Unfortunately, I was not able to make Grammarly work without messing up content of the editor, but in my opinion that's fine.

The fix is disabled by default in a library, but enable in Demo App.

### Test device targeting
1. Launch the app on Samsung device on Android 13
2. Tap in the editor to bring up the keyboard.
3. In Logcat confirm that you see this line `Overriding predictive text behavior on Samsung device with API 33`
4. Try same on other device (not Samsung or different Android version) and confirm that that line is missing from logcat.

### Test the fix
1. Launch the app on Samsung device on Android 13 with default Samsung keyboard (and predictive text enabled, which is default).
2. On a new line, anywhere in demo content type "Hello my name is . "
3. Notice that nothing happens - images are still present, and you are not scrolled to the bottom of the screen.
4. Try typing other things, and confirm that nothing adverse is happening.


AmandaRiu (commented):
This is a brilliant fix @khaykov ! Thank you for taking care of this. We should plan to circle back in a few months to see if it's still an issue since Samsung may fix these issues in a future release. I don't have an Android 13 Samsung device, but I do have one on Android 12. I tested on a Samsung Galaxy S21 FE 5G with the following keyboards:
- Samsung Keyboard
- Gboard
- Grammarly Keyboard

All passed with flying colors 👍🏻

thomashorta (approved):
Changes LGTM, awesome work!

I also tested on my Samsung Galaxy S21+ with Android 13 (One UI 5.0) and it's very buggy on `trunk` version but everything works correctly after this change! 🎉

khaykov:
I added a fix for paragraph spans (like heading) leaking on the next line.
- In empty editor type any heading on the first line.
- Press enter and type something on the new line that will trigger Grammarly (eg. "Hello my name is .")
- Confirm that heading style has not been applied to the second line.

PR #1027, opened 2023-01-30
Limit predictive text override to Samsung keyboard

khaykov:
This PR restricts the fix introduced in #1024 to Samsung keyboards only. Since other keyboards actually work ok, this fix can limit their functionality.

I also changed the API level fix targeting to an open ended `>= 33` since we do not know when or if it's going to be fixed.

### Test (Using Samsung device running Android 13)
1. Make sure default Samsung keyboard is installed on your device.
2. Open the demo app and tap in the editor (this will initialize the Input Connection between keyboard and editor).
3. Check logcat and confirm that you see the `Overriding predictive text behavior on Samsung device with Samsung 
Keyboard with API 33` message.
4. Install any other keyboard (eg. Gboard, Swift Keyboard)
5. Open tap app and tap in the editor.
6. Confirm that the log message from step 3 is not visible.

Voila.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -33,6 +33,7 @@ import android.os.Bundle
 import android.os.Looper
 import android.os.Parcel
 import android.os.Parcelable
+import android.provider.Settings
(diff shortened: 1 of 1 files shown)


nbradbury (approved):
Fix is good! :shipit:

PR #1028, opened 2023-02-03
Placeholder doesn't have correct proportions when drawing the first time

planarvoid:
### Fix
When implementing a placeholder in the Day One repo I've discovered that the width used when building placeholder drawable is incorrect because it doesn't take into account the editor padding and the arbitrary 20 px that's used when drawing placeholders. 

### Review
@[USER_NAME]


danilo04 (approved):
Thanks for the changes. It worked in the clients.

PR #1031, opened 2023-02-17
Implement merging of placeholders

planarvoid:
### Fix
Implement method that allows the user to merge the created placeholder into the previous one. 

### Test
- This can be tested with the related DayOne PR

### Review
@danilo04 


danilo04 (approved):
Hi @planarvoid, tested different scenarios in the client app (Day One Android) and it is working as expected.

PR #1033, opened 2023-02-28
Implement backspace soft override

planarvoid:
### Fix
In this PR I'm adding an option to override the backspace on the soft keyboard. This is useful if you want to show a dialog instead of deleting a certain item (like an image). 

In order to test it you have to:
- Create a placeholder manager instance and register it with Aztec
- Create ImageWithCaptionsAdapter and register it with the placeholder manager
- Change the `EXAMPLE` string to something like this: 
```
        private val EXAMPLE = """
            <p>Line</p>
            <placeholder type="image_with_caption" src="https://file-examples.com/storage/fe1aa0c9d563ea1e4a1fd34/2017/10/file_example_JPG_100kB.jpg" caption="test"
            <p>LIne</p>
        """.trimIndent()
```
- Register the placeholder backspace listener with the aztec like this: 
```
.addBeforeBackSpaceListener(PlaceholderBackspaceListener(visualEditor) {
    true
})
```

### Test 1
1. Do the initial steps
2. Click right behind the placeholder image 
3. Click on "Backspace"
4. Notice the image is not deleted

### Test 2
1. Do the initial steps
2. Click at the beginning of the next line behind the image 
3. Click on "Backspace"
4. Notice the image is not deleted

### Test 3
1. Do the initial steps
2. Click anywhere that's not right behind the image
3. Try backspace
4. Notice the text is deleted as expected

### Review
@khaykov 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/Aztec.kt
@@ -36,6 +36,12 @@ open class Aztec private constructor(
             onMediaDeletedListeners.forEach { it.beforeMediaDeleted(attrs) }
         }
     }
+    private var beforeBackSpaceListeners: MutableList<AztecText.BeforeBackSpaceListener> = mutableListOf()
+    private val beforeBackSpaceListener = object : AztecText.BeforeBackSpaceListener {
+        override fun shouldOverrideBackSpace(position: Int): Boolean {
+            return beforeBackSpaceListeners.any { it.shouldOverrideBackSpace(position) }
+        }
+    }
     private var onVideoInfoRequestedListener: AztecText.OnVideoInfoRequestedListener? = null
     private var onLinkTappedListener: AztecText.OnLinkTappedListener? = null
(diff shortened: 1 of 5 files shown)


khaykov (commented):
I tested with various keyboards, it works well with Gboard and SwiftKey, only one small quirk - if you remove the text around the image, you will be able to remove the placeholder by tapping backspace a couple of times:

https://user-images.githubusercontent.com/728822/222190952-2b6068f8-cab7-42b2-a41c-78a406688448.mp4

It does not work with Samsung and Grammarly keyboards. I can remove the placeholder as normal. And I assume we might see issues with other keyboards. From my experience some keyboards might be using input connection no in the way it was intended.

planarvoid:
Thanks for testing @khaykov ! I think both of these are kind of ok. If the user taps a lot of times, I think it's fine to just let it pass. We'll also show a dialog in the DO app so I think this will not happen IRL. This is a nice-to-have feature so I think it's also OK if it doesn't work on some devices

khaykov:
Got it 👍 Then it's :shipit:

khaykov (approved):

PR #1034, opened 2023-02-28
Feature/implement delete or update

planarvoid:
### Fix
In this PR I'm adding a method that allows users to either remove or update a placeholder based on conditions. This is useful in cases like removing an image from a gallery. If the image is the last item in the gallery, we want to remove it instead of just updating the parameters. If it's not the last item, we'd probably want to remove the image params from the gallery.

### Test
1. This can be tested with the related Day One PR

### Review
@[USER_NAME]


AmandaRiu (approved):
Tested using the day one app [in this PR](https://github.com/bloom/DayOne-Android/pull/2132) and all works well! Checked the code here as well and it looks like the existing logic was moved to a method and tests were added and all look good 👍🏻 :shipit:

Issue #1036, opened 2023-03-13
Update deprecated ExternalStorage function on sample activity


irfano:
[getExternalStoragePublicDirectory()](https://developer.android.com/reference/android/os/Environment#getExternalStoragePublicDirectory(java.lang.String)) function was deprecated in API level 29. Its usage in MainActivity should be updated.

https://github.com/wordpress-mobile/AztecEditor-Android/blob/bfadd9f7456c8f1465c561daa178c83a4e75ec79/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt#L654

ravishanker:
PR #1040

PR #1037, opened 2023-03-13
Update compile sdk to 33

irfano:
### Fix
This updates `compileSdk` to 33 and makes the required changes. 

- Deprecated `getExternalStoragePublicDirectory()` warning was suppressed and opened an [issue](https://github.com/wordpress-mobile/AztecEditor-Android/issues/1036). It's a low-priority task since it's on the sample activity.
- `LayoutParams.SOFT_INPUT_ADJUST_RESIZE` was deprecated in API level 30. I [suppressed](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1037/commits/e56f841865de5cec1d01f021d0617215d8353375) it at first but then found a [way](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1037/commits/0d2a0db8e10249ea60fe15a7288aa2cf0474aa89) to resolve it. 

### Test
Being able to build and basic smoke test should be enough.
Updating `LayoutParams.SOFT_INPUT_ADJUST_RESIZE` may be worth a special test. To test:
1. Run AztecDemo.
2. Scroll down and find the HTML block (It's shown as "?", below "if (value == 5) printf(value)")
3. Tap the HTML block and type a long text to increase the dialog size.
4. Ensure "CANCEL", "SAVE" buttons are still visible and there is a margin between the keyboard and dialog.

### Review
@ParaskP7 

### DIFF
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -642,6 +642,7 @@ open class MainActivity : AppCompatActivity(),
             val intent = Intent(MediaStore.ACTION_IMAGE_CAPTURE)
 
             mediaFile = "wp-" + System.currentTimeMillis() + ".jpg"
+            @Suppress("DEPRECATION")
             mediaPath = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DCIM).toString() +
                     File.separator + "Camera" + File.separator + mediaFile
             intent.putExtra(MediaStore.EXTRA_OUTPUT, FileProvider.getUriForFile(this,
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -511,15 +511,13 @@ open class MainActivity : AppCompatActivity(),
         }
     }
 
-    override fun onRestoreInstanceState(savedInstanceState: Bundle?) {
+    override fun onRestoreInstanceState(savedInstanceState: Bundle) {
         super.onRestoreInstanceState(savedInstanceState)
 
         aztec.initSourceEditorHistory()
 
-        savedInstanceState?.let {
-            if (savedInstanceState.getBoolean("isMediaUploadDialogVisible")) {
-                showMediaUploadDialog()
-            }
+        if (savedInstanceState.getBoolean("isMediaUploadDialogVisible")) {
+            showMediaUploadDialog()
         }
     }
 
--- a/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
+++ b/app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
@@ -27,6 +27,7 @@ import android.view.MotionEvent
 import android.view.View
 import android.widget.PopupMenu
 import android.widget.ToggleButton
+import androidx.activity.OnBackPressedCallback
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.content.res.AppCompatResources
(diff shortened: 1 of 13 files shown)


irfano:
👋🏻 @khaykov!
Can I know your comment about a feature you updated recently? You just commented and uncommented, but maybe you can still help.

`WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE` was deprecated. I want to update it in the recommended way as below.

> This constant was deprecated in API level 30.
> Call [Window#setDecorFitsSystemWindows(boolean)](https://developer.android.com/reference/android/view/Window#setDecorFitsSystemWindows(boolean)) with false and install an [OnApplyWindowInsetsListener](https://developer.android.com/reference/android/view/View.OnApplyWindowInsetsListener) on your root content view that fits insets of type [Type#ime()](https://developer.android.com/reference/android/view/WindowInsets.Type#ime()).

I want to test its effect to be sure it still works as expected, but I couldn't find what this line does exactly. It should resize the block editor position on the screen when the keyboard is shown up, so I removed this line and also removed `windowSoftInputMode` on `AndroidManifest` but couldn't see any behavior difference.

https://github.com/wordpress-mobile/AztecEditor-Android/blob/46a82a36a218cb78bcf1a12e8ed180e0fef8f68b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L2148

Do you know how I can test what this line does?

khaykov:
Hey @irfano ! I remember that one :) I was testing updating `compileSdk` to see if it will fix an issue we had with Samsung devices (it didn't), and that line gave me some compile-time trouble, so I commented it out locally. It made it's way into the commit by accident so I just reverted it back. Sorry, I can't really help you with the `setSoftInputMode`, since I'm not familiar with block editor logic at all :)

irfano:
> Hey @irfano ! I remember that one :) I was testing updating `compileSdk` to see if it will fix an issue we had with Samsung devices (it didn't), and that line gave me some compile-time trouble, so I commented it out locally. It made it's way into the commit by accident so I just reverted it back. Sorry, I can't really help you with the `setSoftInputMode`, since I'm not familiar with block editor logic at all :)

Thank you @khaykov! I found out its function and explained it in the "Test" section.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
```diff
@@ -385,6 +387,20 @@ open class MainActivity : AppCompatActivity(),
...
+            override fun handleOnBackPressed() {
+                mIsKeyboardOpen = false
+                showActionBarIfNeeded()
+
+                // Disable the callback temporarily to allow the system to handle the back pressed event. This usage
+                // breaks predictive back gesture behavior and should be reviewed before enabling the predictive back
+                // gesture feature.
+                isEnabled = false
+                onBackPressedDispatcher.onBackPressed()
+                isEnabled = true
```
ParaskP7, line 400: 1. Praise (❤️): You are becoming a pro at that, aren't you! 😃 
2. Suggestion (💡): Maybe creating a `CompatExtensions.kt` utility class on Aztec, with a `onBackPressedCompat(...)` extension function, copy-pasting it from [here](https://github.com/wordpress-mobile/WordPress-Android/pull/17947/files#diff-b836fc4f03a795e7b9d624220b8ed3d88bbec73d5ecf3848fc55dbee8d5d97a6R16-R30), is the right move for this repo too, just to keep this consistent across every repo, wdyt? 🤔 PS: I know this might seem an overkill as it only applies to this specific and single instance of such `onBackPressedDispatcher` functionality within this repo, so if you want to skip it instead, I would totally understand.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -2099,7 +2098,7 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
 
     @SuppressLint("InflateParams")
     fun showBlockEditorDialog(unknownHtmlSpan: UnknownHtmlSpan, html: String = "") {
-        val builder = AlertDialog.Builder(context)
+        val builder = AlertDialog.Builder(context, R.style.ResizableDialogTheme)
```
ParaskP7, line 2101: Praise (❤️): Thanks for taking the time to resolve this suppression with the most appropriate fix for this specific case, you rock! 🥇

ParaskP7 (approved):
👋 @irfano !

I have reviewed and tested this PR as per the instructions, everything works as expected, good job! 🌟 🌟 🌟

-----

I have left just one suggestion (💡) for you to consider with lots of praise (❤️)! I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

app/src/main/kotlin/org/wordpress/aztec/demo/MainActivity.kt
irfano: > 1. Praise (❤️): You are becoming a pro at that, aren't you! 😃

 and you're becoming a pro at reviewing that. 😁

> 2. Suggestion (💡): Maybe creating a CompatExtensions.kt utility class on Aztec, with a onBackPressedCompat(...) extension function, copy-pasting it from [here](https://github.com/wordpress-mobile/WordPress-Android/pull/17947/files#diff-b836fc4f03a795e7b9d624220b8ed3d88bbec73d5ecf3848fc55dbee8d5d97a6R16-R30), is the right move for this repo too, just to keep this consistent across every repo, wdyt?

Good advice, but this code is already in the demo app, which consists of only one package and four files. Therefore, I will choose to avoid adding a new file.

PR #1044, opened 2023-04-12
Issue/880 center point not in rectangle

### LINKED ISSUE #880: Crash IllegalArgumentException: Center point is not inside any of the rectangles
My app is using Aztec editor and I see many crash reports with such exception.
I was able to reproduce for one case but I believe that it has more cases with the same problem. 

### Reproduced
1. Add some empty block style (quote for example).
2. Long tap.
3. Choose "Select all" from the context menu.

Video with the crash in Aztec demo app: https://drive.google.com/file/d/1BkqoL9gp0e5oEwdO2e_P5SOMqTrmDZ0f/view?usp=sharing

Crash log:
```
    Process: org.wordpress.aztec, PID: 29043
    java.lang.IllegalArgumentException: Center point is not inside any of the rectangles!
        at android.widget.SmartSelectSprite.startAnimation(SmartSelectSprite.java:392)
        at android.widget.SelectionActionModeHelper.startSelectionActionModeWithSmartSelectAnimation(SelectionActionModeHelper.java:325)
        at android.widget.SelectionActionModeHelper.lambda$l1f1_V5lw6noQxI_3u11qF753Iw(Unknown Source:0)
        at android.widget.-$$Lambda$SelectionActionModeHelper$l1f1_V5lw6noQxI_3u11qF753Iw.accept(Unknown Source:4)
        at android.widget.SelectionActionModeHelper$TextClassificationAsyncTask.onPostExecute(SelectionActionModeHelper.java:998)
        at android.widget.SelectionActionModeHelper$TextClassificationAsyncTask.onPostExecute(SelectionActionModeHelper.java:952)
        at android.os.AsyncTask.finish(AsyncTask.java:755)
        at android.os.AsyncTask.access$900(AsyncTask.java:192)
        at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:772)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
2019-12-04 10:46:34.373 29043-29043/org.wordpress.aztec E/WordPress-EDITOR: HTML content of Aztec Editor before the crash:
2019-12-04 10:46:34.377 29043-29043/org.wordpress.aztec E/WordPress-EDITOR: <blockquote></blockquote>
```

nbradbury:
Fixes #880 

We have 37 Sentry users experiencing the "IllegalArgumentException: Center point is not inside any of the rectangles" crash [in Day One](2255-gh-bloom/DayOne-Android), all of which are on Android 13 (Tiramisu). This appears to be a platform issue rather than a problem with Aztec itself.

**Update:** It turns out the crash isn't limited to Android 13.

This PR resolves the problem by detecting and ignoring that crash on Android 13. I'm not well-versed with the Aztec codebase, so if there's a better way to resolve this please let me know.

To test:

- In trunk, clear the example app's editor
- Long press in the empty editor
- Choose "Select all"
- Boom!
- Then pull this branch and observe the problem no longer occurs

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
@@ -65,9 +65,7 @@ class AztecExceptionHandler(private var logHelper: WeakReference<ExceptionHandle
                 visualEditor.get()?.externalLogger?.logException(DynamicLayoutGetBlockIndexOutOfBoundsException("Error #8828", ex))
             }
         // See: https://github.com/wordpress-mobile/AztecEditor-Android/issues/880
-        } else if (ex is java.lang.IllegalArgumentException &&
-                Build.VERSION.SDK_INT == Build.VERSION_CODES.TIRAMISU &&
-                ex.message?.contains("Center point is not inside any of the rectangles") == true) {
+        } else if (ex is java.lang.IllegalArgumentException && ex.message?.contains("Center point is not inside any of the rectangles") == true) {
             visualEditor.get()?.externalLogger?.logException(ex)
             return
         }
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecExceptionHandler.kt
@@ -64,12 +64,7 @@ class AztecExceptionHandler(private var logHelper: WeakReference<ExceptionHandle
             if (detected) {
                 visualEditor.get()?.externalLogger?.logException(DynamicLayoutGetBlockIndexOutOfBoundsException("Error #8828", ex))
             }
-        // See: https://github.com/wordpress-mobile/AztecEditor-Android/issues/880
-        } else if (ex is java.lang.IllegalArgumentException && ex.message?.contains("Center point is not inside any of the rectangles") == true) {
-            visualEditor.get()?.externalLogger?.logException(ex)
-            return
         }
-
         rootHandler?.uncaughtException(thread, ex)
     }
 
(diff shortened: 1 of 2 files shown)


cameocoder:
This issue is not limited to TIRAMISU.

It happens when you "Select All" and the editor just contains empty spans.  Spans that are started but do not actually contain any text/content.

Another way to reproduce the crash is to start the example app with a blank editor (no EXAMPLE text)
1. Start an ordered/unordered list
2. Select All

The reason it is reproducible from the example app when you clear the Example text may be because there are a number of empty spans that are left behind after trying to clear the text.  In particular many AztecVisualLinebreak spans left behind which is a separate issue.

My guess is that the fix for this may be in handling onTextContextMenuItem and Aztec may need to implement its own selectAllText similar to how it custom handles cut/copy/paste.  This may also fix the issue of spans being left behind when you select all of the EXAMPLE text and delete it.

nbradbury:
> This issue is not limited to TIRAMISU.

@cameocoder Thanks for the clarification! I've removed the version check in [690ae78](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1044/commits/690ae78e2592a58909a6aeb0b8dd9d3c70ecf657). I'm not well-versed with this codebase, so I'll wait for someone on the team to chime in about other solutions.

khaykov:
While the fix prevents the crash, if you start typing after selecting empty span, the app will hang :) 
Ideally we do not want to show the context menu when there is only empty span in editor, but I'm not sure we can do it on the library level.

As @cameocoder suggested, we can a add a bit of logic to `onTextContextMenuItem` to deal with it. I think not selecting anything when there is only a block span in editor is a valid option.

Something like this should handle it (have not tested it extensively):
 ```
android.R.id.selectAll -> {
                return if (text.toString() == Constants.END_OF_BUFFER_MARKER.toString()) {
                    return true
                } else {
                    super.onTextContextMenuItem(id)
                }
            }
```

nbradbury:
@khaykov Thanks for the assist! I made the changes you suggested.

planarvoid (approved):
Thanks for the changes 👍 . I've tested this fixes the crash

PR #1046, opened 2023-04-14
Don't report changes for empty text

### LINKED ISSUE #1045: Editor reports it has changes for empty text
The [hasChanges() function](https://github.com/wordpress-mobile/AztecEditor-Android/blob/499459d544af1f7f1ba035449c6f52710bc86250/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L209) always reports it has changes when first opened, even though nothing has been changed.

The problem is that `initialEditorContentParsedSHA256` is initialized as an empty array, which doesn't match `calculateSHA256("")`.

nbradbury:
Fixes #1045 

Currently [hasChanges](https://github.com/wordpress-mobile/AztecEditor-Android/blob/499459d544af1f7f1ba035449c6f52710bc86250/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt#L209) returns `CHANGES` when it is first opened, before any changes have been made. This happens because the comparison relies on `initialEditorContentParsedSHA256`, which is initialized as an empty array - which doesn't match the hash of an empty string.

The PR resolves that problem.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -207,6 +207,9 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         }
 
         fun hasChanges(initialEditorContentParsedSHA256: ByteArray, newContent: String): EditorHasChanges {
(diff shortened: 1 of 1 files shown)


nbradbury:
We decided to close this because although the fix is good for Day One, we're not 100% sure it's the behavior WordPress wants.

PR #1047, opened 2023-04-17
Animate size changes of placeholders

planarvoid:
### Fix
In this PR I'm implementing improvements to the reloading of placeholders. They will now reuse the same view if not changed and animate size changes. This should improve behaviour when working with galleries.

### Test
1. Test this with the DO PR

### Review
@[USER_NAME]


danilo04 (approved):
This PR was reviewed as part of https://github.com/bloom/DayOne-Android/pull/2242. Everything is working as expected.

PR #1049, opened 2023-06-12
Fixed autosuggestion issues (non Samsung)

khaykov:
We got a report that autosuggestion is acting weird on **non-Samsung** devices. This is pretty weird, so investigated and found that the issue was introduced with #1033.

Luckily, the fix is pretty simple - we want to use existing `InputConnectionWrapper`, as a base class for `DeleteOverrideInputConnection` and not our own implementation. 

There are two things we need to test. First is the Autocorrect - you want to use non-Samsung Android 13 with Gboard (emulator works as well).

- Clear the editor, and type "I slept wless good."
- Notice that `wless` is underlined.
- Places a cursor between `wle` and `ss`.
- Tap on the `less` in a gboard suggestion bar.
- Confirm that autocorrection is correctly committed. Before the fix, correction will be inserted before `ss`, which will result in `lessss`.

Next, we need to make sure the changes in #1033 are working as expected. This only applies to soft keyboard. You will also need to modify a bit of Main Activity to do this:
- Change `EXAMPLE` to this:
```
private val EXAMPLE = """
            <p>Line</p>
            <placeholder type="image_with_caption" src="https://examplebloge.files.wordpress.com/2017/02/3def4804-d9b5-11e6-88e6-d7d8864392e0.png" caption="test"
            <p>LIne</p>
        """.trimIndent()
```
- Before the aztec is created with `Aztec.with..`, create placeholder managed and adapter for images with captions:

```
 val  placeholderManager = PlaceholderManager(
                visualEditor,
                findViewById(R.id.container_frame_layout),
                htmlTag = "placeholder"
        )

        val imageWithCaptionAdapter = ImageWithCaptionAdapter()

        placeholderManager.registerAdapter(imageWithCaptionAdapter)
```
- Register adapter and add a backspace listener to Aztec (I just copy/pasted whole thing):
```
aztec = Aztec.with(visualEditor, sourceEditor, toolbar, this)
                .setImageGetter(GlideImageLoader(this))
                .setVideoThumbnailGetter(GlideVideoThumbnailLoader(this))
                .setOnImeBackListener(this)
                .setOnTouchListener(this)
                .setHistoryListener(this)
                .setOnImageTappedListener(this)
                .setOnVideoTappedListener(this)
                .setOnAudioTappedListener(this)
                .addOnMediaDeletedListener(this)
                .setOnVideoInfoRequestedListener(this)
                .addBeforeBackSpaceListener(PlaceholderBackspaceListener(visualEditor) {
                   true
                })
                .addPlugin(WordPressCommentsPlugin(visualEditor))
                .addPlugin(MoreToolbarButton(visualEditor))
                .addPlugin(PageToolbarButton(visualEditor))
                .addPlugin(CaptionShortcodePlugin(visualEditor))
                .addPlugin(VideoShortcodePlugin())
                .addPlugin(AudioShortcodePlugin())
                .addPlugin(HiddenGutenbergPlugin(visualEditor))
                .addPlugin(galleryButton)
                .addPlugin(cameraButton)
                .addPlugin(placeholderManager)
```
- Launch the app.
- Click right behind the placeholder image
- Click on "Backspace"
- Notice the image is not deleted
- Click at the beginning of the next line behind the image
- Click on "Backspace"
- Notice the image is not deleted
- Click anywhere that's not right behind the image
- Try backspace
- Notice the text is deleted as expected


danilo04 (approved):
Tested on Day One and it works as expected.

PR #1055, opened 2023-08-03
Do not remove empty line after a paragraph

planarvoid:
### Fix
Changes necessary to stop removing empty line between paragraphs.

### Test
1. Use the linked PR to test it
### Review
@[USER_NAME]


nbradbury:
:shipit:

nbradbury (approved):
Fix is good!

PR #1056, opened 2023-08-04
Fix crash when history cache file is missing

khaykov:
This PR fixes a crash that can happen when Aztec state with history is restored, but the history cache file was removed by OS.

This crash is pretty rare, but can happen on devices with low storage, when you reopen editor that was destroyed by OS (usually happens if you don't open it for some time).

### Test
1. For easier access to file system consider using an emulator.
2. Enabled "Do not keep activities".
3. Open editor, and make a couple of changes.
4. Confirm that the changes are added to history.
5. Minimize the app.
6. From AS device manager, open file browser of the emulator.
7. Navigate to `data/data/org.wordpress.aztec/cache`
8. Delete files that start with `HISTORY_LIST_KEY`.
9. Go back to the editor from app drawer.
10. Confirm that the app is not crashing, and there is no history.

### Review
@[USER_NAME]


nbradbury (approved):
Looks good! :shipit:

PR #1057, opened 2023-08-10
Fix highlight function in the editor

planarvoid:
### Fix
Highlight action was only partially implemented in the Aztec editor. This PR fixes that. The color for highlight can come from 2 places. It's either set in parameters or it can be passed as part of the css styles (this takes precedence over android styles). 

### Test
Test this with the linked PR

### Review
@[USER_NAME]

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/spans/HighlightSpan.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/spans/HighlightSpan.kt
@@ -13,8 +13,6 @@ class HighlightSpan(
         override var attributes: AztecAttributes = AztecAttributes(),
         val colorHex: Int
 ) : BackgroundColorSpan(colorHex), IAztecInlineSpan {
-    var alpha: Int = 220
-
     override var TAG = HIGHLIGHT_TAG
 
     companion object {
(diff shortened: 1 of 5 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/spans/HighlightSpan.kt
```diff
@@ -6,16 +6,42 @@ import androidx.core.content.ContextCompat
...
+import org.wordpress.aztec.util.ColorConverter
 
 class HighlightSpan(
         override var attributes: AztecAttributes = AztecAttributes(),
-        highlightStyle: InlineFormatter.HighlightStyle = InlineFormatter.HighlightStyle(R.color.grey_lighten_10),
-        context: Context
-) : BackgroundColorSpan(ContextCompat.getColor(context, highlightStyle.color)), IAztecInlineSpan {
+        val colorHex: Int
+) : BackgroundColorSpan(colorHex), IAztecInlineSpan {
+    var alpha: Int = 220
```
khaykov: I believe this alpha value is not used.

khaykov (approved):
Looks good! I was confused about how test it, but with client I was able to confirm that it works.

There is one unused attribute, but feel free to merge when it's removed.

PR #1060, opened 2023-09-15
Fix issue when removing media in the middle of a paragraph

planarvoid:
### Fix
This PR fixes a crash that's happening when a media item is removed in the middle of a paragraph. The fix is to ignore the spans that are wrapping the media, only to handle the ones starting right after the media end.

### Test
1. Run the unit tests without the fix
2. Notice the `PARAGRAPH span must start at paragraph boundary` exception
3. Run the unit tests with the fix

### Review
@danilo04 


danilo04 (approved):
Hi @planarvoid, thank you for the fix. I tested different escenarios around galleries and other placeholders and everything is working as expected.

PR #1061, opened 2023-09-27
Add text size modifier

khaykov:
This PR adds a text size modifier, that is applied to base font size, and scales all the text accordingly. This is mainly aimed at use cases where you have specific heading sizes defined, instead of relying on default scaling behavior.

If you don't have a manually set heading sizes, all the text will scale with base font size.

Modifier can be negative, for downscaling.

To test:
- Try setting the modifier using `setTextSizeModifier` (eg `visualEditor.setTextSizeModifier(25)`) and confirm that the text scales as expected.
- Try settings custom heading font sized using `headingOneFontSize`, `headingTwoFontSize`, etc. Confirm that heading size is scaling properly.


nbradbury (approved):
All good, thanks for this change! :shipit:

Issue #848, opened 2019-08-28
java.lang.IndexOutOfBoundsException: setSpan()


yangshibai:
When set  aztec.visualEditor.filters = arrayOf<InputFilter>(InputFilter.LengthFilter(1000)) ,it menas EditText MaxLength is 1000，next I paste piece of text with more than 500 words, the
first time is all right ,but the second time I continue paste this text ,the application crashed。

     This is the crash Log:

 java.lang.IndexOutOfBoundsException: setSpan (1272 ... 1272) ends beyond length 1000
        at android.text.SpannableStringBuilder.checkRange(SpannableStringBuilder.java:1309)
        at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:680)
        at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:672)
        at android.text.Selection.setSelection(Selection.java:76)
        at android.text.Selection.setSelection(Selection.java:87)
        at android.widget.EditText.setSelection(EditText.java:127)
        at org.wordpress.aztec.AztecText.fromHtml(AztecText.kt:1158)
        at org.wordpress.aztec.AztecText.paste(AztecText.kt:1598)
        at org.wordpress.aztec.AztecText.paste$default(AztecText.kt:1557)
        at org.wordpress.aztec.AztecText.onTextContextMenuItem(AztecText.kt:1493)
        at android.widget.Editor$ActionPopupWindow.onClick(Editor.java:4327)
        at android.view.View.performClick(View.java:6266)
        at android.widget.TouchPanelLayout.resetChildState(TouchPanelLayout.java:58)
        at android.widget.TouchPanelLayout.dispatchTouchEvent(TouchPanelLayout.java:46)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:2963)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2650)
        at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:2963)
        at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2650)
        at android.widget.PopupWindow$PopupDecorView.dispatchTouchEvent(PopupWindow.java:2397)
        at android.view.View.dispatchPointerEvent(View.java:11975)
        at android.view.ViewRootImpl$ViewPostImeInputStage.processPointerEvent(ViewRootImpl.java:4836)
[...]

PR #1062, opened 2023-10-10
Make setSelection safe

planarvoid:
### Fix
We're having occasional issues when the `setSelection` is called out of bounds of the text. While it would be better to fix the underlying issues, I think this is a safe approach to handling this. What usually crashes the app is something like this: 

- Preserve cursor position N at the end of text
- Set the cursor position to 0
- Do some modification to the text that shorten the text to N - 1
- Try to set the cursor position to N
- Crash because text length is N-1 < N

This solution handles that case and if this happens, sets the selection to N-1 instead. I can't think of a case where this might be an issue but let me know if you think it's unsafe.

### Review
@danilo04 


Bibazavr:
possible related to https://github.com/wordpress-mobile/AztecEditor-Android/issues/848

Bibazavr:
I've checked this solution and this fixes crash from https://github.com/wordpress-mobile/AztecEditor-Android/issues/848

PR #1063, opened 2023-10-11
Disabled autocorrect on Samsung device with Samsung keyboard on API 33+

### LINKED ISSUE #1026: After space is deleted when typing new characters on Samsung android 13 words are added.
After the update https://github.com/wordpress-mobile/AztecEditor-Android/pull/1024 new bug is introduced. If the user misspells some words for example the word "hav", and then adds the space character. After deleting the space character and adding for example character "e" to fix misspelling, the whole word "have" is added to the end of the existing word, so "hav" becomes "havhave".

### Expected
Expected behaviour in this case would be for just one character to be added instead of the whole word. In this case if character "e" is added to "hav" it should show "have" instead of "havhave".
Attached is the video of the bug.

### Observed
https://user-images.githubusercontent.com/123743197/215427010-a289a559-1fa1-467c-8d2f-56c142080001.mp4

### Reproduced
1. Open demo app
2. Type "hav" in Heading 1 section
3. Type space
4. Delete space
5. Type "e"

#### Tested
Samsung S21 on 13

khaykov:
Fixes #1026

Autocorrect on Samsung have been broken for some time, which resulted in occasional text duplication.

This PR disables auto-correction on devices with Samsung with Android 13 that are using Samsung keyboard. We will still have suggestions and ability to manually commit them. I believe that this experience is better then what we have now.

### Test
1. Open the app using Samsung device with android 13 and Samsung keyboard.
2. Type some text into the app.
3. Confirm that nothing looks off - focus stays where it should be, image is present, styles are not being removed.
4. Confirm that autocorrect is not working - type something like "I hav " (with the space at the end) and notice that `hav` is not changed into `have`.
5. Switch to Gboard (or other keyboard).
6. Confirm that autocorrect is working - type something like "I hav " (with the space at the end) and notice that `hav` IS changed into `have`.
7. Change back to samsung keyboard and confirm that the autocorrect is still disabled.


nbradbury (approved):
:shipit:

PR #1064, opened 2023-10-23
Move loading from HTML to a background thread

planarvoid:
### Fix
This PR creates a version of the `fromHtml` method which runs on the background thread. This makes the entry loading much faster.

### Test
1. Test with the DO PR

### Review
@danilo04 


danilo04 (approved):
This is working great @planarvoid. Thank you for the changes!

PR #1068, opened 2023-11-13
Add back SamsungInputConnection

danilo04:
### Fix

This PR adds back the `SamsungInputConnection` which helps avoid errors such as the editor transforming photos to `obj` items in the text. 

### Test

**Note:** use a Samsung device with Android 13.

1. Open the app using Samsung device with android 13 and Samsung keyboard.
2. Type some text into the app.
3. Confirm that nothing looks off - focus stays where it should be, image is present, styles are not being removed.
4. Confirm that autocorrect is not working - type something like "I hav " (with the space at the end) and notice that hav is not changed into have.
5. Switch to Gboard (or other keyboard).
6. Confirm that autocorrect is working - type something like "I hav " (with the space at the end) and notice that hav IS changed into have.
7. Change back to samsung keyboard and confirm that the autocorrect is still disabled.

### Review
@planarvoid @khaykov 


khaykov (approved):
Thanks, @danilo04 !

I see we keep `outAttrs.inputType = outAttrs.inputType or InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS`, which is a good idea. Hopefully both this and `SamsungInputConnection` will prevent issue from happening.

PR #1071, opened 2023-12-04
Address `NullPointerException` crash in `AztecHeadingSpan`

SiobhyB:
A potential fix for the crash described at https://github.com/wordpress-mobile/WordPress-Android/issues/18657.

### Related PRs

* `Gutenberg`: https://github.com/WordPress/gutenberg/pull/56757
* `Gutenberg Mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/6435
* `Android`: https://github.com/wordpress-mobile/WordPress-Android/pull/19724

### Description

Sentry has been reporting the following crash for both the WordPress and Jetpack Android apps:

```
NullPointerException
org.wordpress.aztec.spans.AztecHeadingSpan in chooseHeight
```

The stack trace for each individual event specifically references one of the following lines (the specific line varies for each event), all of which are within the `chooseHeight` function and include a reference to `previousFontMetrics`:

* [Line 131](https://github.com/wordpress-mobile/AztecEditor-Android/blob/d93ce9f3a2a6cafdb0c05a630e2bdff5275d6f9f/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt#L131)
* [Line 132](https://github.com/wordpress-mobile/AztecEditor-Android/blob/d93ce9f3a2a6cafdb0c05a630e2bdff5275d6f9f/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt#L132)
* [Line 155](https://github.com/wordpress-mobile/AztecEditor-Android/blob/d93ce9f3a2a6cafdb0c05a630e2bdff5275d6f9f/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt#L155)
* [Line 159](https://github.com/wordpress-mobile/AztecEditor-Android/blob/d93ce9f3a2a6cafdb0c05a630e2bdff5275d6f9f/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt#L159)
* [Line 160](https://github.com/wordpress-mobile/AztecEditor-Android/blob/d93ce9f3a2a6cafdb0c05a630e2bdff5275d6f9f/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt#L160)

As described at https://github.com/wordpress-mobile/WordPress-Android/issues/18657#issuecomment-1835995579, it has proven difficult to reproduce this crash. It's suspected that there are specific instances with lower performing devices or devices on poor connections that can lead to `previousFontMetrics` being set as null elsewhere in the code while `chooseHeight` is running.

Given the difficulty reproducing, it hasn't been possible to get to the heart of this issue. Instead, with this PR, the previous non-null assertions previously used with `previousFontMetrics` have been replaced with safe calls. With this approach, the functionality should remain the same, while guarding against `NullPointerException` crashes.

#### Verify Aztec demo continues to function as expected

1. Load the Aztec demo with this PR's changes applied.
2. Verify that you can add headings with no unexpected side effects.

#### Verify Gutenberg continues to function as expected

1. Using the installable build available from https://github.com/wordpress-mobile/WordPress-Android/pull/19724, install the app and navigate to the post editor.
2. Add some heading blocks and ensure there are no unexpected side effects when performing tasks. Example tasks to test include splitting the block, applying new font sizes and line heights, etc.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt
@@ -127,10 +127,10 @@ open class AztecHeadingSpan(
         // save original font metrics
         if (previousFontMetrics == null) {
             previousFontMetrics = Paint.FontMetricsInt()
-            previousFontMetrics!!.top = fm.top
-            previousFontMetrics!!.ascent = fm.ascent
-            previousFontMetrics!!.bottom = fm.bottom
-            previousFontMetrics!!.descent = fm.descent
+            previousFontMetrics?.top = fm.top
+            previousFontMetrics?.ascent = fm.ascent
+            previousFontMetrics?.bottom = fm.bottom
+            previousFontMetrics?.descent = fm.descent
         }
 
         var addedTopPadding = false
@@ -151,13 +151,17 @@ open class AztecHeadingSpan(
 
         // apply original font metrics to lines that should not have vertical padding
         if (!addedTopPadding) {
-            fm.ascent = previousFontMetrics!!.ascent
-            fm.top = previousFontMetrics!!.top
+            previousFontMetrics?.let {
+                fm.ascent = it.ascent
+                fm.top = it.top
+            }
         }
 
         if (!addedBottomPadding) {
-            fm.descent = previousFontMetrics!!.descent
-            fm.bottom = previousFontMetrics!!.bottom
+            previousFontMetrics?.let {
+                fm.descent = it.descent
+                fm.bottom = it.bottom
+            }
         }
     }
 


SiobhyB:
@khaykov, 👋 , I wanted to give you a heads up on this PR as I know Day One uses Aztec for the Android app. Can you take a look to let me know if you spot anything that might cause issues for you? We're hoping to merge this fix in time for a beta in the next day or so.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt
```diff
@@ -151,13 +151,17 @@ open class AztecHeadingSpan(
...
-            fm.ascent = previousFontMetrics!!.ascent
-            fm.top = previousFontMetrics!!.top
+            previousFontMetrics?.let {
+                fm.ascent = it.ascent
+                fm.top = it.top
+            }
         }
 
         if (!addedBottomPadding) {
-            fm.descent = previousFontMetrics!!.descent
```
fluiddot, line 159: Based on the stack trace, it's not clear to me whether the exception is caused when accessing `fm` or `previousFontMetrics`. We are guarding the access to `previousFontMetrics` but, should we also consider `fm`?

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt
```diff
@@ -127,10 +127,10 @@ open class AztecHeadingSpan(
...
         if (previousFontMetrics == null) {
             previousFontMetrics = Paint.FontMetricsInt()
-            previousFontMetrics!!.top = fm.top
-            previousFontMetrics!!.ascent = fm.ascent
-            previousFontMetrics!!.bottom = fm.bottom
-            previousFontMetrics!!.descent = fm.descent
+            previousFontMetrics?.top = fm.top
+            previousFontMetrics?.ascent = fm.ascent
+            previousFontMetrics?.bottom = fm.bottom
+            previousFontMetrics?.descent = fm.descent
```
fluiddot, line 133: I presume that in this case `previousFontMetrics` should always be defined, as it's initialized in line 129 🤔. If there are crashes pointing to these lines, I wonder if the undefined variable would be `fm`. WDYT?

**NOTE:** This is a similar comment to https://github.com/wordpress-mobile/AztecEditor-Android/pull/1071/files#r1414225141.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt
SiobhyB: @fluiddot, the compiler returns `Unnecessary safe call on a non-null receiver of type Paint.FontMetricsInt` when attempting to guard access to `fm`, so I believe `Paint.FontMetricsInt` (its value) is non-nullable. 

From what I've read, I understand that the Kotlin compiler can generally be trusted to accurately detect non-nullable methods, so feel it'd be okay to stick with guarding `previousFontMetrics` only and revisiting if the crash continues. Wdyt?

fluiddot: Makes sense, thanks for elaborating 🙇.

> From what I've read, I understand that the Kotlin compiler can generally be trusted to accurately detect non-nullable methods, so feel it'd be okay to stick with guarding previousFontMetrics only and revisiting if the crash continues. Wdyt?

Sure, let's stick with this approach.

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt
fluiddot: Answered [here](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1071#discussion_r1414242446).

aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt
SiobhyB: I share your hesitation, though. If the crash continues after these changes, indicating `fm` is the culprit, we might have to dig deeper into potential race conditions that lead to that. 🤔

fluiddot: After checking the code, I presume that the crash might be related to the following code, as it sets `previousFontMetrics` to `null` in some cases:

https://github.com/wordpress-mobile/AztecEditor-Android/blob/6dd6bc5768a6e8eb60945171a1ee65346bf921b4/aztec/src/main/kotlin/org/wordpress/aztec/spans/AztecHeadingSpan.kt#L191-L193

As you pointed out in the PR's description, there might be a case (possibly related to low-end devices and slow performance) where the variable is set to `null` when this code is about to be executed 🤔 .

SiobhyB: I agree, that line was tweaked in https://github.com/wordpress-mobile/AztecEditor-Android/pull/952, which was included in 19.2 (the version this crash began happening in). That leads me to think that `previousFontMetrics` being `null` is indeed the culprit.

fluiddot (approved):
LGTM 🎊 !

Regarding the changes, I forced setting `previousFontMetrics` to `null` in the Aztec demo, and seems there are no visible side effects.

I've tested both the Aztec demo app and Gutenberg, the Headings worked as expected 🎊. The only thing I noticed is some unexpected behaviors in the Aztec demo, although they aren't regressions as they can be reproduced in `trunk`. I can follow-up by opening a GitHub issue.


https://github.com/wordpress-mobile/AztecEditor-Android/assets/14905380/63fb5dc6-ce39-484d-937b-e0f86388d69e

SiobhyB:
@khaykov, I'm going to go ahead to merge this PR, so we can get it ready for an upcoming beta. Please let me know if there's anything that stands out to you as off, though! I think the changes are fairly safe, but thought it was best to make you aware of them regardless.

PR #1073, opened 2024-01-19
Improve Mark formatting

geriux:
**Related PRs:**
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/6517
- https://github.com/WordPress/gutenberg/pull/57650
- https://github.com/wordpress-mobile/WordPress-Android/pull/19988

This PR adds missing implementation for the Mark formatting style.

- Adds a new `MarkPlugin` that will handle adding the needed CSS attributes.
- Updates the `MarkSpan` to receive a color param and to set it as the current color.
- Removes `applyMarkInlineStyle` in favor of `applyAfterMarkInlineStyle` which will handle if the color changes while editing a Mark span.

### Test
Please follow the testing instructions in [this PR's description](https://github.com/WordPress/gutenberg/pull/57650).

This shouldn't affect other Aztec functionalities.

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
@@ -109,6 +109,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
                     }
                     AztecTextFormat.FORMAT_MARK -> {
                         applyInlineStyle(item, textChangedEvent.inputStart, textChangedEvent.inputEnd)
+                        applyAfterMarkInlineStyle(textChangedEvent.inputStart, textChangedEvent.inputEnd)
                     }
                     else -> {
                         // do nothing
@@ -246,27 +247,32 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
         }
     }
 
-    public fun updateMarkStyle(start: Int = selectionStart, end: Int = selectionEnd) {
+    private fun applyAfterMarkInlineStyle(start: Int = selectionStart, end: Int = selectionEnd) {
+        // If there's no new mark style color to update, it skips applying the style updates.
+        if (markStyleColor == null) {
+            return
+        }
+
         val spans = editableText.getSpans(start, end, MarkSpan::class.java)
         spans.forEach { span ->
             if (span != null) {
-                val currentSpanStart = editableText.getSpanStart(span)
-                val currentSpanEnd = editableText.getSpanEnd(span)
                 val color = span.getTextColor()
+                val currentSpanStart = editableText.getSpanStart(span)
 
                 editableText.removeSpan(span)
                 editableText.setSpan(
                     MarkSpan(AztecAttributes(), color),
                     currentSpanStart,
-                    currentSpanEnd,
+                    start,
                     Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
                 )
                 editableText.setSpan(
                     MarkSpan(AztecAttributes(), markStyleColor),
                     start,
                     end,
-                    Spanned.SPAN_INCLUSIVE_INCLUSIVE
+                    Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
                 )
+                markStyleColor = null;
             }
         }
     }
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
@@ -127,8 +127,8 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
 
         // Clear Mark formatting styles
         if (!editor.selectedStyles.contains(AztecTextFormat.FORMAT_MARK) && start >= 1 && end > 1 ) {
-            val previousMarkSpan = editableText.getSpans(start - 1, start, MarkSpan::class.java);
-            val markSpan = editableText.getSpans(start, end, MarkSpan::class.java);
+            val previousMarkSpan = editableText.getSpans(start - 1, start, MarkSpan::class.java)
+            val markSpan = editableText.getSpans(start, end, MarkSpan::class.java)
             if (markSpan.isNotEmpty() || (previousMarkSpan.isNotEmpty() && markStyleColor == null)) {
                 removeInlineCssStyle(start, end)
                 return
--- a/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
@@ -40,6 +40,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
 
     data class CodeStyle(val codeBackground: Int, val codeBackgroundAlpha: Float, val codeColor: Int)
     data class HighlightStyle(@ColorRes val color: Int)
+    var markStyleColor: String? = null
 
     fun toggle(textFormat: ITextFormat) {
         if (!containsInlineStyle(textFormat)) {
@@ -107,12 +108,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
                         applyInlineStyle(item, textChangedEvent.inputStart, textChangedEvent.inputEnd)
                     }
                     AztecTextFormat.FORMAT_MARK -> {
-                        // For cases of an empty mark tag, either at the beginning of the text or in between
-                        if (textChangedEvent.inputStart == 0 && textChangedEvent.inputEnd == 1) {
-                            applyMarkInlineStyle(textChangedEvent.inputStart, textChangedEvent.inputEnd)
-                        } else {
-                            applyInlineStyle(item, textChangedEvent.inputStart, textChangedEvent.inputEnd)
-                        }
+                        applyInlineStyle(item, textChangedEvent.inputStart, textChangedEvent.inputEnd)
                     }
                     else -> {
                         // do nothing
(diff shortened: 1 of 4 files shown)


aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -250,10 +247,33 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
...
+                val color = span.getTextColor()
+                val currentSpanStart = editableText.getSpanStart(span)
+
+                editableText.removeSpan(span)
+                editableText.setSpan(
+                    MarkSpan(AztecAttributes(), color),
+                    currentSpanStart,
+                    start,
+                    Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
+                )
```
geriux: When there's an existing Mark style but the color changes:

This will keep the previous Mark style, from where it starts `currentSpanStart` until the new Mark style starts `start`.

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -250,10 +247,33 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
...
+                editableText.setSpan(
+                    MarkSpan(AztecAttributes(), markStyleColor),
+                    start,
+                    end,
+                    Spanned.SPAN_EXCLUSIVE_EXCLUSIVE
+                )
```
geriux: And this will add the new Mark format style with the new color from the current `start`/`end`

geriux:
Hey there @antonis 👋 

I was wondering if you would be able to help me out with a PR review 😄, I've updated the Highlight color implementation so it fixes some issues for this feature since we first partially implemented it in https://github.com/wordpress-mobile/AztecEditor-Android/pull/943

A lot of testing has been already done in https://github.com/WordPress/gutenberg/pull/57650 so I was wondering if you'd have some time to check the native code for the Aztec side. Let me know if you could take a look! Thank you!

antonis:
> I was wondering if you'd have some time to check the native code for the Aztec side. Let me know if you could take a look! Thank you!

I'll be glad to help with this @geriux :) I plan to get to this later today or tomorrow 🙇

geriux:
> I'll be glad to help with this @geriux :) I plan to get to this later today or tomorrow 🙇

Thank you so much!

aztec/src/main/kotlin/org/wordpress/aztec/formatting/InlineFormatter.kt
```diff
@@ -40,6 +40,7 @@ class InlineFormatter(editor: AztecText, val codeStyle: CodeStyle, private val h
 
     data class CodeStyle(val codeBackground: Int, val codeBackgroundAlpha: Float, val codeColor: Int)
     data class HighlightStyle(@ColorRes val color: Int)
+    var markStyleColor: String? = null
```
antonis: code styling nit pick: We could move this declaration below the `backgroundSpanColor` above

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
```diff
@@ -1,12 +1,47 @@
...
+        configureTextPaint(tp)
+    }
+
+    private fun configureTextPaint(tp: TextPaint) {
+        if (textColor != null) {
+            tp.color = textColor as Int
+        }
+    }
+
+    fun getTextColor(): String {
```
antonis, line 41: If I understand this correctly we are only accessing the `String` value of the `textColor` property via this method from outside the `MarkSpan` class. If this is the case we could make the `textColor` private and rename it to avoid confusion (the method name matches the [generated java getter name](https://kotlinlang.org/docs/java-to-kotlin-interop.html#properties) but returns a different type)

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
```diff
@@ -1,12 +1,47 @@
...
+    }
+
+    override fun updateDrawState(tp: TextPaint) {
+        configureTextPaint(tp)
+    }
+
+    private fun configureTextPaint(tp: TextPaint) {
+        if (textColor != null) {
+            tp.color = textColor as Int
+        }
```
antonis: We could simplify this in Kotlin like:
```suggestion
        textColor?.let { tp.color = it }
```
and probably use the code directly in `updateDrawState` removing this private method.

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
```diff
@@ -1,12 +1,47 @@
...
 import org.wordpress.aztec.AztecAttributes
+import org.wordpress.aztec.source.CssStyleFormatter
 
-class MarkSpan(override var attributes: AztecAttributes = AztecAttributes()) : CharacterStyle(), IAztecInlineSpan {
+class MarkSpan : CharacterStyle, IAztecInlineSpan {
     override var TAG = "mark"
 
-    override fun updateDrawState(tp: TextPaint?) {
+    override var attributes: AztecAttributes = AztecAttributes()
+    var textColor: Int? = null
```
antonis: Since `textColor` is only set in the constructor we could use an immutable property. If we also rename it as suggested below this could look like:

```
private val textColorValue: Int?
```

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
```diff
@@ -1,12 +1,47 @@
...
+            textColor = Color.parseColor(color)
+        }
+    }
+
+    constructor(attributes: AztecAttributes = AztecAttributes(), colorString: String?) : super() {
+        this.attributes = attributes
+
+        if (colorString != null) {
+            textColor = Color.parseColor(colorString)
+        }
```
antonis: If we use an immutable property the this should change to
```
        textColorValue = if (colorString != null) {
            Color.parseColor(colorString)
        } else {
            null
        }
```

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
```diff
@@ -1,12 +1,47 @@
...
-    override fun updateDrawState(tp: TextPaint?) {
+    override var attributes: AztecAttributes = AztecAttributes()
+    var textColor: Int? = null
+
+    constructor(attributes: AztecAttributes = AztecAttributes()) : super() {
+        this.attributes = attributes
+
+        val color = CssStyleFormatter.getStyleAttribute(attributes,
+            CssStyleFormatter.CSS_COLOR_ATTRIBUTE)
+        if (color.isNotEmpty()) {
```
antonis: Following on the above
```
        textColorValue = if (color.isNotEmpty()) {
            Color.parseColor(color)
        } else {
            null
        }
```

antonis (approved):
Awesome work @geriux ⭐ 
The code overall looks good and works as expected. I added some suggestions which are not blocking.

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
geriux: Thank you for the suggestion Antonis! 

Once I make this change with the [following suggestion](https://github.com/wordpress-mobile/AztecEditor-Android/pull/1073/files#r1532075080) I'm getting:

`Val cannot be reassigned`

I'm not sure if I'm missing something 🤔

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
geriux: Ohh I see so for this case, it'd make sense to rename this public function to something like `getTextHexColor` or something like that since the returned value is different from the private val `textColorValue` right?

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
antonis: Hey @geriux 👋 
I think the error occurs since this change depends on the other suggestions. I prepared a [patch](https://github.com/wordpress-mobile/AztecEditor-Android/files/14693797/markspan.patch) with all the suggestions. Let me know if that helps.
Sorry for confusing things with my multiple inline comments.

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
antonis: Yep. Or we can keep the method as is to avoid changes in other places and rename the `textColor` to `textColorValue`.

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
geriux: Ohhhh since I was doing a change at a time I was encountering that error but it was highlighting it where the variable was being declared 🙃 

Thank you for sharing the patch!

aztec/src/main/kotlin/org/wordpress/aztec/spans/MarkSpan.kt
geriux: TIL about this approach! I'll apply this suggestion 🚀

geriux:
Thank you for all of the suggestions @antonis! I've applied them in https://github.com/wordpress-mobile/AztecEditor-Android/pull/1073/commits/17946f2be64e3177a4d1c52d98e53ac79985a1b2 and https://github.com/wordpress-mobile/AztecEditor-Android/pull/1073/commits/2f735642025c364e5a6003505f40d1ee93159237 🚀

PR #1083, opened 2024-07-08
[Build] Update Kotlin to 1.9.24

ParaskP7:
Required By: #1084

This PR updates Kotlin to [1.9.22](https://github.com/JetBrains/kotlin/releases/tag/v1.9.22).

### Description

This update isn't that necessary, but should be considered anyway as it will do the following:
- Align this library, to that Kotlin version, which is already used by most of this library's clients and libraries.
- Unblock the usage of the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin for this project.
- Prepares this project for the Kotlin 2+ major update.

### Test

- Smoke test the `app` example app.
- Verifying that all the CI checks are successful.
- However, if you want to be thorough about reviewing this change, you could:
    - Quickly smoke test, either the [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) and/or [DOAndroid](https://github.com/bloom/DayOne-Android) apps, with this version of `Aztec`, or via composite builds, and see if it works as expected.
    - Test for build time increases (local & CI).

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,5 @@
 pluginManagement {
-    gradle.ext.kotlinVersion = '1.9.22'
+    gradle.ext.kotlinVersion = '1.9.24'
     gradle.ext.agpVersion = '8.1.0'
     gradle.ext.automatticPublishToS3Version = '0.8.0'
 
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,5 +1,5 @@
 pluginManagement {
-    gradle.ext.kotlinVersion = '1.6.10'
+    gradle.ext.kotlinVersion = '1.9.22'
     gradle.ext.agpVersion = '8.1.0'
     gradle.ext.automatticPublishToS3Version = '0.8.0'
 
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,7 @@ allprojects {
     }
     tasks.withType(KotlinCompile).all {
         kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8
             allWarningsAsErrors = true
         }
     }


settings.gradle
```diff
@@ -1,5 +1,5 @@
 pluginManagement {
-    gradle.ext.kotlinVersion = '1.6.10'
+    gradle.ext.kotlinVersion = '1.9.22'
```
wzieba:
```suggestion
    gradle.ext.kotlinVersion = '1.9.24'
```

I think `1.9.24` would be a better choice, to cover against bugs. I understand it might be not in sync with the rest of the projects, but I don't think it's a reason to use lower PATCH version of a depencency.

wzieba (approved):

settings.gradle
ParaskP7: Yea @wzieba and thanks for the suggestion, I honestly was thinking about that too, at first I wanted to use the latest `1.9` version on all the Kotlin updates on all the libs I've touched so far, but then decided against that, not wanting to bring a Kotlin update via a transitive dependency update for the client apps to use this newer version of a library... 🤷

Sure, the newest [1.9.23](https://github.com/JetBrains/kotlin/releases/tag/v1.9.23) & [1.9.24](https://github.com/JetBrains/kotlin/releases/tag/v1.9.24) versions bring some improvements, but nothing too obvious for me.

If you really think this to be the better move here I'll go ahead and update Kotlin to `1.9.24` here and on the other 2 repos I touched recently too (ref: [this](https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/146) and [this](https://github.com/wordpress-mobile/WordPress-MediaPicker-Android/pull/77))

wzieba: I think it's just a good hygiene to keep dependencies up to date, especially if we talk only about PATCH versions.

> not wanting to bring a Kotlin update via a transitive dependency update for the client apps to use this newer version of a library... 🤷

I understand, but in clients apps we still are prone to update Kotlin by a 3rd party transitive dependency. We can address this by using `strict` constraint on Kotlin, but then it has its own consequences.

This is nothing very important, hence my approval, but as we update dependencies now, I think we should do this to the newest, reasonably stable release.

ParaskP7: Ok, let's do that then, update to the latest stable release and take it from there, we can keep a close eye on it anyway and evaluate if anything unexpected was to happen, I am not to concerned about it! 💯 

PS: I'll do the same [here](https://github.com/wordpress-mobile/WordPress-Utils-Android/pull/146) and [here](https://github.com/wordpress-mobile/WordPress-MediaPicker-Android/pull/77).

wzieba: Thanks!

PR #1086, opened 2024-07-18
Fix memory leak in EnhancedMovementMethod

khaykov:
### Fix

With a help from Leak Canary I found a leak in EnhancedMovementMethod.

In this PR we are using weak reference for a click listener. Task list listener is not leaking, so it's ok to keep it as is.

### Test
1. Add link tap handler to aztec instance in Main Activity like this:

```
 aztec = Aztec.with(visualEditor, sourceEditor, toolbar, this)
            ....
            .setLinkTapEnabled(true)
            .setOnLinkTappedListener(object : AztecText.OnLinkTappedListener {
                override fun onLinkTapped(widget: View, url: String) {
                    val intent = Intent(Intent.ACTION_VIEW)
                    intent.data = Uri.parse(url)
                    startActivity(intent)
                }
            })
            ....
```
2. Open the app in tap on the link.
3. Confirm that it works and the browser is opened.

### Review
@[USER_NAME]


zwarm (approved):
All looks good @khaykov 
I was able to tap on a link and launch the browser. 🥳

PR #1087, opened 2024-07-19
Remove redundant context from AztecDynamicImageSpan

khaykov:
### Fix

It looks like the context in AztecDynamicImageSpan was never used, but we still pass it from all the spans that are based on AztecDynamicImageSpan. This is a vector for memory leaks, so I'm removing it.

### Test
1. Make sure the app builds, and all the spans look normal.
2. Make sure the changes make sense.


zwarm (approved):
Great find @khaykov 👏 

I tested according to instructions, as well as two devices side-by-side (this branch and trunk). No differences in behavior. 👍

PR #1088, opened 2024-07-25
Fix media span and task list memory leaks

khaykov:
### Fix
This PR fixes memory leaks in media and task list spans.

### Test
1. Toggle items in task list, make sure they work as expected.
2. Delete images, confirm that you see a "media deleted" snackbar.
3. Perform same actions after changing device orientation or bringing up to foreground from background.


zwarm (approved):
@khaykov - Tested and all works as expected. 👍 

I have one question: Might it be safer to use 
`val drawable = contextRef.get()?.resources?.getDrawable(R.drawable.ic_checkbox, null)`

instead of

`val d: Drawable = contextRef.get()!!.resources.getDrawable(R.drawable.ic_checkbox, null)`

I am approving the PR, but will leave the merge for you so you'll have a chance to review my question.

khaykov:
@zwarm Good point! Switched to elvis operator 👍

PR #1092, opened 2024-10-29
Improve performance of toHtml method by moving it to background thread

planarvoid:
### Fix

In this PR I'm introducing a new approach to `toHtml` method. Previously it would block the UI thread. The reason was that the `toHtml` would work with the `Editable` object that represents the content of the editor. However, the object could change if the user would type anything. This is normally not an issue but since our `toHtml` method is not the fastest, there would be times where the user would change the span at the same time when we were trying to read it and we would have crashes. 

The new approach introduced here is to make copy of the `Editable` object on the main thread and handle the processing from spannable to HTML on a background thread asynchronously. Since there is no standard way to make a copy of an `Editable` object, I'm simply copying the string content and reapplying the spans again. I think this works well but we should watch carefully if there are any new crashes appearing. 

The libraries are upgraded because the app wouldn't let me run otherwise.

### Test
1. Run the new test
2. Try editing/applying spans/saving in the DO app

### DIFF
--- a/aztec/build.gradle
+++ b/aztec/build.gradle
@@ -59,10 +59,11 @@ dependencies {
 
     testImplementation "junit:junit:$jUnitVersion"
     testImplementation "org.robolectric:robolectric:$robolectricVersion"
-    testImplementation 'androidx.test:core:1.4.0'
+    testImplementation 'androidx.test:core:1.6.1'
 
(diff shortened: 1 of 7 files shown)


AmandaRiu (approved):
Reviewed the code, ran the automated test, and verified all works well on the Day One side. :shipit:

PR #1093, opened 2024-11-13
Add redrawViews method to PlaceholderManager

planarvoid:
### Fix

In order to redraw placeholder manager from a composable we need to extract the redrawViews method.

### Test
1. Test with the related DO PR

### Review
@danilo04 


danilo04 (approved):
It does redraw when it is called (e.g. when device is rotated).

PR #1095, opened 2024-12-10
Compose placeholders

planarvoid:
### Fix

In this PR I'm implementing a parallel solution that allows you to draw placeholders in Jetpack Compose instead of Android Views. This is actually more reliable, especially when considering redrawing which was hard to do with AV.
I created the compose placeholder manager from the previous manager and cleaned up things that I think are not necessary with compose. It should be much more straightforward because the compose view should be redrawn when the state changes without any forced redrawns.

This PR is much smaller than it seems because `ViewPlaceholderManager` is just a rename of `PlaceholderManager` so it doesn't need to be reviewed. 

### Test
Test this feature with the related DO PR.

### Review
@danilo04 


media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ViewPlaceholderManager.kt
```diff
@@ -0,0 +1,555 @@
...
+        adapters[placeholderAdapter.type] = placeholderAdapter
+    }
+
+    /**
+     * Call this method to manually insert a new item into the aztec text. There has to be a adapter associated with the
+     * item type.
+     * @param type placeholder type
+     * @param attributes other attributes passed to the view. For example a `src` for an image.
+     */
+    suspend override fun insertItem(type: String, vararg attributes: Pair<String, String>) {
```
AmandaRiu: ⚠️ Non-canonical modifier order

```suggestion
    override suspend fun insertItem(type: String, vararg attributes: Pair<String, String>) {
```

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ViewPlaceholderManager.kt
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    /**
+     * Call this method to insert an item with an option to merge it with the previous item. This could be used to
+     * build a gallery of images on adding a new image.
+     * @param type placeholder type
+     * @param shouldMergeItem this method should return true when the previous type is compatible and should be updated
+     * @param updateItem function to update current parameters with new params
+     */
+    suspend override fun insertOrUpdateItem(
```
AmandaRiu: ⚠️ Non-canonical modifier order
```suggestion
    override suspend fun insertOrUpdateItem(
```

media-placeholders/src/main/java/org/wordpress/aztec/placeholders/ViewPlaceholderManager.kt
```diff
@@ -0,0 +1,555 @@
...
+     * Use this function to either update or remove an item. The decision whether to remove or update will be made
+     * based upon the results of the parameter functions. An example of usage is a gallery of images. If the user wants
+     * to remove one image in the gallery, they would call this method. If the removed image is one of many, they might
+     * want to update the current parameters instead of removing the entire gallery. However, if the removed image is
+     * the last one in the gallery, they will probably want to remove the entire gallery.
+     * @param uuid UUID of the span we want to remove or update
+     * @param shouldUpdateItem This function should return true if the span can be updated, false if it should be removed
+     * @param updateItem Function that updates the selected item
+     */
+    suspend override fun removeOrUpdate(uuid: String, shouldUpdateItem: (Attributes) -> Boolean, updateItem: (currentAttributes: Map<String, String>) -> Map<String, String>): Boolean {
```
AmandaRiu: ⚠️ Non-canonical modifier order
```suggestion
   override suspend fun removeOrUpdate(uuid: String, shouldUpdateItem: (Attributes) -> Boolean, updateItem: (currentAttributes: Map<String, String>) -> Map<String, String>): Boolean {
```

AmandaRiu (commented):
@planarvoid Tested in the Day One app using a Samsung device and these latest changes work nicely! I made a couple minor code comments for your review. I'm not merging this PR since it looks like it may be waiting for Klym and Danilo as well, but on my end it looks good. 👍🏻

danilo04 (approved):
This has been tested in Day One Android and is working as expected!

PR #1096, opened 2025-01-23
Intercept touch event to task list checkboxes to handle it before the cursor is moved

danilo04:
### Fix

This PR improves the UX of task lists checkboxes touch events. The changes are a little hacky but that is how the current solution to work with checkboxes work currently.

### Test
1. Check that touch events work for all kind of formats
2. Tap on the ordered lists numbers and see that the focus and cursor are moved to that item
3. Tap on the checkboxes for task lists and check that they don't move the cursor but the action of check/unchecked is performed.

### Review
@planarvoid 


aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
```diff
@@ -718,7 +718,41 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
...
+            // If the tap event was on the leading margin, we double check whether we are tapping on a task list item.
+            // If that is true, then we return false because we don't want to propagate the tap event to stop moving
+            // the cursor to the item tapped.
+            if (getTaskListHandler()?.handleTaskListClick(
+                text,
+                off,
+                x,
+                totalPaddingStart
+            ) == true) {
+                refreshText()
```
khaykov: I had a whole bunch of code here that was supposed to preserve a cursor, but the only thing that was actually doing was just passing `false` into `refreshText`.

```suggestion
                refreshText(false)
```

khaykov (approved):
Thanks, @danilo04 ! It works well. Maybe we can reduce the padding to 15, what do you think?

I also have a suggestion for preserving a cursor. Let me know what do you think :)

danilo04:
Thank you for the recommendation @khaykov, I implemented both changes.

aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
danilo04: Man, that is what happens when you don't read the method's signature. Thank you for the pointer.

khaykov (approved):
Looks good, thank you!

PR #1097, opened 2025-01-28
Allow setting spanned spanned content with processing

khaykov:
### Fix

In this PR we add a method for setting a spanned content, that will process it before setting it into the editor. This is useful if you want to parse HTML into Spanned text somewhere else (eg, on a different thread which is useful if you only use Aztec to display content).

### Test
1. Make sure demo app work, and content is being set.


zwarm (approved):
@khaykov - this is a nice improvement. The demo app is working as expected. 👍

PR #1099, opened 2025-02-17
Ensure spans validity when getting text copy

khaykov:
Not sure how to reproduce the issue, but sometimes spans that we reset might return -1 for start and end position. I suspect that service spans (like selection, etc) might be causing this.

This PR protects against this case.

We do not have a test case for copying texts in the demo app, so please test this with the client version that is utilizing it.


zwarm (approved):
@khaykov - Nice solution. LGTM :shipit:

PR #1101, opened 2025-03-11
Switch threads when necessary when pre-processing html

khaykov:
### Fix
Since HTML parsing can be performed from BG thread, we need to make sure we execute code that handles views on main thread.

Make sure placeholders are displayed correctly.


planarvoid (approved):
Looks reasonable 👍

PR #1102, opened 2025-03-18
Add touch event handling in AztecText when field is disabled

danilo04:
### Fix

When the `AztecText` is disabled (`isEnabled = false`), touch events aren't handled by `EnhancedMovementMethod.onTouchEvent`. In this PR, we handle touch events for links in `Aztectext.onTouchEvent` is it is disabled. Since touch events at this level are executed, we can reuse the same logic in `EnhancedMovementMethod` to handle touch events on links. 

### Test
1. Test with client side PR

### Review
@khaykov 

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -756,6 +756,12 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
             }
         }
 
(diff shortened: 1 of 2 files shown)


khaykov:
Thanks, @danilo04 ! It appears to be working as expected, but since this changes default behavior, I think we need to put behind an opt-in flag.

danilo04:
@khaykov Added flag for tap on links when aztec text is disabled.

khaykov (approved):
Thanks, @danilo04 ! Works great 👍

PR #1103, opened 2025-04-24
Improve behaviour of Compose Placeholder manager

planarvoid:
### Fix

In this PR I'm fixing behaviour of the compose placeholder manager when handling changes to a placeholder that don't change the placeholder size. For example switching order of photos in a gallery. To achieve this I've added `equals` implementation to `AztecAttributes`. I'm using this to determine of a placeholder has changed and should be redrawn.

### Test
Test this with the linked DayOne PR


obsantos (approved):
LGTM 👍

PR #1104, opened 2025-06-12
Allow multiple selectionChangedListeners to be set in AztecText

Twinsen81:
### Fix

The `AztecText.setOnSelectionChangedListener` wasn't working from the client side, as the editor's internal code was [overriding](https://github.com/wordpress-mobile/AztecEditor-Android/blob/ca1e1e05f578f0636e3d0759c3ac97205aae2583/aztec/src/main/kotlin/org/wordpress/aztec/toolbar/AztecToolbar.kt#L412) the listener for its purposes. Refactored the code to allow multiple listeners.

### Test

- Check out the `plokhoves/fix-selection-listener` branch in AS
- Run the app
- Type some text in the editor
- Select a word and apply the bold style
- [x] Observe that the selected text is displayed in bold
- Tap the selection handles and move them to select more or less text 
- [x] Observe that the bold icon on the toolbar changes the color to black when only the bolded text is selected  

https://github.com/user-attachments/assets/e362c9f7-034c-4e75-84fc-b2adfe4ed021

### Review

@mzorz

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -327,6 +327,8 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
 
     private lateinit var listItemStyle: BlockFormatter.ListItemStyle
 
+    private var selectionChangedListeners = CompositeSelectionChangedListener()
+
     override fun onDraw(canvas: Canvas) {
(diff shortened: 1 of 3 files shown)


mzorz (approved):
Nice addition of the interface so we can expose the listeners 👍 

I wonder if the listeners could mess up with each other in certain occasions, nothing becomes apparent right now on the demo app, but could it happen that a listener has access to the editor and changes the text somehow and as such the other listeners would no longer have valid data. Perhaps we could "block" changes from within a listener 🤔 - but beyond the scope of this change for sure, just leaving it here as a thought.

LGTM :shipit:

PR #1105, opened 2025-06-12
Fix instrumentation tests execution

wzieba:
Closes: [AINFRA-752](https://linear.app/a8c/issue/AINFRA-752/fix-aztec-android-ui-tests)

### Description

This PR fixes instrumented tests execution by updating model to the new default.

Previously used, Pixel 2, has been decommissioned.

To check all available devices, you can use `gcloud firebase test android models list`

### Testing
Green CI means we're good to go.

Or, if you wish, you can run `.buildkite/commands/connected-tests.sh` locally (apply secrets before) and wait for the result.

After the latest changes, I've run tests 3 times on Firebase, and they passed each time.


app/src/androidTest/kotlin/org/wordpress/aztec/demo/pages/EditLinkPage.kt
```diff
@@ -33,7 +33,17 @@ class EditLinkPage : BasePage() {
...
+        try {
+            urlField.perform(click())
+            urlField.perform(replaceText(url), ViewActions.closeSoftKeyboard())
+        } catch (e: RuntimeException) {
+            // If the URL field is not visible, it might be because the keyboard is open.
+            // Close the keyboard and try again.
+            urlField.perform(ViewActions.closeSoftKeyboard())
+            urlField.perform(replaceText(url), ViewActions.closeSoftKeyboard())
+            e.printStackTrace()
+        }
```
wzieba, line 46: This is the core change: a few of additional safety checks and delays to make UI tests not flaky on CI.

app/src/main/res/layout/activity_main.xml
```diff
@@ -46,7 +45,6 @@
                 android:layout_width="match_parent"
                 android:layout_height="match_parent"
                 android:gravity="top|start"
-                android:hint="@string/source_hint"
```
wzieba, line 49: The `android:hint` looked quite strange on the device when running tests: almost like a normal test. After removing them, tests started to pass. I'm not entirely sure why (as they were flaky and the failure was not reproducible locally), so here's my best guess.

I think that's fair as it only affects the test app.

wzieba:
eh, I don't know, the tests are still flaky on Firebase Test Lab - I can't reproduce the problem locally.

Issue #1106, opened 2025-06-25
App crashes with setSpan IndexOutOfBoundsException


prashantsaini1:
Recently we saw over 10 crashes on Firebase, with this stack trace.
- Device - `Samsung S23 Ultra`.
- OS - `Android 15`.
- AztecEditor `v1.6.2`.

```
Fatal Exception: java.lang.IndexOutOfBoundsException: setSpan (120 ... 119) has end before start
       at android.text.SpannableStringBuilder.checkRange(SpannableStringBuilder.java:1319)
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:685)
       at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:677)
       at androidx.emoji2.text.SpannableBuilder.setSpan(SpannableBuilder.java:140)
       at org.wordpress.aztec.util.SpanWrapper.setSpanOrLogError(SpanWrapper.kt:42)
       at org.wordpress.aztec.util.SpanWrapper.setStart(SpanWrapper.kt:28)
       at org.wordpress.aztec.handlers.ListItemHandler.handleNewlineAtStartOfBlock(ListItemHandler.kt:23)
       at org.wordpress.aztec.handlers.BlockHandler.handleTextChanged(BlockHandler.kt:62)
       at org.wordpress.aztec.watchers.BlockElementWatcher.onTextChanged(BlockElementWatcher.kt:81)
       at android.widget.TextView.sendOnTextChanged(TextView.java:13461)
       at android.widget.TextView.handleTextChanged(TextView.java:13590)
       at android.widget.TextView$ChangeWatcher.onTextChanged(TextView.java:17363)
       at android.text.SpannableStringBuilder.sendTextChanged(SpannableStringBuilder.java:1268)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:577)
       at androidx.emoji2.text.SpannableBuilder.replace(SpannableBuilder.java:315)
       at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:508)
       at androidx.emoji2.text.SpannableBuilder.replace(SpannableBuilder.java:305)
       at androidx.emoji2.text.SpannableBuilder.replace(SpannableBuilder.java:49)
       at android.view.inputmethod.BaseInputConnection.replaceTextInternal(BaseInputConnection.java:1027)
[...]

PR #1107, opened 2025-06-25
fix: span end should not be before the start

prashantsaini1:
### Fix
[IndexOutOfBoundsException crash](https://github.com/wordpress-mobile/AztecEditor-Android/issues/1106) issue.

### Test
1. We could not reproduce it on local but we had multiple reports from Firebase on Samsung devices. Looks like it's specific to some custom emoji implementation as per the logs in the original issue.

### Review
@mzorz 


prashantsaini1:
Hi @Twinsen81, request you to please review and merge this PR as this is causing lots of crashes for some of our users.

Twinsen81 (approved):
LGTM! Thanks for the fix!

prashantsaini1:
@Twinsen81 Could you please get this merged? Thanks!

PR #1116, opened 2026-02-20
Add redacted text formatting action

zwarm:
### Fix
Add a new "redacted" inline formatting action that:
- Applies a red background color and strikethrough to selected text
- Adds a corresponding toolbar button
- Serializes to `<del class="redacted">` in HTML with full round-trip support
- Distinguishes redacted spans from regular strikethrough (`<del>`, `<s>`, `<strike>`) via the `class="redacted"` attribute

### Test
1. Open the demo app
2. Type some text and select it
3. Tap the redacted button (rectangle with diagonal line) in the toolbar
4. Verify text gets a red background and strikethrough
5. Tap the button again to toggle it off
6. Switch to HTML view and verify `<del class="redacted">` markup
7. Switch back to visual view and verify formatting is preserved

### Review


zwarm:
This was a test PR to validate the AI skills

PR #1118, opened 2026-03-13
Fix checkbox interaction for indented (nested) task list items

zwarm:
### Fix
 Fix checkbox interaction for indented (nested) task list items. 

> [!NOTE]
> This PR can be tested with DayOne-Android via the companion PR [here](https://github.com/bloom/DayOne-Android/pull/7372)
 
Three issues were addressed:                                                                       
  1. Cannot check indented items: The x-coordinate gate in AztecText.onTouchEvent() and TaskListClickHandler only allowed taps within a single leading margin width, rejecting taps on nested checkboxes positioned further right. The handler now computes the tap boundary based on the nesting depth of the tapped item.           
  2. Checking item above indented item checks both: isChecked() in AztecTaskListSpan fetched all AztecListItemSpan children (including nested ones) but used a line index that only counted direct children, causing an index mismatch. Now filters spans by nestingLevel to match.                                                    
  3. Wrong item gets strikethrough: Same root cause as #2 — after any nested item, every subsequent checkbox rendered the state of the wrong item.
                                                                                                                                                                     
### Test Steps  
1. Check out the branch dodroid-954-checklist-issue
2. Build and run the sample app                                                                                                                                                           
2. Create a task list with a few items
3. Indent one item in the middle using the indent toolbar button                                                                                                   
4. Tap the checkbox on the indented item — it should toggle independently                                                                                          
5. Tap the checkbox on the item directly above the indented item — only that item should be checked                                                                
6. Tap the checkbox on the item directly below the indented item — the correct item should be checked and struck through                                           
7. Verify items at all positions (above, below, indented) can be checked and unchecked independently                     

### DIFF
--- a/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
+++ b/aztec/src/main/kotlin/org/wordpress/aztec/AztecText.kt
@@ -751,16 +751,11 @@ open class AztecText : AppCompatEditText, TextWatcher, UnknownHtmlSpan.OnUnknown
         x += scrollX
         y += scrollY
 
-        // We check whether the user tap on a checkbox of a task list. The aztec text field should be enabled and we
-        // check whether the tap event was on the leading margin which is where the checkbox is located plus a padding
-        // space used to increase the target size for the tap event.
-        if (isEnabled &&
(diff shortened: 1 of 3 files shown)


planarvoid (approved):
Tested with the linked PR, everything works well 👍

PR #1119, opened 2026-03-18
Update wordpress:utils 3.5.0 -> 3.15.0

ygnessin:
### Fix

Updates the `wordpress:utils` dependency to the latest: https://github.com/wordpress-mobile/WordPress-Utils-Android/releases 

Primary motivation is that the GET_ACCOUNTS permission gets pulled into any apps using Aztec. This was fixed in version 3.8.0 of `wordpress:utils`.

Many other dependencies are severely out of date as well, hopefully there is a plan to update them at some point.

### Test
Ran all unit tests and they passed

### Review
??? Not sure

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,7 +105,7 @@ ext {
     robolectricVersion = '4.13'
     jUnitVersion = '4.13.2'
     jSoupVersion = '1.15.3'
-    wordpressUtilsVersion = '3.5.0'
+    wordpressUtilsVersion = '3.15.0'
     espressoVersion = '3.0.1'
 
     // other


ygnessin:
@zwarm It looks like you are the most recent contributor to this project. Can you help land this PR?

zwarm:
> @zwarm It looks like you are the most recent contributor to this project. Can you help land this PR?

Hi @ygnessin, thanks for the contribution! I’ll need to take a deeper look at how this change could impact the apps currently using this library before moving forward. The utils haven’t been updated since 2024, and quite a bit has changed since then.

This review may take some time as I’m currently focused on higher-priority work, but I’ll share an update as soon as I’m able to dig in. 🙇

zwarm (commented):
Thanks for the contribution — removing the `GET_ACCOUNTS` permission bleed is a worthwhile fix!

A few things before this merges:

**1. Changelog review**
This jumps 10 minor versions (3.5.0 → 3.15.0). Could you skim the [WordPress-Utils-Android releases](https://github.com/wordpress-mobile/WordPress-Utils-Android/releases) for any behavioral changes or deprecations in the classes Aztec uses (`AppLog`, `ImageUtils`, `PermissionUtils`, `ToastUtils`, `DeviceUtils`) and confirm nothing breaks?

**2. Transitive dependency check**
The main goal is removing `GET_ACCOUNTS` — worth verifying no new permissions or unexpected transitive dependencies were introduced in the process. You can check with:
```
./gradlew :aztec:dependencies
```
A before/after diff of the output would be ideal.

**3. Manual smoke test on the demo app**
Unit tests passing is great. Since `ImageUtils` is used in real image-handling paths, a quick manual pass on image insertion in the demo app would help confirm nothing regressed there.

